Issue #7, opened 2020-05-06
CloudEvents SDK V2 & Eventing Channel Updates


travis-minke-sap:
Currently the implementation is based on CloudEvents SDK V1 and we need to catch up with the rest of knative eventing by moving the SDK to V2.  Additionally, and probably more importantly, the implementation needs to move to the new/latest knative-eventing channel implementations for sending/receiving CloudEvents.

This is expected to be the final set of changes carried over from the [kyma-incubator/knative-kafka](https://github.com/kyma-incubator/knative-kafka) project and is a patch from [Issue #108](https://github.com/kyma-incubator/knative-kafka/issues/108) in that repo.

Issue #8, opened 2020-05-06
Reevaluate confluent-kafka-go decision 


davyodom:
confluent-kafka-go was originally chosen for this channel implementation over sarama for several reasons (this was more than a year ago)
- Ordering was not guaranteed in error cases by sarama producer
- consumer group distribution with sarama-cluster over multiple processes elastically had issues
- manual offset commit support lacking in sarama
- only supported older kafka versions and did not have full kafka feature support
- confluent backing of confluent-kafka-go and librdkafka is more or less guaranteed and kept up to date

The choice of confluent-kafka-go does introduce some challenges though, primarily around cgo builds.  This causes a divergence in developer experience here vs all other knative components.  We should revisit the initial reasons for the choice as a lot has changed since then, and see if the decision is still justified.  Performance, and open source community / governance model should also be taken into consideration and discussion.

slinkydeveloper:
Completely off topic _but not so much_, @pierDipi proposed to create the alternative kafka broker using java for the data-plane, exactly to avoid issues with "half baked" kafka golang libs:  https://docs.google.com/document/d/10-qylWrj7Tj81EqoIiZ2AANAZNmkKe7XKmFobgZULKY/edit?usp=sharing

davyodom:
/assign @phamilton
prow was broken before :)

phamilton:
Please find our performance comparison results here https://github.com/phamilton/go-kafka-benchmark. From a performance POV it is clear that except when utilizing the Azure EventHubs Kafka compatibility layer Sarama performance is significantly better than that provided by confluent-kafka-go and is comparable to the Java client in most cases.

Since using the Kafka layer of Azure EventHubs is likely pretty rare, and also based on the performance we see, not a great choice. It doesn't seem like that should be a blocker to switch to Sarama.

From a functional standpoint we did not find any blockers for moving to Sarama.

lionelvillard:
> comparable to the Java client in most cases

it seems the Java client is much faster on the producer side. Any insights on why?

pierDipi:
From a quick look, it seems like default configurations among libraries are different.

For example, `acks`
- `Java` = `1` - see https://kafka.apache.org/documentation/#producerconfigs
- `librdkafka` = `-1` = `all brokers` - see https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md#topic-configuration-properties
[...]

PR #9, opened 2020-05-07
CloudEvents SDK V2 & Eventing Channel Updates

travis-minke-sap:
This PR is for Issue #7 and is expected to be the final duplicate delivery from the kyma-incubator repository ([knative-kafka](https://github.com/kyma-incubator/knative-kafka)) project.

The changes are to move the implementation to the CloudEvents SDK V2 and to align with more recent versions of the Knative eventing Channel framework.

The new files (retry*.go) are simply a refactoring of the removed `client` package - the logic therein is the same.


slinkydeveloper:
Hi @davyodom , i don't see on this pr that you're using the `binding.Message`/`binding.MessageWriter` constructs to marshal/unmarshal the message, is that something you're planning to do later?

travis-minke-sap:
Hi slinkydeveloper, 

In the channel's `main.go` we are using the `binding.ToEvent()` to convert the `Message` into a cloud event.  The current implementation is/was based on cloud events and since there currently isn't any protocol for kafka_confluent this was the easiest transition.  Similarly, in the dispatcher's `retry.go` file we are using the `binding.ToMessage()` to convert our cloud event into a `Message`.  

Not sure if this addresses your question?  Let me know if there's a better approach.

slinkydeveloper:
> In the channel's main.go we are using the binding.ToEvent() to convert the Message into a cloud event. The current implementation is/was based on cloud events and since there currently isn't any protocol for kafka_confluent this was the easiest transition. Similarly, in the dispatcher's retry.go file we are using the binding.ToMessage() to convert our cloud event into a Message.

Ok maybe i overlooked this PR :smile: BTW Why are you going to event and back? Do you really need to materialize the event? Can you just use the Message?

travis-minke-sap:
TLDR - We already had the manual logic to go from a Confluent Message to a CloudEvent `Event` and didn't want to re-write that until we know better which kafka library we'll settle on.

Yeah - you're right it's inefficient and there are notes in the code to that effect.  This is only because the current implementation is based on Cloud Events and we're now tying into the knative messaging which is based on the sdk v2 `Messages`.  This was just the least-effort mechanism to convert between the two given the current implementation.

In the longer term one of the following will likely adjust/improve this...
1 - If the aligned implementation settles on the Sarama library, then we can use the protocol `kafka_sarama` to go directly from sarama messages to cloud event messages.
2 - If we stay with confluent the we can either look at writing another protocol `kafka_confluent` to do the same - OR - we could alter our current implementation to manually go between the Confluent Message and the CloudEvent Message instead of having this CloudEvent `Event` intermediary.

slinkydeveloper:
>  If we stay with confluent the we can either look at writing another protocol kafka_confluent to do the same

I think this is the best solution. In the end you don't need a full protocol, but just the implementation of `Message` and `WriteTo[]` (as explained here: https://cloudevents.github.io/sdk-go/protocol_implementations.html)
Also, look at the implementation of the kafka channel in eventing-contrib to check how we leverage `Message` and base code from eventing: https://github.com/knative/eventing-contrib/blob/master/kafka/channel/pkg/dispatcher/dispatcher.go#L135

travis-minke-sap:
Thanks for the documentation link - hadn't seen that - will take a look!

I have looked extensively at the eventing-contrib kafka (we're borrowing lots of things from there and following your lead - thanks!).  The difference is simply the fact that we're on Confluent and didn't have the protocol handy.

slinkydeveloper:
If you want, i can contribute to that :wink:

slinkydeveloper:
But i suppose this depends on https://github.com/knative-sandbox/eventing-kafka/issues/8

travis-minke-sap:
Yeah - exactly - we should settle on the kafka library first - but then we'd definitely welcome any help with the protocol implementation - thanks!

PR #10, opened 2020-05-12
Release build

### LINKED ISSUE #3: Establish release support
A /hack/release.sh script must be added to support nightly builds. An early poc shows that ko and yaml can be used for this purpose.  This was done in the poc by adding and using the librdkafka c libraries in the dependency tree (now supported by current versions of confluent-kafka-go).  

However, a local developer experience will need to be somewhat different as cross compilation has some special requirements with cgo builds.  An option could be running the ko build from inside of a debian container, driven by a makefile, and connecting to the docker daemon and kubectl running on the local host. Otherwise, it seems that there would be quite a few initial setup steps (or script) which would need to be written for each os.

If these approaches are taken or something similar, helm charts and helm dependency can be removed at the same time

Adequate developer documentation should be added for whichever solution is determined here

davyodom:
Fixes #3 
I upgraded confluent-kafka-go libraries, which now include the librdkafka c libraries inside.  That let me convert to ko builds. However, the ko build must be run from inside of a docker container or else there are what seem to be unavoidable cross compilation issues.  I wrote a simple script in the hack folder to build a local development docker image and run a ko build inside of it, using docker and kubeconfig from the host directly.


davyodom:
/hold
Need to figure out how to test the release script reasonably before merging

davyodom:
/cc @lberk @matzew 
Adding as you are listed as the kafka reviewers in eventing-contrib

davyodom:
/cc @vaikas 
in case you want to look as well

Abd4llA:
> or else there are what seem to be unavoidable cross compilation issues

@davyodom do these cross compilation issues happen when building on a non-linux os?

davyodom:
> @davyodom do these cross compilation issues happen when building on a non-linux os?

Yes that's correct. E.g. when building on a mac, os specific c libraries are used for the build and the go binary is built against those.  If we put that go binary in a debian container (as we are with the ko builds), then the binary does not work even if all of the (again os specific) c libraries are present.

PR #14, opened 2020-05-18
Go Modules Conversion and Udpates

### LINKED ISSUE #6: Convert to Go Mod
Get rid of dep and switch over! Lots of examples in eventing and eventing-contrib to reference here.

travis-minke-sap:
Fixes #6

Convert to go modules (from dep)
Update to latest (as of 5/15/20) eventing-contrib (and eventing / pkg /etc)
Fix breakages from update (SubscribableStatus moved to v1beta1, and kncontroller.StartAll() signature change)
Re-sort imports from initial repo migration.


davyodom:
/cc @vaikas @matzew @lberk 
sorry to spam, need to figure out who should be reviewers/approvers on this repo

Makefile
```diff
@@ -59,12 +59,12 @@ clean-all: clean-channel clean-common clean-controller clean-dispatcher
...
 #
-# Dep / Fmt / Vet / Lint
+# Mod / Fmt / Vet / Lint
 #
 
-dep:
+mod:
 	@echo 'Ensuring Dependencies'
-	cd $(BUILD_ROOT); dep ensure -v
+	cd $(BUILD_ROOT); go mod tidy; go mod vendor
```
vaikas, line 67: nit, but should this use the &&?
go mod tidy && go mod vendor?

pkg/dispatcher/testing/kafkachannel.go
```diff
@@ -76,10 +77,10 @@ func WithSubscriber(uid types.UID, uri string) KafkaChannelOption {
 func WithSubscriberReady(uid types.UID) KafkaChannelOption {
 	return func(nc *v1alpha1.KafkaChannel) {
 		if nc.Status.SubscribableStatus == nil {
-			nc.Status.SubscribableStatus = &eventingduck.SubscribableStatus{}
+			nc.Status.SubscribableStatus = &eventingduckv1alpha1.SubscribableStatus{}
 		}
 
-		nc.Status.SubscribableStatus.Subscribers = append(nc.Status.SubscribableStatus.Subscribers, eventingduck.SubscriberStatus{
+		nc.Status.SubscribableStatus.Subscribers = append(nc.Status.SubscribableStatus.Subscribers, eventingduckv1beta1.SubscriberStatus{
```
vaikas, line 83: If you change the status to be v1beta1, you should also annotate it appropriately.
annotations: messaging.knative.dev/subscribable: v1beta1https://github.com/knative/eventing/blob/master/docs/spec/channel.md

vaikas: And it's a bit odd that the subscribable shape is v1alpha1 and status is v1beta1. I think that's going to be a problem. Also, what is the actual channel type? As in, which duck is it? v1alpha1 or v1beta1?

Makefile
travis-minke-sap: 👍

pkg/dispatcher/testing/kafkachannel.go
travis-minke-sap: Yeah - I found that odd as well  ;)  but... the v1alpha1 SubscribableStatus (See eventing/pkg/apis/duck/v1alpha1/subscribable_types.go) was changed by you a few weeks ago (see https://github.com/knative/eventing/commit/6a84ca3211ebb3e590ad1d70d0872ded66a594c5#diff-f2c5f66fb8660a1ab3f4cbb3b3a8ae2e)  to use the v1beta1 SubscriberStatus. 

I was simply following the eventing-contrib/kafka implementation's adaptation to that as well by updating our use of v1alpha1 to be able to compile.  Am I missing something?

Also - would we need/want the v1beta1 annotation since we're not fully v1beta compliant yet (still using the v1alpha1 parent structs) ?  Our implementation is based on the eventing-contrib/kafka KafkaChannel CRD which still has the legacy value of "true" for the annotation.

vaikas: aha, never mind :) I was being dump, yeah, those two are wire compatible so that's all good. Sorry for the noise :( 

Yeah, so if your shape is still v1alpha1 duck (not that subfield which is identical between (v1beta1 / v1alpha1), just leave it be. One thing I'd advice is to set the annotation however to v1alpha1 for this duck shape.

travis-minke-sap: No worries - always appreciate the close review - still learning the ecosystem ; )   I will add/set the annotation as you suggest.   fwiw - It doesn't look like any of the eventing-contrib implementations are doing this as of yet? (just double checking my code-read)

vaikas:
/test pull-knative-sandbox-eventing-kafka-go-coverage
from what I understand coverage has been having a bad day, retrying...

davyodom:
Added a comment here https://github.com/knative/test-infra/issues/1945 for follow up on the coverage build

PR #17, opened 2020-05-20
Adding approvers

davyodom:
Adding in the kafka approvers from eventing-contrib and the lead developers responsible for the initial channel implementation who will be the most active in ongoing development here


matzew:
Yes!

I am very supportive on this.  In Apache in the Incubator there is a concept that when code moves to the foundation. It usually comes w/ _n_ committers (aka approvers).

That is needed to build a community. We should not shortfall here on this w/ just a few initial approvers, on this incubation (or sandbox) project.

Perhaps that's something for steering or TOC ?

/assign @evankanderson 
/assign @vaikas

matzew:
> Perhaps that's something for steering or TOC ?

But IMO we should not "block" this PR from any potential TOC/steering outcome...

lberk:
I'd support this PR as well

lberk:
actually I'd go a step further and mark each person under reviewers as well (prow pings reviewers first so it'll spread the review load)

davyodom:
> actually I'd go a step further and mark each person under reviewers as well (prow pings reviewers first so it'll spread the review load)

Didn't realize that's how prow works, I'll do that

PR #20, opened 2020-05-27
Handle missing or invalid kafka secrets

### LINKED ISSUE #12: Missing / Invalid Kafka Secret Handling
When configured for `azure` KAFKA_PROVIDER usage, the current implementation handles the case where there are no Kafka Secrets (ie - those labelled with `eventing-kafka.knative.dev/kafka-secret: "true"`).   It does not do so when using standard `kafka` KAFKA_PROVIDER.  

In general we should have the controller be stable when there are no Kafka Secrets, or when the Kafka Secrets are malformed or include the dummy/default data from the yaml.

 It would also be good to document the current handling of edge cases (e.g. deleting secret when KafkaChannels exist, etc) so that we can determine requirements for them in a future issue.

travis-minke-sap:
Fixes #12 

Added minimal stability fixes so that the "Kafka" use case behaves similar to the "EventHubs" use case when the Kafka Secret is missing or invalid.

Also re-added the Makefile which was accidentally removed.  We can refactor it out into hack scripts or something else at some point, but it still has useful development utilities (partial test runs with code coverage and such...)

### DIFF
--- a/pkg/common/kafka/admin/admin_kafka.go
+++ b/pkg/common/kafka/admin/admin_kafka.go
@@ -56,11 +56,12 @@ func NewKafkaAdminClient(ctx context.Context, namespace string) (AdminClientInte
 		return nil, err
 	}
 
-	// Should Only Be One Kafka Secret In The Namespace
+	// Handle Various Numbers Of Kafka Secrets - Currently Only Support One!
 	if len(kafkaSecrets.Items) != 1 {
-		logger.Error(fmt.Sprintf("Expected 1 Kafka Secret But Found %d", len(kafkaSecrets.Items)))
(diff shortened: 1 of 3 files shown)


pkg/common/kafka/admin/admin_kafka.go
```diff
@@ -56,11 +56,12 @@ func NewKafkaAdminClient(ctx context.Context, namespace string) (AdminClientInte
 		return nil, err
 	}
 
-	// Should Only Be One Kafka Secret In The Namespace
+	// Handle Various Numbers Of Kafka Secrets - Currently Only Support One!
 	if len(kafkaSecrets.Items) != 1 {
```
phamilton, line 60: Comment is a little weird, if it only supports 1.

travis-minke-sap: yep - the implementation evolved out from under the comment - will clarify ; )

Issue #16, opened 2020-05-18
Dispatcher Retry & Dead Letter Queues


travis-minke-sap:
Currently the implementation has custom retry logic which is preventing the use of the eventing MessageDispatcher.DispatchMessage()'s ability to send to the Dead Letter Queue.  Doing so would result in a separate Dead Letter Queue message for every retry attempt.  Need to look at retry implementations in conjunction with the Knative Eventing implementations to design a more flexible solution.

sbcd90:
Hi @travis-minke-sap , I would like to work on this. Is it possible to kindly assign this to me?

davyodom:
/assign @sbcd90 
Sure, give it a shot! It's possible that some change might be required for the eventing library.  The goal here is to share library usage with eventing, but maintain the ability for a reasonable retry, some of which now spans the cloud event sdk (v2 usage coming in an open pr).  Some configurability might need to be added separately there or the knative library as per some discussions in the WG.  Take a look and sketch out your approach/proposal and we can discuss on Wednesday (we have a holiday until then)

sbcd90:
Hi @davyodom , thanks a lot. I created a document to initially propose my approach & trigger further discussions.
https://docs.google.com/document/d/1z3psNbyH7YnskzgzjLPdOFcLel4npld0ea7YwUJkZ4o/edit?usp=sharing

travis-minke-sap:
Hi @sbcd90 - Thanks for putting together the proposal!  Unfortunately I don't think what you're suggesting will meet our requirements / needs.  Aside from being rather heavy-weight in it's use of another Topic / Consumer, it breaks our `in-order` guarantees by going async.  I haven't done much investigation into possible solutions as of yet, but my initial thought was one of the following...

  - A small change to what we currently have whereby we detect the fact that we're on the "last" retry attempt, and then pass the DeadLetter to the MessageDispatcher.  This is essentially the `Max no. of retries reached` decision point in your proposed solution.  I have no idea how easy/hard that might be.

  - Stepping back and taking a larger view by digging into the Knative MessageDispatcher implementation to see whether it could natively support the "retry" ability for all channels (optional, configurable, etc.) thereby eliminating the need for our wrapping retry library.  This might be achieved via the native CloudEvent retry capability which I don't believe Knative is currently making use of.

...and we're certainly open to other ideas / approaches as well as long as they preserve our `at-least-once` / `in-order` guarantees and don't impact cluster resource consumption and / or performance too much.

[...]

PR #27, opened 2020-06-14
#16 Dispatcher Retry & Dead Letter Queues

sbcd90:
@travis-minke-sap this is a solution proposal as per your approach 1.

### DIFF
--- a/pkg/dispatcher/dispatcher/retry.go
+++ b/pkg/dispatcher/dispatcher/retry.go
@@ -26,14 +26,16 @@ func (d *Dispatcher) Dispatch(event *cloudevents.Event, subscription Subscriptio
 		logger = d.Logger.With(zap.String("Event", event.String()))
 	}
 
+	timesOfRetries := d.calculateNumberOfRetries()
+
 	// Create A New Retry Runner With Configured Backoff Behavior
-	retryRunner := retry.New(
+	retryRunner := NewMiddleware(
 		retry.Config{
 			DisableBackoff: !d.ExponentialBackoff, // Note Negation: Env Var Is Whether To Enable & Retry Config Is Whether To Disable ; )
-			Times:          d.calculateNumberOfRetries(),
+			Times:          timesOfRetries,
 			WaitBase:       time.Millisecond * time.Duration(d.InitialRetryInterval),
 		},
-	)
+	)(nil)
 
 	//
 	// Convert The CloudEvent To A binding/Message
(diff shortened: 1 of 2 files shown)


pkg/dispatcher/dispatcher/retry.go
```diff
@@ -58,13 +60,21 @@ func (d *Dispatcher) Dispatch(event *cloudevents.Event, subscription Subscriptio
...
 	if subscription.Delivery != nil && subscription.Delivery.DeadLetterSink != nil && !subscription.Delivery.DeadLetterSink.URI.IsEmpty() {
 		// TODO - Currently ignoring dead-letter configuration due to wrapping retry implementation - would send one deadletter for every retry :(
-		// deadLetterURL = subscription.Delivery.DeadLetterSink.URI.URL()
-		logger.Warn("Subscription Delivery DeadLetterSink Not Currently Supported!")
+		deadLetterURL = subscription.Delivery.DeadLetterSink.URI.URL()
+		// logger.Warn("Subscription Delivery DeadLetterSink Not Currently Supported!")
 	}
 
 	// Attempt To Dispatch The CloudEvent Message Via Knative Message Dispatcher With Retry Wrapper
 	err := retryRunner.Run(context.Background(), func(ctx context.Context) error {
```
lionelvillard, line 68: FYI, the cloudevent SDK supports retries: https://github.com/cloudevents/sdk-go/blob/master/v2/protocol/http/protocol_retry.go#L16

Do you think you could leverage it?

pkg/dispatcher/dispatcher/retry.go
```diff
@@ -58,13 +60,21 @@ func (d *Dispatcher) Dispatch(event *cloudevents.Event, subscription Subscriptio
 	var deadLetterURL *url.URL
 	if subscription.Delivery != nil && subscription.Delivery.DeadLetterSink != nil && !subscription.Delivery.DeadLetterSink.URI.IsEmpty() {
```
lionelvillard, line 61: you may want to use `pkg/resolver/URIResolver` to support object reference

travis-minke-sap: 👍  Not a part of original scope (existing code borrowed from eventing-contrib) but would be great to harden - thanks for suggesting!

travis-minke-sap:
Hi @sbcd90 - Thanks for the PR and for looking into the proposed approach (detecting the final retry attempt)!

After going through the PR I see that you've been forced to copy the `slock/goresilience` retry logic out into a new function so that it could be modified to update the context in order to track the number of attempts.  Obviously that's not a great approach since we'd want to keep that copy of the implementation up to date with newer versions of the `slock/goresilience` library.  I don't see another way, however, to be able to track the state across the retry calls given the limited nature of what the library exposes.

So... instead, I think we'd prefer a simple solution like the one below where we just make an **additional** send request if the retries fail out with an error.  It's not fancy, but is easy to understand/maintain and is effective.

I've not tested this solution as of yet - only mocked it up in code.  And there are a couple **TODOs** there to look into the resolver library that @lionelvillard mentioned above.  If you're willing to continue with it (test it out, write/improve the unit tests, research the resolvers, etc) that would be awesome.  I realize most of the "new code" aspect (the fun part) has been eliminated, but it seems like you've done a great job of jumping into the code base and it would be great to get this delivered.  I'm open to your thoughts if I've mis-understood something or your have another idea ; ). Thanks again for contributing!


Proposed solution...
```
	var deadLetterURL *url.URL
	if subscription.Delivery != nil && subscription.Delivery.DeadLetterSink != nil && !subscription.Delivery.DeadLetterSink.URI.IsEmpty() { // TODO - use pkg/resolver/URIResolver?
		deadLetterURL = subscription.Delivery.DeadLetterSink.URI.URL() // TODO - use pkg/resolver/URIResolver?
	}

	// Attempt To Dispatch The CloudEvent Message Via Knative Message Dispatcher With Retry Wrapper
	err := retryRunner.Run(context.Background(), func(ctx context.Context) error {
		err := d.messageDispatcher.DispatchMessage(ctx, message, nil, destinationURL, replyURL, nil)
		return d.logResponse(err)
	})

	// Retry Attempts Failed
	if err != nil {

		// If The DeadLetterSink Is Configured
		if deadLetterURL != nil {

			// Then Make One Final Attempt With It & Return Result
			logger.Info("Making One Final Attempt With DeadLetterSink Configured")
			err := d.messageDispatcher.DispatchMessage(context.Background(), message, nil, destinationURL, replyURL, deadLetterURL)
			if err != nil {
				logger.Error("Failed To Send Final Attempt With DeadLetterSink", zap.Error(err))
			}
			return err

		} else {

			// Otherwise Return Original Retry Failure (DeadLetter Sink Not Configured)
			logger.Error("Failed To Send After Configured Number Of Retries - DeadLetterSink Not Configured", zap.Error(err))
			return err
		}
	}

	// Return Success
	return nil
```

sbcd90:
@travis-minke-sap @lionelvillard thanks a lot for reviewing. I'll continue working on this PR as per your suggestions.

travis-minke-sap:
/close

Closing PR as work has been done by #34 and will be further refined to use the new-ish knative DispatchWithRetry() functionality in the future.

PR #30, opened 2020-06-22
Add InitializeTracing function and call it from controller and dispatcher main

### LINKED ISSUE #22: Support Eventing Tracing & Observability Configuration
In order to improve our alignment with the core Knative Eventing implementations we should support the `config-tracing` and `config-observability` global ConfigMaps in the `knative-eventing` namespace.  This is similar to the work we did to align on the `config-logging` implementation.

eric-sap:
Fixes #22

In order to have tracing (via the config-tracing configmap) work properly, a call to SetupDynamicPublishing() needs to be made.  This adds those calls to the main functions of the controller and dispatcher.

### DIFF
--- a/pkg/common/k8s/tracing.go
+++ b/pkg/common/k8s/tracing.go
@@ -1,0 +1,35 @@
+package k8s
+
+import (
+	"context"
+	"go.uber.org/zap"
+	"k8s.io/client-go/kubernetes"
+	"knative.dev/eventing/pkg/tracing"
+	injectionclient "knative.dev/pkg/client/injection/kube/client"
+	"knative.dev/pkg/configmap"
+	"knative.dev/pkg/system"
+	tracingconfig "knative.dev/pkg/tracing/config"
+)
+
+//
+// Initialize The Specified Context With A Tracer (ConfigMap Watcher)
+// Assumes ctx Has K8S Client Injected Via LoggingContext (Or Similar)
+//
+func InitializeTracing(logger *zap.SugaredLogger, ctx context.Context, service string) {
+	k8sClient, ok := ctx.Value(injectionclient.Key{}).(kubernetes.Interface)
+	if !ok {
+		logger.Fatalw("Error getting kubernetes client from context (invalid interface)")
+	}
+
+	// Create A Watcher On The Tracing ConfigMap & Dynamically Update Tracing Configuration
+	cmw := configmap.NewInformedWatcher(k8sClient, system.Namespace())
+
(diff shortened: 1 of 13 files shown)


pkg/common/k8s/tracing.go
```diff
@@ -0,0 +1,35 @@
...
+
+	if err := tracing.SetupDynamicPublishing(logger, cmw, service, tracingconfig.ConfigName); err != nil {
+		logger.Fatalw("Error setting up dynamic trace publishing", zap.Error(err))
+	}
+
+	// Start The Tracing ConfigMap Watcher
+	if err := cmw.Start(ctx.Done()); err != nil {
+		logger.Fatalw("Failed To Start ConfigMap Watcher", zap.Error(err))
+	}
+}
```
travis-minke-sap, line 35: Possible to have a tracing_test.go to verify the logic?

eric-sap: Always a good idea.  Not super-easy to test the failure conditions since they exit the process but it can at least have a happy-path test.

pkg/common/k8s/tracing_test.go
```diff
@@ -0,0 +1,55 @@
...
+	logtesting "knative.dev/pkg/logging/testing"
+	"knative.dev/pkg/system"
+	tracingconfig "knative.dev/pkg/tracing/config"
+	"os"
+	"testing"
+)
+
+// Test The LoggingContext() Functionality
+func TestInitializeTracing(t *testing.T) {
+
```
travis-minke-sap, line 20: copy paster comments - need to be updated ; ). this is not a logging test

pkg/common/k8s/tracing_test.go
```diff
@@ -0,0 +1,55 @@
...
+	// Test Data
+	ctx := context.TODO()
+	service := "TestService"
+
+	// Obtain a Test Logger (Required By Tracing Function)
+	logger := logtesting.TestLogger(t)
+
+	// Setup Environment
+	assert.Nil(t, os.Setenv(system.NamespaceEnvKey, constants.KnativeEventingNamespace))
+
```
travis-minke-sap, line 30: Is this all it needed - what about the new SERVICE_NAME env var?

pkg/common/k8s/tracing_test.go
eric-sap: oops, fixed

pkg/common/k8s/tracing_test.go
eric-sap: I could put that in here, but it isn't actually used in the InitializeTracing function (it's passed in directly from main, which isn't an explicitly-tested function).

Issue #42, opened 2020-08-17
Define Dynamic Spec Config Requirements For Channel/Dispatcher


travis-minke-sap:
This issue is a placeholder reminder to circle back and define specific requirements around the handling of various configuration values for the Channel (Producer) and Dispatcher (Consumer) K8S Deployments.

The **eventing-kafka** ConfigMap exposes values related to the runtime characteristics of the Channel / Dispatcher deployments created by the Controller (i.e.  replicas, cpu_request, cpu_limit, number of consumers, etc...).  These values are used when creating new Deployments for the Channel and Dispatcher.  Currently, however, changing them in the ConfigMap will only apply the new values to new Deployments and will not update existing Deployments.

It is unclear whether the current behavior is desirable or not, as there are pros and cons for either approach.  For instance, making the ConfigMap always update existing Deployments removes the ability to fine-tune or manually change existing deployments (since the Controller would re-reconcile them back to their ConfigMap values).

So... we need to think through the needs of an administrator responsible for running an **eventing-kafka** cluster and what level of automation vs customizability is appropriate.

PR #43, opened 2020-08-19
Add support for ConfigMap watching to update Sarama and other Eventing-Kafka settings dynamically

### LINKED ISSUE #36: Add support for a ConfigMap containing all the Sarama Kafka configuration
Currently the Sarama configuration is set (in pkg/common/kafka/util/util.go) via some parameters and constants.  We would like to use a configmap for these settings where possible and reasonable.  The eventing-contrib/kafka implementation already has support for such a map, though with minimal content.  We’d like to reuse whatever makes sense (such as map watching) but with our own defined content and parsing.  This configmap may also include any other custom eventing-kafka config as well (i.e. it will not be Sarama-specific).

#### Watching the Configmap
This is what eventing-contrib/kafka does, which seems like a good starting-point for eventing-kafka work toward a similar end:

kafka/channel/pkg/reconciler/controller/controller.go::NewController (passed in from sharedmain) uses the passed-in configmap.Watcher to call cmw.Watch("config-kafka", updateKafkaConfig), where "updateKafkaConfig" is a function that loads the configmap data and replaces the Reconciler.kafkaConfig with the one parsed out of the configmap.

The configmap parser itself uses code from knative.dev/pkg/configmap/parse.go that might be useful for our purposes as well (common way to parse out durations, ints, floats, etc).

If we don't want to switch to using sharedmain, we will need to create the watcher ourselves via sharedmain.SetupConfigMapWatchOrDie, which is what the pkg/common/k8s/observability.go code already does (not sure if there's a benefit to sharing a single watcher or not)

#### Loading Sarama Settings from the Configmap
Whether we need the huge block of _all_ of the sarama settings or not, putting them as JSON into a configmap is a relatively ordinary procedure and results in something that looks like this:
```
apiVersion: v1
data:
  sarama: '{"Admin":{"Retry":{"Max":5,"Backoff":100000000},"Timeout":3000000000},"Net":{"MaxOpenRequests":5,"DialTimeout":30000000000,"ReadTimeout":30000000000,"WriteTimeout":30000000000,"TLS":{"Enable":true},"SASL":{"Enable":true,"Mechanism":"PLAIN","Version":1,"Handshake":true,"AuthIdentity":"","User":"TestUsername","Password":"TestPassword","SCRAMAuthzID":"","TokenProvider":null,"GSSAPI":{"AuthType":0,"KeyTabPath":"","KerberosConfigPath":"","ServiceName":"","Username":"","Password":"","Realm":""}},"KeepAlive":30000000000,"LocalAddr":null,"Proxy":{"Enable":false,"Dialer":null}},"Metadata":{"Retry":{"Max":3,"Backoff":250000000},"RefreshFrequency":300000000000,"Full":true,"Timeout":0},"Producer":{"MaxMessageBytes":1000000,"RequiredAcks":1,"Timeout":10000000000,"Compression":0,"CompressionLevel":-1000,"Idempotent":false,"Return":{"Successes":false,"Errors":true},"Flush":{"Bytes":0,"Messages":0,"Frequency":0,"MaxMessages":0},"Retry":{"Max":3,"Backoff":100000000}},"Consumer":{"Group":{"Session":{"Timeout":10000000000},"Heartbeat":{"Interval":3000000000},"Rebalance":{"Timeout":60000000000,"Retry":{"Max":4,"Backoff":2000000000}},"Member":{"UserData":null}},"Retry":{"Backoff":2000000000},"Fetch":{"Min":1,"Default":1048576,"Max":0},"MaxWaitTime":250000000,"MaxProcessingTime":100000000,"Return":{"Errors":false},"Offsets":{"CommitInterval":0,"AutoCommit":{"Enable":true,"Interval":1000000000},"Initial":-1,"Retention":0,"Retry":{"Max":3}},"IsolationLevel":0},"ClientID":"TestClientId","RackID":"","ChannelBufferSize":256,"Version":{}}'
kind: ConfigMap
metadata:
  name: test
  namespace: test
```
Of course, the point here is that you can omit any entries from the "sarama" JSON that you want to leave as the defaults, so you could just have something that looks considerably simpler:
```
apiVersion: v1
data:
  sarama: '{"Net":{"TLS":{"Enable":true},"SASL":{"Enable":true,"Mechanism":"PLAIN","Version":1,"User":"TestUsername","Password":"TestPassword"}},"Metadata":{"RefreshFrequency":300000000000},"ClientID":"TestClientId"}'
kind: ConfigMap
metadata:
  name: test
  namespace: test
```
Merging the JSON with the default sarama.Config is also not complex:
```
config := NewSaramaConfig(...)
json.Unmarshal([]byte(configMap.Data["sarama"]), &config)
```
If we need to save our current settings back to a configmap, that gets complicated (and will likely require our own in-sync copy of the sarama.Config struct with custom tags to ignore some JSON-unfriendly fields).  This is possible and has been POC'd but will hopefully not be necessary for this issue (writing to the configmap will only be done manually when an administrator wants to make changes).

#### Implementing Config Changes
Regarding the procedures necessary to rebuild/restart sarama infrastructure when the config map changes, there are three main components of note:

- The admin client (only one, recreated every time)

The ReconcileKind function calls SetKafkaAdminClient(), which calls kafkaadmin.CreateAdminClient() every time, so we don't have an "old admin client with old settings" to worry about for the moment. This will change if we decide to keep the admin client around in the future but for now this appears to be a non-issue (the only change would be to load the settings from whatever the current configmap is instead of using constants and/or environment variables).

- The ConsumerGroup (potentially several; the dispatcher is horizontally scalable, so each pod would make ConsumerGroups for a collection of subscriberSpecs)

The implementation for the ConsumerGroup is all in one place, but there could be several pods, which would need to be reconfigured (or recreated). The UpdateSubscriptions() function loops through the subscribers and calls CreateConsumerGroup whenever it finds no subscriber wrapper for a particular subscriber UID. The ConsumerGroup goes into the SubscriberWrapper and remains as part of the DispatcherImpl (created in the dispatcher's main.go and remaining for the life of the pod).
If the ConfigMap watcher is only going to run on the controller, it might be easiest to restart all of the dispatcher pods to pick up any new configuration data (the controller could parse the configmap and see if the changes are actually relevant to the ConsumerGroup first if desired)

Risk of trouble from restarting dispatcher pods is low, since the critical data is all stored in a kafka topic, not the eventing-kafka pods themselves, but unnecessary pod churn is, of course, preferably avoided. We could easily start with just having the controller restart the dispatcher pods when the configmap changes (which will presumably not be particularly often). Adjustments to a working system are likely easier than trying to work through all of the potential logic of changing the ConsumerGroups on the fly (either by having the dispatchers also watch the configmap or having a different IPC mechanism from the controller).

- The Producer (only one at the moment)

The Producer is similar to the ConsumerGroup in that it is created (once, in the NewProducer() function called by the channel's main.go code) and then used repeatedly (during handleMessage() which is called by when a message is received (due to the eventingchannel.NewMessageReceiver() call in the channel's main.go).

If we handle configmap changes by restarting the producer pod, we must be certain that we do not encounter any situation in which an event has been received (i.e. handleMessage has been called) and a non-error returned, but the message not persisted in the Kafka topic.

I don't see any real way for this to happen, as the ProduceKafkaMessage waits for the delivery report via the SyncProducer's SendMessage function before giving handleMessage the go-ahead to return no-error. If the process is terminated anywhere in that execution path, the caller of handleMessage (which is the ServeHTTP function in eventing's message_receiver.go file) will get an error (network or otherwise) that will be passed back to the caller (that is, the customer). Errors are acceptable for a brief time during reconfiguration; if less downtime is desired, the producer could watch the configmap directly and reconfigure the SyncProducer itself instead of letting the controller bounce it.

As with the ConsumerGroup it is probably best to start with a naïve restarting of pods and see if the required performance during configuration changes demands a modification to that approach.

eric-sap:
Fixes #36

Controller, Producer, and Dispatcher all now watch the config-eventing-kafka configmap for changes to sarama and other settings.  The Sarama settings are applied to the SyncProducer and ConsumerGroups immediately, but other eventing-kafka settings are not currently used in that way (they are loaded by the controller at startup and used instead of the old environment variables, but changes do not restart the other pods yet).  The AdminClient will use any new sarama settings whenever necessary.


pkg/common/config/constants.go
```diff
@@ -0,0 +1,11 @@
+package config
+
+// Package Constants
+const (
+	// The name of the configmap used to hold eventing-kafka settings
+	SettingsConfigMapName = "config-eventing-kafka"
+
```
mattmoor, line 7: Remove trailing whitespace:
```suggestion

```

pkg/channel/constants/constants.go
```diff
@@ -19,4 +19,6 @@ const (
 	CeKafkaHeaderKeySubject      = "ce_subject"
 	CeKafkaHeaderKeyDataSchema   = "ce_dataschema"
 	CeKafkaHeaderKeyPartitionKey = "ce_partitionkey"
+
+	KnativeEventingNamespace     = "knative-eventing"
```
mattmoor, line 23: Format Go code:
```suggestion
	KnativeEventingNamespace = "knative-eventing"
```

pkg/common/config/constants.go
```diff
@@ -0,0 +1,10 @@
+package config
+
+// Package Constants
+const (
+	// The name of the configmap used to hold eventing-kafka settings
+	SettingsConfigMapName = "config-eventing-kafka"
+	// The name of the keys in the Data section of the eventing-kafka configmap that holds Sarama and Eventing-Kafka configuration YAML
```
mattmoor, line 7: Format Go code:
```suggestion

```

pkg/channel/constants/constants.go
mattmoor, line 23: Format Go code:
```suggestion
	KnativeEventingNamespace = "knative-eventing"
```

travis-minke-sap:
/assign @travis-minke-sap 

fwiw - we have a follow-up issue (#42) to define requirements around some of the non-sarama configuration.

pkg/controller/env/environment.go
```diff
@@ -228,43 +107,78 @@ func GetEnvironment(logger *zap.Logger) (*Environment, error) {
...
+func (err ControllerConfigurationError) Error() string {
+	return "controller: invalid configuration (" + string(err) + ")"
+}
+
+// VerifyOverrides overwrites an EventingKafkaConfig struct with the values from an Environment struct
+// The fields here are not permitted to be overridden by the values from the config-eventing-kafka configmap
+// VerifyOverrides returns an error if mandatory fields in the EventingKafkaConfig have not been set either
+// via the external configmap or the internal variables.
+func VerifyOverrides(configuration *config.EventingKafkaConfig, environment *Environment) error {
+	configuration.ServiceAccount = environment.ServiceAccount
```
travis-minke-sap, line 130: see comments from other components VerifyOverrides() functions...

pkg/controller/kafkachannel/dispatcher.go
```diff
@@ -252,14 +253,14 @@ func (r *Reconciler) newDispatcherDeployment(channel *kafkav1beta1.KafkaChannel)
 					},
 				},
 				Spec: corev1.PodSpec{
-					ServiceAccountName: r.environment.ServiceAccount,
+					ServiceAccountName: r.config.ServiceAccount,
 					Containers: []corev1.Container{
```
travis-minke-sap, line 257: yeah, again here i don't see why we can't continue to use the environment (if that's where they are) instead of artificially adding to the config ?

pkg/controller/kafkachannel/reconciler.go
```diff
@@ -171,3 +187,32 @@ func (r *Reconciler) reconcile(ctx context.Context, channel *kafkav1beta1.KafkaC
...
+
+	// Load the Sarama settings from our configmap, ignoring the eventing-kafka result.
+	saramaConfig, _, err := commonconfig.LoadConfigFromMap(configMap)
+	if err != nil {
+		r.logger.Fatal("Failed To Load Eventing-Kafka Settings", zap.Error(err))
+	}
+
+	r.logger.Info("ConfigMap Changed; Updating Sarama Configuration")
+	r.saramaConfig = saramaConfig
+}
```
travis-minke-sap, line 218: is this the same function as in the channel?  can/should it be in common/config ?

pkg/controller/kafkasecret/controller.go
```diff
@@ -39,11 +41,23 @@ func NewController(ctx context.Context, _ configmap.Watcher) *controller.Impl {
...
+		logger.Fatal("Invalid / Missing Settings - Terminating", zap.Error(err))
+	}
+
 	// Create The KafkaSecret Reconciler
 	r := &Reconciler{
 		logger:             logging.FromContext(ctx),
 		kubeClientset:      kubeclient.Get(ctx),
-		environment:        environment,
+		config:             configuration,
 		kafkaChannelClient: injectionclient.Get(ctx),
```
travis-minke-sap, line 61: again - it's ok to have both - don't have to eliminate by mushing data into config - loses some clarity as to where stuff is coming from

pkg/controller/util/channel.go
```diff
@@ -63,27 +64,27 @@ func ChannelHostName(channelName, channelNamespace string) string {
...
-func NumPartitions(channel *kafkav1beta1.KafkaChannel, environment *env.Environment, logger *zap.Logger) int32 {
+func NumPartitions(channel *kafkav1beta1.KafkaChannel, configuration *config.EventingKafkaConfig, logger *zap.Logger) int32 {
 	value := channel.Spec.NumPartitions
 	if value <= 0 {
-		logger.Debug("Kafka Channel Spec 'NumPartitions' Not Specified - Using Default", zap.Int32("Value", environment.DefaultNumPartitions))
-		value = environment.DefaultNumPartitions
+		logger.Debug("Kafka Channel Spec 'NumPartitions' Not Specified - Using Default", zap.Int32("Value", configuration.Kafka.Topic.DefaultNumPartitions))
+		value = configuration.Kafka.Topic.DefaultNumPartitions
 	}
 	return value
```
travis-minke-sap, line 73: all this might have to be reverted if we go back to using the environment

pkg/dispatcher/controller/kafkachannel_test.go
```diff
@@ -177,6 +178,10 @@ func NewMockDispatcher(t *testing.T) MockDispatcher {
 func (m MockDispatcher) Shutdown() {
 }
 
-func (m MockDispatcher) UpdateSubscriptions(_ []eventingduck.SubscriberSpec) map[eventingduck.SubscriberSpec]error {
+func (m MockDispatcher) UpdateSubscriptions(subscriberSpecs []eventingduck.SubscriberSpec) map[eventingduck.SubscriberSpec]error {
+	return nil
+}
```
travis-minke-sap, line 183: you added back in the variable name instead of _ which will cause intellij to flag it as unused?  any reason - still appears not to be used?

pkg/dispatcher/dispatcher/dispatcher.go
```diff
@@ -3,29 +3,35 @@ package dispatcher
...
+	Username               string
+	Password               string
+	ChannelKey             string
+	StatsReporter          metrics.StatsReporter
+	ExponentialBackoff     bool
+	InitialRetryInterval   int64
+	MaxRetryTime           int64
+	SaramaConfig           *sarama.Config
+	currentSubscriberSpecs []eventingduck.SubscriberSpec
 }
```
travis-minke-sap, line 35: Would call this `SubscriberSpecs` (caps) for consistency and there's no alternative to "current"

pkg/dispatcher/dispatcher/dispatcher.go
```diff
@@ -84,6 +91,14 @@ func (d *DispatcherImpl) Shutdown() {
 // Update The Dispatcher's Subscriptions To Align With New State
 func (d *DispatcherImpl) UpdateSubscriptions(subscriberSpecs []eventingduck.SubscriberSpec) map[eventingduck.SubscriberSpec]error {
 
+	// Save the current subscriber specs so that reconfigure() can use them to recreate the Dispatcher if necessary
+	d.currentSubscriberSpecs = subscriberSpecs
+
```
travis-minke-sap, line 96: Do we want to save them all?  I think instead you want to save the `activeSubscriptions` below on line 158 (after they've been filtered/checked/closed) ?

pkg/dispatcher/dispatcher/dispatcher.go
```diff
@@ -228,3 +243,67 @@ func (d *DispatcherImpl) closeConsumerGroup(subscriber *SubscriberWrapper) {
...
+// Reconfigure takes a new sarama.Config struct and applies the updated settings,
+// restarting the Dispatcher/ConsumerGroups if required
+func (d *DispatcherImpl) reconfigure(config *sarama.Config) Dispatcher {
+	d.Logger.Info("Configuration received; applying new Consumer settings")
+
+	// "Reconfiguring" the Dispatcher involves creating a new one, but we can re-use some
+	// of the original components, such as the list of current subscribers
+	d.Shutdown()
+	d.DispatcherConfig.SaramaConfig = config
+	newDispatcher := NewDispatcher(d.DispatcherConfig)
```
travis-minke-sap, line 305: Why do you need to set the `d.DispatcherConfig.SaramaConfig` when you're creating a new one and that one will be garbage collected ?

pkg/dispatcher/env/environment.go
```diff
@@ -127,3 +97,61 @@ func GetEnvironment(logger *zap.Logger) (*Environment, error) {
...
+func (err DispatcherConfigurationError) Error() string {
+	return "dispatcher: invalid configuration (" + string(err) + ")"
+}
+
+// VerifyOverrides overwrites an EventingKafkaConfig struct with the values from an Environment struct
+// The fields here are not permitted to be overridden by the values from the config-eventing-kafka configmap
+// VerifyOverrides returns an error if mandatory fields in the EventingKafkaConfig have not been set either
+// via the external configmap or the internal variables.
+func VerifyOverrides(configuration *config.EventingKafkaConfig, environment *Environment) error {
+	configuration.Metrics.Port = environment.MetricsPort
```
travis-minke-sap, line 114: ditto

PR #51, opened 2020-09-14
Use server flag for kubernetes server url

julz:
Part of https://github.com/knative/community/issues/193.

Similarly to https://github.com/knative/pkg/pull/1693 in pkg, and in line with latest kubectl, drop `--master` flags in favour of `--server`.

### DIFF
--- a/cmd/channel/main.go
+++ b/cmd/channel/main.go
@@ -33,7 +33,7 @@ const (
 // Variables
 var (
 	logger        *zap.Logger
-	masterURL     = flag.String("masterurl", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
+	serverURL     = flag.String("server", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
 	kubeconfig    = flag.String("kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
 	kafkaProducer *producer.Producer
 )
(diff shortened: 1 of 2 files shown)


cmd/channel/main.go
```diff
@@ -33,7 +33,7 @@ const (
 // Variables
 var (
 	logger        *zap.Logger
-	masterURL     = flag.String("masterurl", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
+	serverURL     = flag.String("server", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
```
markusthoemmes, line 36: can we potentially rework to just use the pkg helper, if there are no other flags anyway?

julz: I don't have much context on this package or why it doesn't use pkg tbh, but that sure sounds good if we could: @travis-minke-sap @eric-sap @slinkydeveloper wdyt?

julz: it looks to me like it's not an entirely trivial swap we pass `masterURL` to `InitializeKafkaChannelLister` below, so my vote would be to merge this to fix the immediate issue and then look at adopting helpers from pkg separately.. but defer to folks with more context on all these things :).

markusthoemmes: indeed! SG!

travis-minke-sap: The changes look fine and I agree on the approach (deferring rework to pkg helper).  The only reason for not using some "pkg helper" is just ignorance of its existence.  Can someone point to a github location of this code and I'll take a look at incorporating it into a future delivery?

markusthoemmes: Sure thing: https://github.com/knative/pkg/blob/master/injection/sharedmain/main.go. There are a few `OrDie` helpers, that generally help setting up a component (getting a config etc.)

travis-minke-sap: 👍

travis-minke-sap: /approve

travis-minke-sap: /lgtm

PR #64, opened 2020-09-29
Refactor configmap comparison

### LINKED ISSUE #46: Refactor ConfigMap Loading/Comparison
The recently introduced ConfigMap support could be improved in a few ways and we'd like to refactor to reduce the complexity / maintenance.  The following areas need some love...

**NOTE** - We need to keep the new custom extraction logic which custom loads the Sarama.Version and TLS.Config info from the YAML string.  The overall lifecycle of the ConfigMap watchers is reasonable as well and doesn't necessarily require attention at this point.

### Lifecycle
There are a lot of different exposed (and package private) lifecycle functions related to the ConfigMap handling including... 

  - LoadSettings
  - LoadConfigFromMap
  - NewSaramaConfig
  - MergeSaramaSettings
  - UpdateSaramaConfig
  - enforceSaramaConfig

..which have a tangled call stack and expose too many different usage patterns.  We do need the ability to support a couple of slightly different usage patterns between the SharedMain Controller and non-SharedMain Channel/Dispatcher but it should be possible to reduce the complexity here so that it's easier to reason about.   Essentially just create a single entry point constructor and maybe another merge function or something along those lines.

### Comparison
Because the original requirements weren't super well defined the comparison of before/after Sarama Config structs was made more complicated than is necessary.  We don't need to Marshall the struct out to JSON String, and can ignore the function pointer fields, etc.  Currently we have a mirror struct with JSON advice as a means of excluding things from comparison but a more straightforward approach would be to use the `cmp` package with ignores and some manual comparision similar to this unit test...

```
func TestCmp(t *testing.T) {

	usernameA := "TestUsernameA"
	usernameB := "TestUsernameB"

	config1 := sarama.NewConfig()
	config1.Net.SASL.User = usernameA

	config2 := sarama.NewConfig()
	config2.Net.SASL.User = usernameA

	config3 := sarama.NewConfig()
	config3.Net.SASL.User = usernameB

	ignoredTypes := cmpopts.IgnoreTypes(config1.Consumer.Group.Rebalance.Strategy, config1.MetricRegistry, config1.Producer.Partitioner)
	ignoredUnexported := cmpopts.IgnoreUnexported(config1.Version)

	diff := cmp.Diff(config1, config2, ignoredTypes, ignoredUnexported)
	t.Log("Diff = " + diff)

	assert.True(t, cmp.Equal(config1, config2, ignoredTypes, ignoredUnexported))
	assert.Equal(t, config1.Version, config2.Version)
	assert.Equal(t, config1.Consumer.Group.Rebalance.Strategy, config2.Consumer.Group.Rebalance.Strategy)

	if config1.Version != config2.Version {
		t.Errorf("Version Inequality")
	}

	assert.False(t, cmp.Equal(config1, config3, ignoredTypes, ignoredUnexported))
}
```

eric-sap:
Fixes #46 

**Kept**
LoadSettings
UpdateSaramaConfig
MergeSaramaSettings

**Removed**
NewSaramaConfig (new values are set directly in MergeSaramaSettings, which was the only place that needed them)
LoadConfigFromMap (merged with LoadSettings)
enforceSaramaConfig (no longer necessary after other refactoring)

Comparisons now use go-cmp package which obviates the necessity for a duplicate CompareSaramaConfig struct.


pkg/channel/distributed/common/config/configwatcher.go
```diff
@@ -2,13 +2,10 @@ package config
 
 import (
 	"context"
```
mattmoor, line 4: Format Go code:
```suggestion
	"context"

```

pkg/channel/distributed/common/kafka/sarama/sarama.go
```diff
@@ -4,179 +4,43 @@ import (
 	"context"
 	"crypto/tls"
 	"crypto/x509"
-	"encoding/json"
 	"fmt"
```
mattmoor, line 7: Format Go code:
```suggestion
	"fmt"
	"log"
	"os"
	"regexp"

```

pkg/channel/distributed/common/kafka/sarama/sarama.go
```diff
@@ -4,179 +4,43 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	commonconfig "knative.dev/eventing-kafka/pkg/channel/distributed/common/config"
 	"knative.dev/eventing-kafka/pkg/channel/distributed/common/kafka/constants"
 	"knative.dev/eventing-kafka/pkg/channel/distributed/common/testing"
 	kubeclient "knative.dev/pkg/client/injection/kube/client"
+	"knative.dev/pkg/system"
+	"log"
+	"os"
+	"regexp"
```
mattmoor, line 21: Format Go code:
```suggestion

```

pkg/channel/distributed/common/kafka/sarama/sarama_test.go
```diff
@@ -247,30 +237,30 @@ func TestMergeSaramaSettings(t *testing.T) {
...
 
-// Verify that a sarama.Config struct can be serialized into a string without causing JSON errors
-func TestMarshalSaramaJSON(t *testing.T) {
-	config := sarama.NewConfig()
-	jsonResult, err := MarshalSaramaJSON(config)
-	assert.Nil(t, err)
-	assert.NotEqual(t, "", jsonResult)
+	// Verify error when an invalid Version is provided
+	configMap := commontesting.GetTestSaramaConfigMap(commontesting.NewSaramaConfig, commontesting.TestEKConfig)
+	regexVersion := regexp.MustCompile(`Version:\s*\d*\.[\d\.]*`)	// Must have at least one period or it will match the "Version: 1" in Net.SASL
```
mattmoor, line 245: Format Go code:
```suggestion
	regexVersion := regexp.MustCompile(`Version:\s*\d*\.[\d\.]*`) // Must have at least one period or it will match the "Version: 1" in Net.SASL
```

pkg/channel/distributed/common/testing/util.go
```diff
@@ -1,6 +1,7 @@
 package testing
 
 import (
+	"knative.dev/eventing-kafka/pkg/channel/distributed/common/kafka/constants"
```
mattmoor, line 4: Format Go code:
```suggestion

```

pkg/channel/distributed/common/testing/util.go
```diff
@@ -1,6 +1,7 @@
...
 	"net/http"
 	"testing"
 	"time"
```
mattmoor, line 7: Format Go code:
```suggestion
	"time"

	"knative.dev/eventing-kafka/pkg/channel/distributed/common/kafka/constants"
```

mattmoor (commented):
Produced via:
  `gofmt -s -w $(find -path './vendor' -prune -o -path './third_party' -prune -o -name '*.pb.go' -prune -o -type f -name '*.go' -print)`
  `goimports -w $(find -name '*.go' | grep -v vendor | grep -v third_party | grep -v .pb.go | grep -v wire_gen.go)`

pkg/channel/distributed/common/kafka/sarama/sarama.go
```diff
@@ -308,73 +173,64 @@ func extractRootCerts(saramaConfigYamlString string) (string, *x509.CertPool, er
...
-	if err != nil {
-		return saramaConfigYamlString, nil, err
-	} else {
-		updatedSaramaConfigYamlBytes := regex.ReplaceAll([]byte(saramaConfigYamlString), []byte{})
-		return string(updatedSaramaConfigYamlBytes), certPool, nil
-	}
+	regex := regexp.MustCompile(`(?s)\s*RootPEMs:.*-----END CERTIFICATE-----`)
+	updatedSaramaConfigYamlBytes := regex.ReplaceAll([]byte(saramaConfigYamlString), []byte{})
+	return string(updatedSaramaConfigYamlBytes), certPool, nil
 }
```
travis-minke-sap, line 179: So, you're changing the behavior here - it used to continue on if there was a problem with the Regexp but now it will panic.   Is that intentional?  I think what you have is ok in this case because the regexp is static.  We should. probably move it up to the top as a package private variable so that we're not creating the regexp every time through?

pkg/channel/distributed/common/kafka/sarama/sarama.go
```diff
@@ -407,42 +263,37 @@ func MergeSaramaSettings(config *sarama.Config, configMap *corev1.ConfigMap) err
...
 	}
 
-	// Create A New Base Sarama Config To Start From
-	saramaConfig := NewSaramaConfig()
-
-	// Merge The Sarama Settings In The ConfigMap Into the New/Base Sarama Config
-	err := MergeSaramaSettings(saramaConfig, configMap)
+	configMaps := kubeclient.Get(ctx).CoreV1().ConfigMaps(system.Namespace())
+	configMap, err := configMaps.Get(ctx, commonconfig.SettingsConfigMapName, metav1.GetOptions{})
 	if err != nil {
```
travis-minke-sap, line 278: seems odd (non-standard) to do the configmap loading on two lines instead of just....
```
configMap, err := kubeclient.Get(ctx).CoreV1().ConfigMaps(system.Namespace()).Get(ctx, commonconfig.SettingsConfigMapName, metav1.GetOptions{})
```

pkg/channel/distributed/common/kafka/sarama/sarama.go
```diff
@@ -308,73 +173,64 @@ func extractRootCerts(saramaConfigYamlString string) (string, *x509.CertPool, er
...
-// An error of "cannot convert *config (type sarama.Config) to type CompareSaramaConfig" indicates that the
-// CompareSaramaConfig structure above needs to be brought in sync with the current config.Sarama struct.
-func MarshalSaramaJSON(config *sarama.Config) (string, error) {
-	bytes, err := json.Marshal(CompareSaramaConfig(*config))
-	if err != nil {
-		return "", err
-	}
-	return string(bytes), nil
-}
+	ignoredUnexported := cmpopts.IgnoreUnexported(config1.Version, x509.CertPool{}, tls.Config{})
```
travis-minke-sap, line 210: Would be awesome if we could go another level deep on the `tls.Config{}` exclusion.  We have an immediate need to be able to specify some of the bools in tls.Config from the configmap.  I know it's a pain (lot to exclude) but otherwise we have more custom extraction logic similar to Version to deal with.   Thoughts?

pkg/channel/distributed/common/kafka/sarama/sarama.go
eric-sap: This was one of the linter suggestions ("use MustCompile when the string is constant"); I'll move it to the top as suggested though, that makes sense.

pkg/channel/distributed/common/kafka/sarama/sarama.go
eric-sap: That's true; originally I was checking to see if the configMaps variable was nil and I'm not sure why I removed that check but left it as separate lines.  Most other places that call CoreV1().ConfigMaps don't check for nil either though, and after a quick glance it looks like buried in the k8s.io code it just calls newConfigMaps() which can't ever return nil.  Will recombine as suggested.

pkg/channel/distributed/common/kafka/sarama/sarama.go
travis-minke-sap: yeah, i think it's a good change - thanks!

pkg/channel/distributed/common/kafka/sarama/sarama.go
travis-minke-sap: 👍

pkg/channel/distributed/common/kafka/sarama/sarama.go
eric-sap: The IgnoreUnexported call specifies structures which need to have private fields ignored in the comparison.  For tls.Config this includes only mutex, sessionTicketKeys, and autoSessionTicketKeys.  InsecureSkipVerify is public and as such should still be comparable; I've added calls in the TestMergeSaramaSettings() and TestSaramaConfigEqual() functions to explicitly check this particular action for peace of mind.

travis-minke-sap: perfect - thanks Eric!

PR #74, opened 2020-09-30
Refactor contrib channel pkgs

travis-minke-sap:
This PR is another step in the eventing-contrib/kafka repo-consolidation saga...

The goal of the PR is to move the KafkaChannel implementation from the temporary contrib/ directory into it's intended location at pkg/channel/consolidated.  Some common and webhook code was moved as well.  The cmd/main.go and config YAML was also moved.  

> Note: While the config/ YAML was updated so that the ko:// references are correct, there are a bunch of labels of the form `contrib.eventing.knative.dev/release: devel` which were not changed.


matzew (approved):
/lgtm
/approve

also tested it end-to-end on my cluster

PR #90, opened 2020-10-08
http to HTTP from eventing

n3wscott:
Follow up to knative/eventing#4263
deps on knative/eventing#4265


matzew:
Thanks, Scott!
looks like some tests failed.

I did similar update w/ #91

/close

PR #100, opened 2020-10-13
Allow consumer group to be cancelled

lionelvillard:
## Proposed Changes

- 🐛 KafkaSource now properly close the session when the parent context is closed. 

**Release Note**

```release-note
🐛 KafkaSource now properly close the session when the parent context is closed.  
```

**Docs**

### DIFF
--- a/pkg/channel/consolidated/dispatcher/dispatcher.go
+++ b/pkg/channel/consolidated/dispatcher/dispatcher.go
@@ -344,7 +344,7 @@ func (d *KafkaDispatcher) subscribe(channelRef eventingchannels.ChannelReference
 
 	handler := &consumerMessageHandler{d.logger, sub, d.dispatcher}
 
-	consumerGroup, err := d.kafkaConsumerFactory.StartConsumerGroup(groupID, []string{topicName}, d.logger, handler)
+	consumerGroup, err := d.kafkaConsumerFactory.StartConsumerGroup(context.Background(), groupID, []string{topicName}, d.logger, handler)
 
 	if err != nil {
 		// we can not create a consumer - logging that, with reason
--- a/pkg/common/consumer/consumer_handler.go
+++ b/pkg/common/consumer/consumer_handler.go
@@ -74,14 +74,6 @@ func (consumer *SaramaConsumerHandler) ConsumeClaim(session sarama.ConsumerGroup
 	// https://github.com/Shopify/sarama/blob/master/consumer_group.go#L27-L29
 	for message := range claim.Messages() {
 
-		// Exit if the session has been cancelled due to either rebalancing
-		// or adapter being shutting down
-		select {
-		case <-session.Context().Done():
-			return nil
-		default:
-		}
-
 		if ce := consumer.logger.Desugar().Check(zap.DebugLevel, "debugging"); ce != nil {
 			consumer.logger.Debugw("Message claimed", zap.String("topic", message.Topic), zap.Binary("value", message.Value))
 		}
--- a/pkg/common/consumer/consumer_handler.go
+++ b/pkg/common/consumer/consumer_handler.go
@@ -73,6 +73,15 @@ func (consumer *SaramaConsumerHandler) ConsumeClaim(session sarama.ConsumerGroup
 	// The `ConsumeClaim` itself is called within a goroutine, see:
 	// https://github.com/Shopify/sarama/blob/master/consumer_group.go#L27-L29
 	for message := range claim.Messages() {
+
+		// Exit if the session has been cancelled due to either rebalancing
+		// or adapter being shutting down
+		select {
+		case <-session.Context().Done():
+			return nil
+		default:
+		}
+
 		if ce := consumer.logger.Desugar().Check(zap.DebugLevel, "debugging"); ce != nil {
 			consumer.logger.Debugw("Message claimed", zap.String("topic", message.Topic), zap.Binary("value", message.Value))
 		}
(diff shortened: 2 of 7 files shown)


pkg/source/adapter/adapter.go
```diff
@@ -95,10 +91,13 @@ func (a *Adapter) start(stopCh <-chan struct{}) error {
 	if err != nil {
 		return fmt.Errorf("failed to create the config: %w", err)
 	}
-	config.Consumer.Offsets.AutoCommit.Enable = false
+
+	// Autocommit offset marked as processed. This is usually faster
+	// than committing after each message has been processed
+	config.Consumer.Offsets.AutoCommit.Enable = true
```
matzew, line 97: @pierDipi I think you originally were setting this to `false`

https://github.com/slinkydeveloper/eventing-contrib/commit/a667d57d8b5b04ac5153e22db45a36497b4abbc6

pierDipi: I think it's safe to enable auto-commit since sarama keeps track of the latest marked offset and it commits that at specified intervals, so we should keep at-least-once delivery semantics anyway.
Although this increases the probability of processing messages multiple times, so it depends on what we want to optimize.

matzew: How about... making this _configurable_ ?  As the setting comes with Pros/Cons :man_shrugging:

lionelvillard: yes this should be configurable, not only AutoCommit, but everything else (?) relevant to KafkaSource. However the current implementation does not commit offsets at all so I just pick one policy.

pkg/channel/consolidated/dispatcher/dispatcher.go
```diff
@@ -328,7 +328,7 @@ func (d *KafkaDispatcher) subscribe(channelRef eventingchannels.ChannelReference
 
 	handler := &consumerMessageHandler{d.logger, sub, d.dispatcher}
 
-	consumerGroup, err := d.kafkaConsumerFactory.StartConsumerGroup(groupID, []string{topicName}, d.logger, handler)
+	consumerGroup, err := d.kafkaConsumerFactory.StartConsumerGroup(context.Background(), groupID, []string{topicName}, d.logger, handler)
```
pierDipi, line 331: Can we use a "real" context?

lionelvillard: yes we should. I'm opening an issue to track it as I don't know enough the dispatcher to fix this.

lionelvillard: here: https://github.com/knative-sandbox/eventing-kafka/issues/105

lionelvillard:
@pierDipi @matzew I believe this one is ready to go, right?

pierDipi (approved):
/approve
/cc slinkydeveloper matzew

pkg/common/consumer/consumer_handler.go
```diff
@@ -73,6 +73,15 @@ func (consumer *SaramaConsumerHandler) ConsumeClaim(session sarama.ConsumerGroup
 	// The `ConsumeClaim` itself is called within a goroutine, see:
 	// https://github.com/Shopify/sarama/blob/master/consumer_group.go#L27-L29
 	for message := range claim.Messages() {
+
+		// Exit if the session has been cancelled due to either rebalancing
+		// or adapter being shutting down
+		select {
+		case <-session.Context().Done():
```
slinkydeveloper, line 80: You don't need to add this code, you should invoke `consumerGroup.Close()`. sarama closes the`claim.Messages()` when `consumerGroup.Close()` is invoked

pkg/source/adapter/adapter.go
slinkydeveloper: Do we really need to do this change in this PR?

pkg/common/consumer/consumer_handler.go
slinkydeveloper: It's up to the `consumerGroup` owner to listen on a context and invoke `consumerGroup.Close()` when a specific context is finished

slinkydeveloper (changes requested):

pkg/source/adapter/adapter.go
matzew: yeah, I was wondering the same.

@lionelvillard for the configruation enhancements: can you file an issue?

pierDipi: It's a bug fix we don't commit the offset at all, so it's better than nothing.

pkg/common/consumer/consumer_handler.go
lionelvillard: ok cool thanks! I'll look into this.

pkg/source/adapter/adapter.go
lionelvillard: @matzew done: https://github.com/knative-sandbox/eventing-kafka/issues/108

pkg/common/consumer/consumer_handler.go
lionelvillard: I removed it.

pkg/source/adapter/adapter.go
lionelvillard: done in a separate PR.

lionelvillard:
I made this PR smaller.

/unhold

slinkydeveloper:
So at this point this PR just adds the context to the signature of `StartConsumerGroup` right?

lionelvillard:
right.

slinkydeveloper:
So the description of the pr is wrong :smile:

lionelvillard:
description or title?

slinkydeveloper:
both

slinkydeveloper:
i mean, what if you fix the whole issue here? This change is small enough and pretty clear

lionelvillard:
will submit another PR.

PR #103, opened 2020-10-13
Move constants to common for clean dependency graph.

travis-minke-sap:
The KafkaChannel `pkg/apis/` content has a couple of dependencies into the `pkg/channel/consolidated/` implementation which results in a circular dependency chain.  This is not causing a problem (build or otherwise), but is just generally wrong and should be de-tangled.  It's only a couple constant values that need to be moved - just basic cleanup : )


pkg/common/constants/constants.go
```diff
@@ -0,0 +1,8 @@
+package constants
```
matzew, line 1: let's add the license header

travis-minke-sap: agreed - will have to get used to that (setup my intellij file creation template ; )

travis-minke-sap:
/test pull-knative-sandbox-eventing-kafka-integration-tests

grrr... flaky integration tests...

PR #104, opened 2020-10-14
running misspell

matzew:

Fixes #

## Proposed Changes

* running [misspell](https://github.com/client9/misspell)

```
export FILES=( $(find . -type f -not -path './vendor/*' -not -path './third_party/*' -not -path './.git/*') )
misspell -w "${FILES[@]}"
```


pierDipi (approved):
We should have an action for this.
/lgtm

PR #97, opened 2020-10-13
Update for channel metrics

skonto:
Fixes [downstram build](https://github.com/knative/eventing/pull/4259) for channel metrics support.

## Proposed Changes

- Updates channel API calls that support a stat reporter. This is useful in order to expose Kafka channel metrics in the future.

**Release Note**

```release-note

```

**Docs**

/cc @slinkydeveloper @matzew

PR #115, opened 2020-10-20
Updates

matzew:
Produced via:
```shell
./hack/update-deps.sh --upgrade
./hack/update-codegen.sh
```
/assign @slinkydeveloper @travis-minke-sap @pierDipi


matzew:
@skonto I added your #97 to here

can you look as well ?

PR #128, opened 2020-10-21
DeepCopy the entire set of conditions

### LINKED ISSUE #127: [KafkaSource] Conversion webhook drops conditions
**Describe the bug**
The conversion webhook from v1beta1 to v1alpha1 (stored version) does not preserve conditions, leading to infinite reconciling loop

 **Expected behavior**
All conditions should be carried over.

 **To Reproduce**
Install a v1beta1 KafkaSource and dump the logs

 **Knative release version**
head

 **Additional context**

lionelvillard:
Fixes #127

## Proposed Changes

-
-
-

**Release Note**

```release-note
🐛 v1beta1 KafkaSource conditions are now fully preserved 
```

**Docs**

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/ghodss/yaml v1.0.0
 	github.com/golang/protobuf v1.4.2
 	github.com/google/go-cmp v0.5.2
+	github.com/google/gofuzz v1.1.0
 	github.com/google/uuid v1.1.1
 	github.com/influxdata/tdigest v0.0.1 // indirect
 	github.com/kelseyhightower/envconfig v1.4.0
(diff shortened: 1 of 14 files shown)


pierDipi (commented):
Can you add a regression test?

vaikas:
I'd suggest adding fuzzer tests which will catch things like these among other things.
Something like this:
https://github.com/knative/eventing/pull/3574

And in particular here's a smaller set of changes you will need:
https://github.com/knative/eventing/blob/master/pkg/apis/eventing/v1/fuzzer.go
https://github.com/knative/eventing/blob/master/pkg/apis/eventing/v1/roundtrip_test.go

lionelvillard:
Thanks @vaikas, I added the fuzzer tests. I also changed the v1alpha1 api shape to support round tripping `CloudEventOverrides`. I believe that's how we did it in the past, LMK otherwise.

lionelvillard:
Flake?

/test pull-knative-sandbox-eventing-kafka-integration-tests

vaikas:
Sorry to chime in late here, but I wonder if we should be inlining the Source duck here for the API types?
https://github.com/knative/pkg/blob/master/apis/duck/v1/source_types.go

I'm fine with doing it as a follow up (maybe create an issue if we think it's worth doing)? Sorry, wasn't thinking about it earlier.
/lgtm
/approve
/hold

holding in case you want to inline now. @n3wscott thoughts?

vaikas:
aaaand, if I would have been smarter, I would've looked and it already does use it. Sigh... long day...
/unhold

PR #132, opened 2020-10-22
Use CRD v1 API for KafkaChannel, KafkaSource and KafkaBinding

pierDipi:
CRD v1beta1 API is deprecated and will be removed in [k8s 1.22](https://github.com/kubernetes/kubernetes/issues/82022).
Similar to https://github.com/knative/eventing/pull/3360.
Webhook API bump in a follow-up.

## Proposed Changes

- Use CRD v1 API for KafkaChannel, KafkaSource and KafkaBinding

**Release Note**

```release-note
- 🎁 Add new feature
KafkaChannel CustomResourceDefinition now uses apiextensions.k8s.io/v1 APIs
```

/cc slinkydeveloper matzew


matzew:
/hold

we need upgrade tests for that 

I think @aliok has some reference GH issue for this, what all needs to be done

pierDipi:
@matzew, would you mind helping me understand why? 
To clarify, I'm changing the CRD apiVersion (apiextensions.k8s.io/v1), not adding the KafkaChannel v1.

matzew:
If we tackle this - can we also tackle the source ? 


Does it work when applying the `V1` based version on-top of a `v1beta` version?  and than "rolling back from V1 to beta` ? 

I think when I did that for "eventing" repo, the tests there were useful

matzew:
@pierDipi  Getting back to this.

We we structure those CRDs, like we have them on "eventing" repo ? 

Like:
* have the versions after the `group` (see https://github.com/matzew/eventing/blob/3974f6c67372dbc02c2fb2a0c024390ea5725553/config/channels/in-memory-channel/300-in-memory-channel.yaml#L25-L56)
* have `names`, `scope` at the end, before the `conversion` 

Can we also have the same lifting for the binding, the source and the distributed channel?

matzew:
Does that make sense @pierDipi  ?

config/channel/consolidated/resources/kafka-channel.yaml
```diff
@@ -24,52 +24,58 @@ metadata:
...
+          << : *openAPIV3Schema
   conversion:
     strategy: Webhook
-    conversionReviewVersions: ["v1beta1", "v1alpha1"]
-    webhookClientConfig:
-      service:
-        name: kafka-webhook
-        namespace: knative-eventing
+    webhook:
+      conversionReviewVersions: [ "v1", "v1beta1" ]
```
aliok, line 77: @pierDipi 
Please only change the CRD to use Kube's v1 ApiExtensions. 

Bumping the version of `messaging.knative.dev` to `v1` is a separate work.

pierDipi: This refers to the conversion request version.

pierDipi:
@matzew it makes sense to me.

config/channel/consolidated/resources/kafka-channel.yaml
pierDipi: From: https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/
```
# conversionReviewVersions indicates what ConversionReview versions are understood/preferred by the webhook.
# The first version in the list understood by the API server is sent to the webhook.
# The webhook must respond with a ConversionReview object in the same version it received.
```

aliok: Ok, so, not related to KafkaChannel's version (messaging.knative.dev)?

pierDipi: Yes, it's unrelated, see https://github.com/knative/eventing/blob/master/config/core/resources/eventtype.yaml, `EventType` is in `v1alpha1` and `v1beta1` and the conversionReviewVersions is set to [`[ "v1", "v1beta1" ]`](https://github.com/knative/eventing/blob/dccb17e10b6bc04e51f0d6f1a871f4676fe89e2f/config/core/resources/eventtype.yaml#L158)

aliok: ha, ok.

matzew: Isn't that the conversion of the CRD api itself ?

pierDipi: Nope, they are versions of the conversion request and response, see https://github.com/kubernetes/apiextensions-apiserver/blob/1fa7e1a40da83d381ce598c1dc4095debc040c60/pkg/apis/apiextensions/v1/types.go#L434-L476 and https://github.com/kubernetes/apiextensions-apiserver/blob/1fa7e1a40da83d381ce598c1dc4095debc040c60/pkg/apis/apiextensions/v1beta1/types.go#L475-L522

`v1alpha1` version has never been created, (first CRDs release branch) https://github.com/kubernetes/apiextensions-apiserver/tree/release-1.7/pkg/apis/apiextensions

matzew:
@pierDipi I think this needs a rebase

pierDipi:
Flaky, I opened an issue: https://github.com/knative-sandbox/eventing-kafka/issues/218
/retest

matzew:
VEry cool!

travis-minke-sap:
Seem fine to me!   I had the same questions regarding `conversionReviewVersions` and so learned something from the conversation above ; )

PR #136, opened 2020-10-22
Update azure eventhub lib

travis-minke-sap:
Historically the "distributed" KafkaChannel implementation used `dep` for dependency management.  This made it difficult to use the latest versions of the Azure EventHubs SDK as they had moved to Go Modules.  We have now been using Go Modules for some time and it would be good to update this dependency to the latest version.

While we would like to do this just to stay current, the importance has been raised due to failures in downstream checks from the `pkg` repo....
https://github.com/knative/pkg/pull/1832/checks?check_run_id=1290109844#step:6:192
...which is being discussed here...
https://knative.slack.com/archives/CR2141CGY/p1603331021072100

The `pkg` failure is not completely understood since the `eventing-kafka` dependencies resolve themselves locally without problem.  This PR is an "attempt" to resolve the issue (since it is something we want to do anyway) but further efforts might be required.

The are only a few minor source changes (updated import paths) and most of the PR is updated vendor/ content.  Multiple commits were made to separate these items for easier review ; )


markusthoemmes (approved):
/lgtm
/approve

PR #138, opened 2020-10-22
Enable KafkaSource scale subresource

### LINKED ISSUE #137: [KafkaSource] Enable scaling via scale subresource
**Problem**
The eventing autoscaler is currently tightly-coupled with the KafkaSource implementation, making both components difficult to update. Instead the eventing autoscaler should only be interacting with the KafkaSource API

**Exit Criteria**
`k scale` should work on KafkaSource

**Time Estimate (optional):**
~1

https://github.com/knative-sandbox/eventing-autoscaler-keda/issues/26

lionelvillard:
Fixes #137

## Proposed Changes

- Enable KafkaSource scale subresource to scale the underlying deployment
-
-

**Release Note**

```release-note
🎁 The KafkaSource scale subresource can now be used to scale up and down the underlying deployment
```

**Docs**

Depends on https://github.com/knative/pkg/pull/1833 

~I'll remove the dependency on my fork my the PR above is merged (or another PR handling the scale subresource properly)~. Done

FYI @zroubalik. This PR breaks eventing Keda integration. I'm working on a PR to fix it.


config/source/resources/kafkasource.yaml
```diff
@@ -48,6 +48,11 @@ spec:
   scope: Namespaced
   subresources:
     status: {}
+    scale:
+      # specReplicasPath defines the JSONPath inside of a custom resource that corresponds to Scale.Spec.Replicas.
+      specReplicasPath: .spec.replicas
+      # statusReplicasPath defines the JSONPath inside of a custom resource that corresponds to Scale.Status.Replicas.
+      statusReplicasPath: .status.replicas
```
zroubalik, line 55: Are `replicas` defined on `KafkaSourceStatus` ?

lionelvillard: Not yet. Good catch!!

lionelvillard:
/hold

Let's wait for the scheduled dependency upgrade to pick up knative/pkg#1833

lionelvillard:
@zroubalik @matzew do you mind looking at this one?

pkg/apis/sources/v1alpha1/kafka_types.go
```diff
@@ -121,6 +127,11 @@ type KafkaSourceStatus struct {
 	// * SinkURI - the current active sink URI that has been configured for the
 	//   Source.
 	duckv1.SourceStatus `json:",inline"`
+
+	// Total number of non-terminated replicas targeted by this source
+	// +optional
+	// For round-tripping only
+	Replicas int32 `json:"replicas,omitempty"`
```
matzew, line 134: Is that causing some "migration" problems?

I know we do lack "upgrade tests" for the kafka stuff

pkg/apis/sources/v1beta1/kafka_types.go
```diff
@@ -67,6 +67,11 @@ type KafkaResourceSpec struct {
 
 // KafkaSourceSpec defines the desired state of the KafkaSource.
 type KafkaSourceSpec struct {
+	// Number of desired replicas. This is a pointer to distinguish between explicit
+	// zero and not specified. Defaults to 1.
+	// +optional
+	Replicas *int32 `json:"replicas,omitempty"`
```
matzew, line 73: Doing it on `v1`? instead ? :grimacing:

lionelvillard: We should allow adding new fields to an existing API without changing the version. Kubernetes allows it. See https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#adding-unstable-features-to-stable-versions

Excerpt: `.. It is generally allowed to add new fields without changing the API version, ..`

pkg/apis/sources/v1alpha1/kafka_types.go
lionelvillard: I don't think so since this is a new field

zroubalik: Does it make a sense to move this to `duckv1.SourceStatus`?

lionelvillard: I think so, as an optional field. From a conformance PoV, a source *may* support horizontal scaling. The counter argument is we tend to only declare fields that are used. Not all sources supports `replicas`. 

If we decide to add it to `duckv1.SourceStatus` later on, it won't be breaking the API so I suggest we leave it this way for now.

slinkydeveloper (commented):
Does that field makes sense to exist when `KafkaSource` will be implemented with eventing-kafka-broker data plane? @pierDipi 

My feeling (i might be wrong) is that it kinda creates a "stronger" bond between the implementation, which today single tenant (vs an eventual one with eventing-kafka-broker **multi-tenant**). What this `replicas` means in a multi tenant impl?

pierDipi:
I guess it means multiple consumers with the same consumer group.

slinkydeveloper:
> I guess it means multiple consumers with the same consumer group.

Technically a single instance can have multiple consumers in the same consumer group, in java this is done automatically by the kafka client

slinkydeveloper:
What I'm trying to say is that my feeling is that this field leaks an implementation detail of this particular implementation of `KafkaSource`

pierDipi:
Can you explain a little bit more? 
I mean in a multi-tenant scenario replicas means run N consumers with the same consumer group where they run I don't care.

slinkydeveloper:
If `replicas` means `consumers`, then why naming it `replicas`? My understanding is that this proposal wants to add `replicas`, in the meaning of the kubernetes replicas for a specific deployment

lionelvillard:
The actual definition of `replicas` depends on the implementation. For the single-tenant KafkaSource, it really means N pods replicas. For a simple multi-tenant KafkaSource implementation it means run N consumers (by lowering throttling). 

Maybe `replicas` is not the right name. In the design document I call it `virtual replicas`.

slinkydeveloper:
Then what about `consumers` to map with the concepts of kafka?

lionelvillard:
+1 for `consumers`. 

The only caveat is a multi tenant source *might* provide more *virtual* consumers than the number of partitions, potentially leading to some confusion (source says 10 consumers but there is only 5 partitions in the topic). The documentation should leave the door open to such implementation.

lionelvillard:
@travis-minke-sap @eric-sap do you have any comments/objections on this PR? (sorry I meant to ping you earlier on this one).

pierDipi:
It seems that more consumers than the number of partitions are allowed by this implementation as well.

lionelvillard:
> It seems that more consumers than the number of partitions are allowed by this implementation as well.

that's right. Currently only KEDA makes sure that the number of consumers does not go over. This is a problem especially if consumers are virtual consumers. I'm working on a PoC right now to give more meat to the discussion.

pkg/apis/sources/v1alpha1/kafka_types.go
```diff
@@ -95,6 +95,17 @@ type KafkaSourceSpec struct {
 	// +optional
 	// Needed for supporting round-tripping
 	CloudEventOverrides *duckv1.CloudEventOverrides `json:"ceOverrides,omitempty"`
+
+	// Number of desired consumers. Defaults to 1.
+	//
+	// A consumer is an unit for measuring resource allocation,
+	// for example the number of pods, or the allocated bandwidth.
```
slinkydeveloper, line 102: The allocated bandwidth?

What about "the number of consumers running in the consumer group"?

pkg/apis/sources/v1alpha1/kafka_types.go
```diff
@@ -121,6 +132,11 @@ type KafkaSourceStatus struct {
 	// * SinkURI - the current active sink URI that has been configured for the
 	//   Source.
 	duckv1.SourceStatus `json:",inline"`
+
+	// Total number of consumers targeted by this source
+	// +optional
+	// For round-tripping only.
+	Consumers int32 `json:"consumers,omitempty"`
```
slinkydeveloper, line 139: Why do we need this? Isn't the one in the spec enough?

lionelvillard: yes it's a required field for enabling the scale subresource. See https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#scale-subresource

pkg/apis/sources/v1alpha1/kafka_types.go
lionelvillard: vaguer is a better :-) More consumers = faster, that's the only thing we need to know at the level.

slinkydeveloper: +1 for the wording you just committed

lionelvillard:
@slinkydeveloper @matzew can we merge this one now?

slinkydeveloper:
Fine for me

/lgtm
/approve

lionelvillard:
only fine?! :-)

Issue #24, opened 2020-06-10
CloudEvent Tracing Compliance


travis-minke-sap:
The Knative Channel Specification specifies the expected behavior of handling W3C Tracecontext headers. The current implementation does NOT meet these requirements in that the traceparent and tracestate headers are lost in transmission. A completely new traceparent value is instead sent to the subscriber.

Currently, the Knative In-Memory Channel reference implementation DOES adhere to the specification, but the eventing-contrib/kafka implementation does not. Similar to this implementation the latter has the intermediate step of persisting to/from a Kafka Topic. There is currently confusion and a debate around the best way to handle this persistence, how it relates to the potential use of ce-traceparent and ce-tracestate, and where in the Knative / CloudEvent stack to handle this. Further, it is unclear whether the CloudEvents DTE Extension is useful or not.

See the following Github Issues/PRs for the history...

Knative / CloudEvents
knative/eventing#2918 & cloudevents/spec#603 & cloudevents/spec#607
eventing/in-memory
knative/eventing#1757
eventing-contrib/kafka
knative/eventing-contrib#1139 & knative/eventing-contrib#1155
knative/eventing-contrib#563
Previous Knative Implementation Removal
knative/eventing#2873
Therefore, we are going to wait until this is resolved before attempting to implement a solution that complies with the expected specification.

There are also larger tracing issue to be dealt with within the Knative Eventing ecosystem such as how to integrate Istio B3 Header usage with the current OpenTelemetry Headers. See knative/eventing#2581.

W3C Tracecontext Notes
The traceparent field is composed of the following values...

<version> - <traceid> - <parentid> - <flags>

...where the <version> (usually 00) specifies the format of the following data fields, and the <traceid> is the end-to-end ID of the entire trace through a distributed system. The <parentid> is also referred to as the <spanid>, and can/should change with each new HTTP Client Request being made. Therefore the expectation is that when the Channel receives an HTTP request with a traceparent header, that it is forwarded to the Subscriber with the same traceparent header, albeit with a new <parentid>.

eric-sap:
The tracing has been enabled for the channel and dispatcher (via #30) by calling tracing.SetupDynamicPublishing() inside an InitializeTracing() function called from their respective main() functions.

This currently results in traces that have a different TraceID for the channel and dispatcher tracing spans, which might be a bit confusing when looking through the results (zipkin, etc) for what is a single event.  Fixing the Tracecontext headers may alleviate this.

[...]

PR #140, opened 2020-10-23
Add distributed trace headers

eric-sap:
This fix is related to issue #24 in that it adds the traceparent header that is necessary for the conformance tests to execute successfully for the distributed channel (those tests will be added in a future PR, hopefully shortly).

It may not "fix" #24 as there may be other things that need to be done to be fully compliant with tracing requirements; this will require a bit of research.


pkg/channel/distributed/receiver/producer/producer.go
```diff
@@ -117,6 +120,8 @@ func (p *Producer) ProduceKafkaMessage(ctx context.Context, channelReference eve
 		return err
 	}
 
+	producerMessage.Headers = append(producerMessage.Headers, commonutil.SerializeTrace(trace.FromContext(ctx).SpanContext())...)
+
```
travis-minke-sap, line 124: Can we add a comment here to make clear what is happening...

eric-sap: Absolutely

travis-minke-sap:
@slinkydeveloper and or @matzew  - Would appreciate your eyes on this one since it's tracing related and involves a move of some "consolidated" channel code to "common".   thanks! 

/hold
/lgtm
/assign
/approve

pkg/channel/distributed/common/config/observability_test.go
```diff
@@ -41,7 +43,7 @@ func TestInitializeObservability(t *testing.T) {
 
 	// Test Data
 	ctx := context.TODO()
-	metricsPort := 9876
+	metricsPort := rand.Intn(10000) + 9000
```
matzew, line 46: was there a test issue?

pkg/channel/distributed/dispatcher/dispatcher/handler.go
```diff
@@ -175,7 +176,7 @@ func (h *Handler) checkRetry(_ context.Context, response *http.Response, err err
 	//        status codes from the subscriber and returning 400s instead.  Once this has,
 	//        been resolved we can remove 400 from the list of codes to retry.
 	//
-	if statusCode >= 500 || statusCode == 400 || statusCode == 404 || statusCode == 429 {
+	if statusCode >= 500 || statusCode == 400 || statusCode == 404 || statusCode == 429 || statusCode == 409 {
```
matzew, line 179: was there some `Conflict` ?

travis-minke-sap: Historically some of these codes were to workaround broker/filter "bugs" and I believe the new "409" is because that is what the eventing retry e2e test is sending and expecting a retry to happen - Eric can confirm.

pkg/channel/distributed/common/config/observability_test.go
eric-sap: Yes, occasionally this test and the TestMetricsServer_Report would run at exactly the same time and one or the other would fail.  It was somewhat rare but I did run into it more than once.

pkg/channel/distributed/dispatcher/dispatcher/handler.go
eric-sap: The TestBrokerRedelivery fails for the distributed channel because eventing/test/test_images/recordevents returns http.StatusConflict (409) when given a skip algorithm.  The distributed e2e tests are not quite finished yet for some other reasons but this file was already being changed so I left this pre-emptive fix in place.  It can be moved to the e2e PR if that would be cleaner.

pkg/channel/distributed/common/config/observability_test.go
eric-sap: They could also be two unique constant ports, but since I suspect this issue arose by one test being copied from another, I was hoping to stave that off in the future.  I am open to that change instead, however.

pkg/channel/distributed/dispatcher/dispatcher/handler.go
slinkydeveloper: @eric-sap I would prefer if you move this particular change to the e2e pr

slinkydeveloper:
Thanks for this pr @eric-sap !

/lgtm
/approve

/hold 
for another reviewer

eric-sap:
Okay, I've moved the status code out of this branch.

travis-minke-sap:
/lgtm
/unhold

*the LGTM analysis has been running/stuck for over 1.5hrs now so merging ahead without it.  PR previously passed this evaluation multiple times.

PR #164, opened 2020-11-04
Move ConfigKafkaVersionDefault around

matzew:

Fixes #

## Proposed Changes

- more code moving around, to evetually share...
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/common/constants/constants.go
+++ b/pkg/common/constants/constants.go
@@ -16,6 +16,8 @@ limitations under the License.
 
 package constants
 
+import "github.com/Shopify/sarama"
+
 const (
 
 	// KafkaChannel Spec Defaults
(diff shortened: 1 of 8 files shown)


pkg/common/constants/constants.go
```diff
@@ -25,3 +27,16 @@ const (
...
+	//
+	// Default Kafka Version
+	//
+	// This is the default value which will be used when creating Sarama.Config if not
+	// otherwise specified in the ConfigMap.  It is set to the lowest common denominator
+	// version to provide the most compatible and likely to succeed solution.  Specifically,
+	// Sarama's ConsumerGroups repeatedly close due to EOF failures when working against
+	// Azure EventHubs if this is set any higher than V1_0_0_0.
+	//
+	ConfigKafkaVersionDefault = sarama.V1_0_0_0
```
matzew, line 41: Is that version still a real thing, @eric-sap @travis-minke-sap ?

The "consolidated" is weird, uses 1.x and 2.x ... 

I was thinking.... like:

```
MinConfigKafkaVersionDefault
BestConfigKafkaVersionDefault
```

sure, the names suck... but I'd rather a better mix of versions.

But I have no clue on EventHubs...

matzew:
@travis-minke-sap mind looksie?

pkg/common/constants/constants.go
travis-minke-sap: Hey Matthias - Not sure I understand your question?    We are still using that value in the case where the user didn't provide one in the ConfigMap.  The notes about EventHubs still apply, etc.   But... it sounds fine to refactor the name to `MinKafkaVersion` or  something like that if you need/want another version.  The distributed ConfigMap has it defaulted to **2.0.0** (which is converted into `sarama.V2_0_0_0`)

I can approve/merge now if you want, or if you're going to refactor then let me know to wait for those changes...

matzew: OK, cool - thanks

I will do the refactors on Monday!

PR #170, opened 2020-11-06
E2e remove source from distributed

eric-sap:
## Proposed Changes

- 🧽 Remove the "source" tests from the distributed channel tests, as they are duplicated by the consolidated tests.  A future enhancement to separate the source tests into their own distinct test would probably be good but for now this at least avoids the glaring redundancy.
- 🐛 Increased timeout on channel teardown/cleanup for "ko delete" and added some namespace/kafkachannel logging to help diagnose a sporadic failure of the distributed tests


matzew (approved):
/lgtm
/approve

ah, fun the SRC tests were twice in there :-)

PR #175, opened 2020-11-09
Add defensive coding around reconciling deleted resources.

travis-minke-sap:
We were concerned that the distributed KafkaChannel could potentially hit an extremely rare (if even possible?) race condition whereby it is doing a normal Reconciliation on a resource that exists in K8S but was "being" deleted.  Therefore we'd like to add a small check for the resources Deletion Timestamp and force a re-reconciliation just in case.

> Note - The Dispatcher resources (Service / Deployment) are "cross-namespace" and thus have their own Finalizer in place and will not be modified to force re-reconciliation if the user removes the Finalizer.  Some additional state logging was added to be explicit about the handling of the resource during reconciliation.

### DIFF
--- a/pkg/channel/distributed/controller/testing/data.go
+++ b/pkg/channel/distributed/controller/testing/data.go
@@ -19,7 +19,6 @@ package testing
 import (
 	"fmt"
 	"strconv"
-	"time"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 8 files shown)


pkg/channel/distributed/controller/testing/data.go
```diff
@@ -162,6 +162,18 @@ var (
 type ServiceOption func(service *corev1.Service)
 type DeploymentOption func(service *appsv1.Deployment)
 
+// Set The Service's DeletionTimestamp To Current Time
```
eric-sap, line 165: I'm not sure this matters to the tests, but 1e9 isn't the Current Time (it's sometime in 2001)

travis-minke-sap: Yeah, I had just copied that from down below - you're right though and I'll fix it - good catch!

eric-sap (approved):
Looks solid

PR #176, opened 2020-11-09
Don't crash when KafkaSource dispatcher fails to start consumer group

lionelvillard:
## Proposed Changes

- Loop forever when config is invalid and/or the broker is not yet accessible
-
-

**Release Note**

```release-note
- 🐛 Fix crash when the KafkaSource dispatcher fails to start consumer group
```

**Docs**

### DIFF
--- a/pkg/source/adapter/adapter_test.go
+++ b/pkg/source/adapter/adapter_test.go
@@ -28,7 +28,6 @@ import (
 
 	"github.com/Shopify/sarama"
 	"github.com/cloudevents/sdk-go/v2/types"
-	"github.com/stretchr/testify/require"
 	"go.uber.org/zap"
 	"knative.dev/eventing/pkg/adapter/v2"
 	"knative.dev/pkg/source"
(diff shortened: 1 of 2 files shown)


lionelvillard:
/test pull-knative-sandbox-eventing-kafka-integration-test-channel-consolidated

Flaky?

lionelvillard:
@matzew @slinkydeveloper mind taking a look at it?

slinkydeveloper:
@lionelvillard I might misunderstood the point of this PR, but isn't the crash loop of the pod implemented by k8s enough for that purpose? If kafka is not available, crash. When the crash limit is exceeded, the pod is not restarted and the user can access to logs to get why. With this PR, you have a pod hang forever and sucking resource, but it's running fine for kube (and even the source is marked ready because the pod is running) and user recognize something is definitely wrong just accessing to the logs. 

Also, the kafka source controller check if the cluster is reachable and the topic is there before start accessing to it?

lionelvillard:
> I might misunderstood the point of this PR, but isn't the crash loop of the pod implemented by k8s enough for that purpose?

it is enough (assuming we change the default backoff policy), but it's bad UX.

> Also, the kafka source controller check if the cluster is reachable and the topic is there before start accessing to it?

Not yet. @devguyio common admin library will help

slinkydeveloper:
> it is enough (assuming we change the default backoff policy), but it's bad UX.

I guess it's a matter of taste, IMO it's even worse the pod is always up and running keeping retrying, with the source being ready. This way, the user notices something wrong is going only looking at the pods of the source.

pkg/source/adapter/adapter_test.go
```diff
@@ -441,9 +440,24 @@ func TestAdapter_Start(t *testing.T) { // just increase code coverage
...
 	a := NewAdapter(ctx, NewEnvConfig(), nil, nil)
-	require.Panics(t, func() {
-		_ = a.Start(ctx)
-	})
+
+	stopped := make(chan bool, 1)
+	go func() {
+		err := a.Start(ctx)
+		if err != nil {
+			t.Error("unexpected error:", err)
```
lberk, line 448: What about adding a status field? marking whether or not the consumer group was successfully started or not?

slinkydeveloper: how do you propagate such status back to the controller/crd?

lionelvillard: we can leverage @devguyio work on channel subscriptions for that. The controller does it, not the dispatcher/adapter.

When we get this, this PR becomes irrelevant, so closing

lionelvillard:
Let's do the work in the controller side first.

PR #181, opened 2020-11-10
[Consolidated] Use SyncProducer to provide safer delivery  

lionelvillard:
## Proposed Changes

- Consolidated KafkaChannel dispatcher now handles produced event synchronously 
-
-

**Release Note**

```release-note
- 🧽 The consolidated KafkaChannel now relies by default on SyncProducer for safer event production.
```

**Docs**

### DIFF
--- a/pkg/channel/consolidated/dispatcher/dispatcher.go
+++ b/pkg/channel/consolidated/dispatcher/dispatcher.go
@@ -50,7 +50,7 @@ type KafkaDispatcher struct {
 	receiver   *eventingchannels.MessageReceiver
 	dispatcher *eventingchannels.MessageDispatcherImpl
 
-	kafkaAsyncProducer   sarama.AsyncProducer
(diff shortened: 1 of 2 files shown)


matzew:
could this be configured?

async VS sync has each PROS/CONS

lionelvillard:
yes absolutely!

devguyio:
@lionelvillard what do you mean by "at least once" guarantee on the producer side"? I'm not sure I understand it

lionelvillard:
"at least once" is misleading in this context. I updated the release notes. PTAL @devguyio. Thanks!

slinkydeveloper (commented):
/lgtm

/hold
Sounds good to me but I wonder: is there any way to test the response is 500 when the message cannot be published without mocking the world? If not, remove the hold and let this one go in.

lionelvillard:
Sarama comes with a mocking API, that could help. It's quite some work to do as a follow up

/unhold

PR #182, opened 2020-11-11
Set subscription readiness based on consumer group status

### LINKED ISSUE #98: [consolidated] Dispatcher deployment readonly
**Problem**
The dispatcher deployment updates the status of the KafkaChannel and every update needs to be done by the leader and replicas don't receive informer updates, since we use only `ReconcileKind`.
This behavior [breaks dispatcher horizontal scalability](https://github.com/knative/eventing-contrib/issues/1556) and if there are multiple replicas of a dispatcher the information that is in the status isn't really accurate since it reflects the status of the leader only.

In https://github.com/knative/eventing-contrib/pull/1446 we had a discussion on how we can solve this problem.

There are other problems related to updating the status in the dispatcher like in https://github.com/knative/eventing/pull/4280.

`*`

**Exit Criteria**
One of the solutions outlined in the linked discussion.

**Time Estimate (optional):**
1

- Discussion: https://github.com/knative/eventing-contrib/pull/1446#discussion_r469492807

### LINKED ISSUE #134: Consolidated KafkaChannel marks Subscriptions Ready too early
**Describe the bug**
[The Channel spec](https://github.com/knative/eventing/blob/master/docs/spec/channel.md) requires

> Channels MUST send events to all subscribers which are marked with a status of ready: "True" in the channel's status.subscribableStatus.subscribers (v1alpha1) or status.subscribers (v1beta1).

There is, however, a delay between the subscribers status being ready in the KafkaChannel status, and the startup of the partition consumer for the topic.

Any events sent to the channel during the window are not being sent to subscribers.

 **Expected behavior**
kafka-channel-dispatcher should not update the subscribers status if it isn't actually ready to send events to subscribers, so it doesn't violate the spec.

 **Additional context**
The KafkaDispatcher logic is

1. Create a sarama ConsumerGroup
    1. Start a `Consume()` go routine (async)
1. Errors? no-> mark subscription ready

The Consumer connects and joins the Kafka ConsumerGroup in step 2 which is async and can happen indeed after step 3. is executed.

**Solution idea** 
To have a status for the ConsumerGroup (ours, not Sarama) and set it to Ready once the actual join occurs (Sarama has a `Setup()` lifecycle callback that will be called on a successful join) and consequently mark the subscription ready only when the CG is ready

devguyio:
Fixes #134
Fixes #98 
Fixes knative/eventing-contrib#1556

## Proposed Changes

- Set subscription readiness based on consumer group status in the consolidated Kafka
- Set's KafkaChannel consolidated dispatcher to read-only
- Enables HA in KafkaChannel consolidated dispatcher

**Release Note**

```release-note
- A bug was fixed in the consolidated KafkaChannel where subscriptions would show up in the channel's `status.subscribers` before the dispatcher becomes ready to dispatch messages for those subscribers.
- Consolidated KafkaChannel dispatcher's horizontal scalability works now seamlessly with reconciler leader election.
```


pkg/channel/consolidated/reconciler/controller/kafkachannel.go
```diff
@@ -248,6 +261,76 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, kc *v1beta1.KafkaChannel
 	return newReconciledNormal(kc.Namespace, kc.Name)
 }
 
+func (r *Reconciler) setupSubscriptionStatusWatchers(ctx context.Context, channel *v1beta1.KafkaChannel) error {
+	var err error
+	if channel.Spec.SubscribableSpec.Subscribers != nil {
+		for _, s := range channel.Spec.SubscribableSpec.Subscribers {
+			groupID := fmt.Sprintf("kafka.%s.%s.%s", channel.Namespace, channel.Name, string(s.UID))
+			err = r.kafkaWatcher.WatchCosumerGroup(groupID, func(event CGEvent) {
```
lionelvillard, line 269: nit:  WatchCosumerGroup -> .WatchConsumerGroup

pkg/channel/consolidated/reconciler/controller/kafkawatcher.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (w *KafkaWatcher) Stop() error {
+	//TODO handle if closed
+	close(w.closed)
+	return nil
+}
+
+// TODO Add batch notification
+func (w *KafkaWatcher) WatchCosumerGroup(cgname string, cb CGCallBack) error {
```
lionelvillard, line 115: same nit above

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
```diff
@@ -248,6 +261,76 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, kc *v1beta1.KafkaChannel
 	return newReconciledNormal(kc.Namespace, kc.Name)
 }
 
+func (r *Reconciler) setupSubscriptionStatusWatchers(ctx context.Context, channel *v1beta1.KafkaChannel) error {
+	var err error
+	if channel.Spec.SubscribableSpec.Subscribers != nil {
+		for _, s := range channel.Spec.SubscribableSpec.Subscribers {
+			groupID := fmt.Sprintf("kafka.%s.%s.%s", channel.Namespace, channel.Name, string(s.UID))
```
lionelvillard, line 268: can you create a function for this and use it in `dispatcher.go` as well?

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
```diff
@@ -248,6 +261,76 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, kc *v1beta1.KafkaChannel
...
+			found = true
+			if !ready {
+				// we need to make the subscription not ready, which means
+				// skip adding it to Status.Subscribers
+				continue
+			}
+		}
+		after.Status.Subscribers = append(after.Status.Subscribers, s)
+	}
+	if !found && ready {
```
lionelvillard, line 304: what happens when a subscription has been deleted? This function will be triggered when the consumer group is purged from Kafka. Adding it again to status.subscriber feels wrong.

devguyio: I refactored that whole part so might be good to check the code again

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
devguyio: I refactored that part, can you check again?

pkg/channel/consolidated/kafka/admin.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+// AdminClientManager manages a ClusterAdmin connection and recreates one when needed
+// it is made to overcome https://github.com/Shopify/sarama/issues/1162
+type AdminClientManager struct {
+	ca sarama.ClusterAdmin
+}
+
+func NewAdminClientManager(ctx context.Context, clientID string,
+	kc utils.KafkaConfig) (AdminClientManager, error) {
```
matzew, line 35: let's have some better formatting here

pkg/channel/consolidated/kafka/admin.go
```diff
@@ -0,0 +1,105 @@
...
+	// ListConsumerGroups Lists the consumer groups
+	ListConsumerGroups() ([]string, error)
+}
+
+// AdminClientManager manages a ClusterAdmin connection and recreates one when needed
+// it is made to overcome https://github.com/Shopify/sarama/issues/1162
+type AdminClientManager struct {
+	logger    *zap.SugaredLogger
+	caFactory ClusterAdminFactory
+	ca        sarama.ClusterAdmin
```
matzew, line 46: let's not call this `ca`, I'd confuse that with _Certificate Authority_ ...

pkg/channel/consolidated/kafka/admin.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+func NewAdminClient(ctx context.Context, caFactory ClusterAdminFactory) (AdminClient, error) {
+	logger := logging.FromContext(ctx)
+	logger.Info("Creating a new AdminClient")
+	kafkaClusterAdmin, err := caFactory()
+	if err != nil {
+		logger.Error("error while creating ClusterAdmin", err)
+		return &AdminClientManager{}, err
```
matzew, line 55: return `nil, err` ?

pkg/channel/consolidated/kafka/admin.go
```diff
@@ -0,0 +1,105 @@
...
+
+	"go.uber.org/zap"
+
+	"github.com/Shopify/sarama"
+	"knative.dev/pkg/logging"
+)
+
+var mutex sync.Mutex
+
+type ClusterAdminFactory func() (sarama.ClusterAdmin, error)
```
matzew, line 34: can we perhaps stash this code else where?

Channel currently uses this for `ClusterAdmin`:
https://github.com/knative-sandbox/eventing-kafka/blob/master/pkg/source/client.go#L131-L142

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
```diff
@@ -530,12 +590,25 @@ func (r *Reconciler) updateKafkaConfig(ctx context.Context, configMap *corev1.Co
 	// Eventually the previous config should be snapshotted to delete Kafka topics
 	r.kafkaConfig = kafkaConfig
 	r.kafkaConfigError = err
+	factory := func() (sarama.ClusterAdmin, error) {
+		return source.MakeAdminClient(controllerAgentName, r.kafkaAuthConfig, kafkaConfig.Brokers)
+	}
+	ac, err := kafka.NewAdminClient(ctx, factory)
```
matzew, line 596: not sure I like the `kafka.NewAdminClient` name, and can't we just pass the functtion to the `NewAdminClient`?

pkg/channel/consolidated/reconciler/dispatcher/kafkachannel.go
```diff
@@ -172,36 +169,13 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, kc *v1beta1.KafkaChannel
...
-			status.Ready = corev1.ConditionFalse
-			status.Message = err.Error()
-		}
-		subscriberStatus = append(subscriberStatus, status)
-	}
-	return eventingduckv1.SubscribableStatus{
-		Subscribers: subscriberStatus,
-	}
-}
-
```
matzew, line 204: good, this goes out!

pkg/channel/consolidated/kafka/admin.go
```diff
@@ -0,0 +1,105 @@
...
+		if err != nil {
+			// skip this attempt
+			continue
+		}
+		cgsMap, err = c.ca.ListConsumerGroups()
+	}
+
+	if r > max {
+		return []string{}, fmt.Errorf("failed to refresh the culster admin and retry: %v", err)
+	}
```
matzew, line 94: can we have some comments here, what's going on - and why?

pkg/channel/consolidated/reconciler/controller/kafkawatcher.go
```diff
@@ -0,0 +1,185 @@
...
+	return w
+}
+
+func (w *KafkaWatcher) Start() error {
+	w.logger.Info("KafkaWatcher starting. Polling for consumer groups", zap.Duration("poll duration", w.pollDuration))
+	go func() {
+		for {
+			select {
+			case <-after(w.pollDuration):
+				cgs, err := w.ac.ListConsumerGroups()
```
matzew, line 90: let's use better names (`consumerGroups`)

pkg/channel/consolidated/reconciler/controller/kafkawatcher.go
```diff
@@ -0,0 +1,185 @@
...
+				// Look for disappeared CGs
+				for _, c := range w.cgCache {
+					if !Find(cgs, c) {
+						// This CG was cached but it's no longer there.
+						w.logger.Debug("Consumer group deleted.",
+							zap.String("consumer group", c))
+						cg = c
+						notify = true
+						break
+					}
```
matzew, line 117: Isn't that mostly same as https://github.com/knative-sandbox/eventing-kafka/pull/182/files#diff-85f097d6979c83d4872e73b80c62c39f59e8ef0cc922b60e251536b141b3822cR99-R106  ?

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
```diff
@@ -251,6 +256,60 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, kc *v1beta1.KafkaChannel
...
+		cg := fmt.Sprintf("kafka.%s.%s.%s", ch.Namespace, ch.Name, s.UID)
+		if Find(cgs, cg) {
+			logging.FromContext(ctx).Debugw("marking subscription", zap.Any("subscription", s))
+			after.Status.Subscribers = append(after.Status.Subscribers, v1.SubscriberStatus{
+				UID:                s.UID,
+				ObservedGeneration: s.Generation,
+				Ready:              corev1.ConditionTrue,
+			})
+		}
+	}
```
matzew, line 289: how does this look when we have a few subscribers?

pkg/channel/consolidated/reconciler/controller/kafkawatcher.go
devguyio: so the first loop iterates over the new state to see which one are new (i.e are not cached), and this other loop iterates over the cached ones to see which ones no longer exist on the new state. does that make sense?

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
devguyio: sorry I'm not sure I understand the question. can you elaborate?

pkg/channel/consolidated/kafka/admin.go
```diff
@@ -0,0 +1,105 @@
...
+		t := int(math.Pow(2, float64(r)) * 100)
+		d := time.Duration(t) * time.Millisecond
+		c.logger.Error("listing consumer group failed. Refreshing the ClusterAdmin and retrying.",
+			zap.Error(err),
+			zap.Duration("retry after", d),
+			zap.Int("Retry attempt", r),
+			zap.Int("Max retries", max),
+		)
+		time.Sleep(d)
+		c.ca, err = c.caFactory()
```
slinkydeveloper, line 83: Why recreating the cluster admin here?

pkg/channel/consolidated/kafka/admin.go
```diff
@@ -0,0 +1,105 @@
...
+
+func (c *AdminClientManager) ListConsumerGroups() ([]string, error) {
+	c.logger.Info("Attempting to list consumer group")
+	mutex.Lock()
+	defer mutex.Unlock()
+	r := 0
+	// This gives us around ~13min of exponential backoff
+	max := 13
+	cgsMap, err := c.ca.ListConsumerGroups()
+	for err != nil && r <= max {
```
slinkydeveloper, line 72: Why retry? AFAIK the sarama client already does retries internally

pkg/channel/consolidated/kafka/admin.go
```diff
@@ -0,0 +1,105 @@
...
+
+	if r > max {
+		return []string{}, fmt.Errorf("failed to refresh the culster admin and retry: %v", err)
+	}
+
+	return keys(cgsMap), nil
+}
+
+func keys(m map[string]string) []string {
+	keys := make([]string, 0, len(m))
```
slinkydeveloper, line 100: Just use `make([]string, len(m))` and then add with with indexes (`keys[i] = k`), it will alloc an array more than a slice (better for perfs and semantically equivalent when you iterate on it to read)

pkg/channel/consolidated/kafka/admin.go
```diff
@@ -0,0 +1,105 @@
...
+			zap.Int("Max retries", max),
+		)
+		time.Sleep(d)
+		c.ca, err = c.caFactory()
+		r += 1
+		if err != nil {
+			// skip this attempt
+			continue
+		}
+		cgsMap, err = c.ca.ListConsumerGroups()
```
slinkydeveloper, line 89: Why locking? Isn't the ca already thread safe?

pkg/channel/consolidated/reconciler/controller/kafkawatcher.go
```diff
@@ -0,0 +1,185 @@
...
+
+var (
+	watchersMtx sync.RWMutex
+	cacheMtx    sync.RWMutex
+	// Hooks into the poll logic for testing
+	after = time.After
+	done  = func() {}
+)
+
+type CGCallBack func()
```
slinkydeveloper, line 38: Maybe rename to `ConsumerGroupHandler`?

pkg/channel/consolidated/reconciler/controller/kafkawatcher.go
```diff
@@ -0,0 +1,185 @@
...
+	// that will be used to identify his callbacks when provided to Forget(watcherID).
+	// Successive calls with the same watcherID will queue the callbacks successfully.
+	//
+	//To ensure this is event-triggered, level-driven,
+	// we don't pass the updates to the callback, instead the observer is expected
+	// to use List() to get the updated list of ConsumerGroups.
+	Watch(watcherID string, callback CGCallBack) error
+
+	// Forget removes all callbacks that correspond to the watcherID
+	Forget(watcherID string)
```
slinkydeveloper, line 57: Do we ever use this? Do we need this?

pkg/channel/consolidated/reconciler/controller/kafkawatcher.go
```diff
@@ -0,0 +1,185 @@
+/*
```
slinkydeveloper, line 1: Can you rename the file to `consumer_group_watcher`?

pkg/channel/consolidated/reconciler/controller/kafkawatcher.go
```diff
@@ -0,0 +1,185 @@
...
+	cacheMtx    sync.RWMutex
+	// Hooks into the poll logic for testing
+	after = time.After
+	done  = func() {}
+)
+
+type CGCallBack func()
+type Matcher func(string) bool
+
+type ConsumerGroupWatcher interface {
```
slinkydeveloper, line 41: Is this the best solution we have? Pinging the kafka admin apis? I wonder if this can cause problems on the long run...

slinkydeveloper: Isn't there any mechanism like a long polling event system?

slinkydeveloper (commented):
Coming late to the party! Sorry for the dumb questions :smile:

pkg/channel/consolidated/kafka/admin.go
devguyio: the purpose of this wrapper is to overcome this known issue https://github.com/Shopify/sarama/issues/1162 which afaik in not solved. Sarama ClusterAdmin client does not have reconnect and the work around is to create a new one.

pkg/channel/consolidated/kafka/admin.go
devguyio: name answer as here https://github.com/knative-sandbox/eventing-kafka/pull/182#discussion_r529459584

pkg/channel/consolidated/kafka/admin.go
devguyio: I was presuming there would be a race between lines 71 and 83. I did a hacky test on my machine and indeed there is

```
WARNING: DATA RACE
Write at 0x00c00022e070 by goroutine 13:
  knative.dev/eventing-kafka/pkg/channel/consolidated/kafka.(*AdminClientManager).ListConsumerGroups()
      /home/abd4lla/Workspace/redhat/src/knative.dev/eventing-kafka/pkg/channel/consolidated/kafka/admin.go:83 +0x6bd
  knative.dev/eventing-kafka/pkg/channel/consolidated/kafka.doList()
      /home/abd4lla/Workspace/redhat/src/knative.dev/eventing-kafka/pkg/channel/consolidated/kafka/admin_test.go:86 +0x48
  knative.dev/eventing-kafka/pkg/channel/consolidated/kafka.TestAdminClient.func2()
      /home/abd4lla/Workspace/redhat/src/knative.dev/eventing-kafka/pkg/channel/consolidated/kafka/admin_test.go:63 +0x50

Previous read at 0x00c00022e070 by goroutine 16:
  knative.dev/eventing-kafka/pkg/channel/consolidated/kafka.(*AdminClientManager).ListConsumerGroups()
      /home/abd4lla/Workspace/redhat/src/knative.dev/eventing-kafka/pkg/channel/consolidated/kafka/admin.go:71 +0xe8
  knative.dev/eventing-kafka/pkg/channel/consolidated/kafka.doList()
      /home/abd4lla/Workspace/redhat/src/knative.dev/eventing-kafka/pkg/channel/consolidated/kafka/admin_test.go:86 +0x48
  knative.dev/eventing-kafka/pkg/channel/consolidated/kafka.TestAdminClient.func2()
      /home/abd4lla/Workspace/redhat/src/knative.dev/eventing-kafka/pkg/channel/consolidated/kafka/admin_test.go:63 +0x50

Goroutine 13 (running) created at:
  knative.dev/eventing-kafka/pkg/channel/consolidated/kafka.TestAdminClient()
      /home/abd4lla/Workspace/redhat/src/knative.dev/eventing-kafka/pkg/channel/consolidated/kafka/admin_test.go:62 +0x2c9
  testing.tRunner()
      /usr/lib/golang/src/testing/testing.go:1127 +0x202
```

pkg/channel/consolidated/reconciler/controller/kafkawatcher.go
devguyio: yeah I'm gonna push some code that uses it when the channel is deleted

pkg/channel/consolidated/reconciler/controller/kafkawatcher.go
devguyio: I'm afraid I'm unaware of any. what are you thinking of? long polling provided by kafka itself for the consumer groups ?

pkg/channel/consolidated/kafka/admin.go
devguyio: as far as I understand and from what I read, make always creates a slice

devguyio: in both cases, it creates a backing array, only difference is that the 3rd arg is the initial capacity, without it, it'll create a zero array. so I guess the version in the pull request is the most efficient, no?

pkg/channel/consolidated/reconciler/controller/kafkawatcher.go
lionelvillard: Since this watcher can also be used by KafkaSource and KafkaBroker to potentially poke the same Kafka cluster, I wonder if we should have a dedicated controller periodically updating etcd, or have a proxy serving as a cache layer and supporting push notifications.

pkg/channel/consolidated/kafka/admin.go
slinkydeveloper: Interesting how there are retries but not in case of disconnection...

pkg/channel/consolidated/kafka/admin.go
slinkydeveloper: ah wow, that's pretty bad

pkg/channel/consolidated/kafka/admin.go
slinkydeveloper: https://stackoverflow.com/questions/38654729/golang-slice-append-vs-assign-performance

devguyio: aaah gotcha, yeah of course, append copies the old slice. I thought you meant create an array not a lice. okay got it. thanks!

pkg/channel/consolidated/reconciler/controller/kafkawatcher.go
slinkydeveloper: > if we should have a dedicated controller periodically updating etcd, or have a proxy serving as a cache layer and supporting push notifications.

I guess it exists and it's called strimzi :laughing: We could get some ideas from them regarding this issue, but it sounds like down that path we'll end up representing kafka resources (consumer groups, topics) as CRDs, virtually "cloning" what they're already doing.

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
devguyio: can you please elaborate more about not liking the name? naming is hard and sometimes it becomes a personal preference, unless I'm breaking some naming convention?

pkg/channel/consolidated/kafka/admin.go
devguyio: where would you rather this code? and why?

matzew: I _think_ in the future we can have a better "admin" package, in a more common way - for both channels.

so, yeah - ignore me here

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
matzew: I'd do:

```golang
	ac, err := kafka.NewAdminClient(ctx, func() (sarama.ClusterAdmin, error) {
		return source.MakeAdminClient(controllerAgentName, r.kafkaAuthConfig, kafkaConfig.Brokers)
	})
```

than I don't like the `kafka.NewAdminClient` name... but yeah... let's leave the name alone  but the above code, I'd inline :shrug:

pkg/channel/consolidated/reconciler/controller/kafkawatcher.go
matzew: @devguyio let's please capture this in the comments ? I am sure... this will likely cause question in the future

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
matzew: the rendered printed status

pkg/channel/consolidated/reconciler/controller/kafkawatcher.go
devguyio: @matzew sure, but I thought I did :D , did you notice the comments before each loop? or were they not clear enough? I also made the var names more descriptive. take another look and let me know if you still think it's not clear enough

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
devguyio: f we have 400 subscriptions, they must all show in the after.Status.Subscribers cause that's what define they are ready to receive events

devguyio:
I created a follow up that we need to add a conformance test here https://github.com/knative-sandbox/eventing-kafka/issues/232

pkg/channel/consolidated/reconciler/controller/kafkachannel_test.go
```diff
@@ -682,7 +695,12 @@ func (ca *mockClusterAdmin) DeleteACL(filter sarama.AclFilter, validateOnly bool
 }
 
 func (ca *mockClusterAdmin) ListConsumerGroups() (map[string]string, error) {
-	return nil, nil
+	cgs := map[string]string{
+		fmt.Sprintf("kafka.%s.%s.%s", kcName, testNS, sub1UID): "consumer",
+		fmt.Sprintf("kafka.%s.%s.%s", kcName, testNS, sub2UID): "consumer",
+	}
+	fmt.Println("ListConsumerGroup-------------------")
```
matzew, line 702: please remove

pkg/channel/consolidated/reconciler/dispatcher/kafkachannel.go
```diff
@@ -117,7 +114,9 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
 		kafkachannelLister:   kafkaChannelInformer.Lister(),
 		kafkachannelInformer: kafkaChannelInformer.Informer(),
 	}
-	r.impl = kafkachannelreconciler.NewImpl(ctx, r)
+	r.impl = kafkachannelreconciler.NewImpl(ctx, r, func(impl *controller.Impl) controller.Options {
+		return controller.Options{SkipStatusUpdates: true}
+	})
```
matzew, line 120: ah. yeah - I recall - similar to IMC

matzew:
I think this will also fix https://github.com/knative-sandbox/eventing-kafka/issues/98   filed by @pierDipi

matzew:
@slinkydeveloper or @pierDipi any comments? Otherwise I'd like to get this in

slinkydeveloper:
/lgtm 
robust work!

pkg/channel/consolidated/reconciler/controller/consumer_group_watcher.go
```diff
@@ -0,0 +1,197 @@
...
+	defer watchersMtx.Unlock()
+	cbs, ok := w.watchers[watcherID]
+	if !ok {
+		cbs = []ConsumerGroupHandler{}
+	}
+	w.watchers[watcherID] = append(cbs, cb)
+	// notify at least once to get the current state
+	cb()
+	return nil
+}
```
pierDipi, line 157: IIUC, we call `Watch` every time we reconcile a channel. 
The registered callback will be called every 2 seconds if something changes, right?

Let's say we have the following events:
- Create channel (callback 1)
- Update channel (callback 2)
- then the loop in `Start` calls notify
- callback 1 and callback 2 are called, and both do the same thing, here: https://github.com/knative-sandbox/eventing-kafka/pull/182/files#diff-8d0207acc30e8507b21472d7c7ec7070441d085b47345fc60e6f53ec601da3fdR266-R311

Do we really need to register multiple callbacks for a given channel?

pkg/channel/consolidated/reconciler/controller/consumer_group_watcher.go
```diff
@@ -0,0 +1,197 @@
...
+
+// TODO explore returning a channel instead of a taking callback
+func (w *WatcherImpl) Watch(watcherID string, cb ConsumerGroupHandler) error {
+	w.logger.Debugw("Adding a new watcher", zap.String("watcherID", watcherID))
+	watchersMtx.Lock()
+	defer watchersMtx.Unlock()
+	cbs, ok := w.watchers[watcherID]
+	if !ok {
+		cbs = []ConsumerGroupHandler{}
```
pierDipi, line 151: nit: we know that the capacity of `cbs` is one.
```suggestion
		cbs = make([]ConsumerGroupHandler, 0, 1)
```

pkg/channel/consolidated/reconciler/controller/consumer_group_watcher.go
```diff
@@ -0,0 +1,197 @@
...
+					if !Find(observedCGs, c) {
+						// This CG was cached but it's no longer there.
+						w.logger.Debugw("Consumer group deleted.",
+							zap.String("consumer group", c))
+						changedGroup = c
+						notify = true
+						break
+					}
+				}
+				if notify {
```
pierDipi, line 123: If a consumer group is coming from a different application, notify (in `Start()`) will always be true, so every watcher will be called. (since `cg` in notify(cg) isn't used.)
I think we need a mechanism to exclude observed consumer groups that are not part of our application.

pkg/channel/consolidated/reconciler/controller/consumer_group_watcher.go
```diff
@@ -0,0 +1,197 @@
...
+	cgs := make([]string, 0)
+	for _, cg := range w.cachedConsumerGroups {
+		if matcher(cg) {
+			cgs = append(cgs, cg)
+		}
+	}
+	return cgs
+}
+
+func (w *WatcherImpl) notify(cg string) {
```
pierDipi, line 177: See comment of unsed `cg`.

pkg/channel/consolidated/reconciler/controller/consumer_group_watcher.go
```diff
@@ -0,0 +1,197 @@
...
+
+	// Terminate instructs the watcher to stop polling and clear the watchers cache
+	Terminate()
+
+	// Watch registers callback on the event of any changes observed
+	// on the consumer groups. watcherID is an arbitrary string the user provides
+	// that will be used to identify his callbacks when provided to Forget(watcherID).
+	// Successive calls with the same watcherID will queue the callbacks successfully.
+	//
+	//To ensure this is event-triggered, level-driven,
```
pierDipi, line 54: nit: 
```suggestion
	// To ensure this is event-triggered, level-driven,
```

pkg/channel/consolidated/reconciler/controller/consumer_group_watcher.go
```diff
@@ -0,0 +1,197 @@
...
+func (w *WatcherImpl) notify(cg string) {
+	watchersMtx.RLock()
+	cacheMtx.RLock()
+	defer watchersMtx.RUnlock()
+	defer cacheMtx.RUnlock()
+
+	for _, cbs := range w.watchers {
+		for _, w := range cbs {
+			w()
+		}
```
pierDipi, line 186: I think we need to select watchers to notify based on `cg`.

pierDipi:
Very nice! I want this to become a library usable by the Broker impl as well!

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
```diff
@@ -251,6 +256,60 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, kc *v1beta1.KafkaChannel
...
+
+	jsonPatch, err := duck.CreatePatch(ch, after)
+	if err != nil {
+		return fmt.Errorf("creating JSON patch: %w", err)
+	}
+	// If there is nothing to patch, we are good, just return.
+	// Empty patch is [], hence we check for that.
+	if len(jsonPatch) == 0 {
+		return nil
+	}
```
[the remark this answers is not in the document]
devguyio: So I've to admit I more or less copied this from knative/eventing#4435 by @vaikas , but my understanding is that he meant that `jsonPatch` type is 
```
type JSONPatch []jsonpatch.JsonPatchOperation
```

hence to we check for its emptiness via `len`. Is the wording confusing or did I get your comment totally wrong?

pkg/channel/consolidated/reconciler/controller/consumer_group_watcher.go
```diff
@@ -0,0 +1,197 @@
...
+			}
+		}
+	}()
+	return nil
+}
+
+func (w *WatcherImpl) Terminate() {
+	w.watchers = nil
+	w.cachedConsumerGroups = nil
+	w.done <- struct{}{}
```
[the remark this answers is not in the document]
devguyio: Correct, I'm afraid the whole `Terminate` and `Forget` logic was a last addition so my guard on synchronization was down, I'll review this and the Forget, good catch!

pkg/channel/consolidated/reconciler/controller/consumer_group_watcher.go
```diff
@@ -0,0 +1,197 @@
...
+	watchersMtx.RLock()
+	cacheMtx.RLock()
+	defer watchersMtx.RUnlock()
+	defer cacheMtx.RUnlock()
+
+	for _, cbs := range w.watchers {
+		for _, w := range cbs {
+			w()
+		}
+	}
```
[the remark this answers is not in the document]
devguyio: nope, that's a leftover from old logic :+1:

pkg/channel/consolidated/reconciler/controller/consumer_group_watcher.go
devguyio: nope.

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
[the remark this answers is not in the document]
pierDipi: I see, I think without that comment is already clear but feel free to keep it.

pkg/channel/consolidated/reconciler/controller/consumer_group_watcher.go
devguyio: so, initially I designed this in a typical observer pattern where `notify` watch/notify can notify a subset of watchers giving them which objects added/deleted based on them watching for specific consumer groups. But that was more of a pure edge-triggered observer. To make it more "level driven" similar to k8s reconciliation, I stopped providing "what" changed, and instead notified of a change, where it's the responsibility of the observer to pull current state via `List` and a matcher.

That said, maybe we can have a filtered watch, `Watch(watcherID, watcher, callback)` where observers provide a matcher, and only modifications passing those would get notified. As long as it is still level driven by not passing what changes and forcing them to query for current state of the world I'm fine by that approach. What do you think? @pierDipi

cmd/channel/consolidated/dispatcher/main.go
```diff
@@ -35,8 +35,5 @@ func main() {
 		ctx = injection.WithNamespaceScope(ctx, ns)
 	}
 
-	// Do not run the dispatcher in leader-election mode
-	ctx = sharedmain.WithHADisabled(ctx)
-
```
lionelvillard, line 40: why?

devguyio: are you suggesting that we keep leader election disabled since this is just a read only reconciler? are there down sides of enabling it and observing how it behaves in this mode? since now we set up the consumers on all replicas with `ObserveKind` and on the leader with `ReconcileKind`

pkg/channel/consolidated/reconciler/controller/consumer_group_watcher.go
pierDipi: Ok, thanks for the clarification. 

> That said, maybe we can have a filtered watch, Watch(watcherID, watcher, callback) where observers provide a matcher, and only modifications passing those would get notified. As long as it is still level driven by not passing what changes and forcing them to query for current state of the world I'm fine by that approach. What do you think?

It makes sense to me, it's mostly an optimization, so I think we can do it later as long as we keep track of it.

pkg/channel/consolidated/reconciler/controller/consumer_group_watcher.go
devguyio: as agreed, will use a matcher on a later PR

devguyio:
@pierDipi @matzew all comments addressed

cmd/channel/consolidated/dispatcher/main.go
pierDipi: The end result is the same since we're doing the same thing in both `ObserveKind` and `ReconcileKind` but we don't really need to have replicas competing to acquire the lock.

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
```diff
@@ -530,15 +590,32 @@ func (r *Reconciler) updateKafkaConfig(ctx context.Context, configMap *corev1.Co
...
 	// in which case the finalizer will never get removed.
-	if kafkaClusterAdmin, err := r.createClient(ctx, kc); err == nil && r.kafkaConfig != nil {
+	if kafkaClusterAdmin, err := r.createClient(); err == nil && r.kafkaConfig != nil {
 		if err := r.deleteTopic(ctx, kc, kafkaClusterAdmin); err != nil {
 			return err
 		}
 	}
+	r.consumerGroupWatcher.Forget(string(kc.ObjectMeta.UID))
 	return newReconciledNormal(kc.Namespace, kc.Name) //ok to remove finalizer
 }
```
pierDipi, line 621: This won't be called on non-leader replicas, so to properly clean up resources we have to implement [`ReadOnlyFinalizer`](https://github.com/knative/pkg/blob/a2394e88959df2ca1900855474ae43d7da0a9046/codegen/cmd/injection-gen/generators/reconciler_reconciler.go#L255) or keep HA disabled and have only `ReconcileKind` and `FinalizeKind`. 
I think the latter solution is better (less code, no wasted resource to acquire locks, etc).

devguyio: yeah after some thinking of @lionelvillard comment, I think so too

devguyio: @pierDipi sorry I got a little confused, this is the controller not the dispatcher. The controller runs in HA with a single leader, and stand-by replicas. We don't need this called on non-leader replicas cause they don't set watchers first place.

pierDipi: I was confused too :)

But I think the comment is valid (in the wrong place) since the dispatcher doesn't implement `Finalizer` at all, and that's a bug that we can address later.

cmd/channel/consolidated/dispatcher/main.go
devguyio: my understanding is that the preference is to have HA enabled, with `ReconcileKind` and `ObserveKind` , over disabling HA and having replicas with `ReconcileKind` since this works as intended. I might be mistaken. In all cases, I'd appreciate if this detail doesn't stop the merging of the PR if it's functionally correct. 

I'm working on implementing a mix of `ReconcileKind` and `ObserveKind` in the dispatcher in a PoC as was suggested here https://github.com/knative/eventing-contrib/pull/1446#discussion_r479407258 , so I need HA eventually anyway.

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
pierDipi: Opened an issue: https://github.com/knative-sandbox/eventing-kafka/issues/245

pierDipi (approved):
/lgtm

cmd/channel/consolidated/dispatcher/main.go
lionelvillard: > we don't really need to have replicas competing to acquire locks.

That. Allowing HA seems confusing to me as there is no need it.

matzew:
@devguyio   can you backport this to the `release-0.19` branch ?

PR #186, opened 2020-11-12
multi-tenant KafkaSource

### LINKED ISSUE #219: Multitenant KafkaSource
**Problem**
There is no multi-tenant KafkaSource to be used in multi-tenant environment. 

**Exit Criteria**

**Time Estimate (optional):**

Feature Track: https://docs.google.com/document/d/1jARXfiD4YrAbLQCMo0GJc2tA6CscDpwZvZKEeccjujM/edit#

lionelvillard:
fix #219

## Proposed Changes

- Provide a configuration for deploying KafkaSource instances into a single MT receive adapter. 
-
-

**Release Note**

```release-note
🎁 Add a new alternative KafkaSource implementation in which a single global StatefulSet handles all KafkaSource instances.
```

**Docs**

Tasks:
- [x] initial KafkaSource MT support
- [x] unit tests (to improve as a follow up)
- [x] Throttling (for fairness)
- [x] Scaling wrt. number of source instances
- [x] scaling wrt. kafka lag 
- [ ] e2e tests (follow up)
- [ ] doc (follow up)


pkg/common/scheduler/statefulset/snapshot.go
```diff
@@ -0,0 +1,55 @@
...
+*/
+
+package statefulset
+
+import "go.uber.org/zap"
+
+// Snapshot provides some information about the current scheduling of all vpods
+type Snapshot struct {
+	// free tracks the free capacity of each pod.
+	free map[string]int32
```
steven0711dong, line 24: we should put a comment on what the key and value is

pkg/common/scheduler/statefulset/autoscaler.go
```diff
@@ -0,0 +1,115 @@
...
+// avgUsedCapacityPerPod returns the average used capacity per replica,
+// including pending virtual replicas
+func (s *StatefulSetScheduler) avgUsedCapacityPerPod(snapshot *Snapshot) float64 {
+	if s.replicas == 0 {
+		if s.pendingReplicas() > 0 {
+			return float64(s.capacity) // full
+		}
+		return float64(s.capacity) / 2 // 50%
+	}
+	return float64(s.usedCapacity(snapshot)) / float64(s.replicas)
```
steven0711dong, line 114: According to this usedCapacity() function. It looks like s.replicas refers to the number of pods. If so, shouldn't it return float64(s.usedCapacity(snapshot)) / float64(s.replicas * s.capacity)?

pkg/common/scheduler/statefulset/scheduler.go
```diff
@@ -0,0 +1,277 @@
...
+	logger.Infow("scheduling successful", zap.Any("placement", placements))
+	delete(s.pending, vpod.GetKey())
+	return placements, nil
+}
+
+func (s *StatefulSetScheduler) reschedule(placements []duckv1alpha1.Placement) []duckv1alpha1.Placement {
+	if s.needRescheduling(placements) {
+		newPlacements := make([]duckv1alpha1.Placement, 0, len(placements))
+		for i := 0; i < len(placements); i++ {
+			if ordinalFromPodName(placements[i].PodName) > s.replicas {
```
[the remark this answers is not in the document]
lionelvillard: yes fixed. Thanks!

pkg/common/consumer/consumer_handler.go
```diff
@@ -73,6 +73,7 @@ func (consumer *SaramaConsumerHandler) ConsumeClaim(session sarama.ConsumerGroup
 	// The `ConsumeClaim` itself is called within a goroutine, see:
 	// https://github.com/Shopify/sarama/blob/master/consumer_group.go#L27-L29
 	for message := range claim.Messages() {
+
```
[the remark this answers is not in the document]
lionelvillard: Sure. Can we do this in a different PR since this is not related to MT specifically

pkg/common/scheduler/statefulset/autoscaler.go
```diff
@@ -0,0 +1,119 @@
...
```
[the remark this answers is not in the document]
lionelvillard: I moved the logic into `autoscale`. PTAL

pkg/common/scheduler/statefulset/autoscaler.go
```diff
@@ -0,0 +1,119 @@
...
+				return
+			}
+
+			s.logger.Infow("checking adapter capacity",
+				zap.Int32("free", s.freeCapacity(snapshot)),
+				zap.Int32("used", s.usedCapacity(snapshot)),
+				zap.Int32("pending", s.pendingReplicas()),
+				zap.Int32("replicas", s.replicas))
+
+			ratio := s.avgUsedCapacityPerPod(snapshot) / float64(s.capacity)
```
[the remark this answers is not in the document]
lionelvillard: yes good idea. For now it's not clear how/where the capacity is set. Something to keep in mind.

pkg/common/scheduler/statefulset/autoscaler.go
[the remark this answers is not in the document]
steven0711dong: Got it. Yeah it looks ok.

pkg/common/consumer/consumer_handler.go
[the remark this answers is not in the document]
steven0711dong: Of course.

cmd/source/mt_receive_adapter/main.go
```diff
@@ -0,0 +1,31 @@
+/*
+Copyright 2019 The Knative Authors
```
matzew, line 2: 2020 ?

config/source/multi/500-adapter.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: kafkasource-mt-adapter
+  namespace: knative-eventing
+  labels:
+    contrib.eventing.knative.dev/release: devel
+spec:
+  replicas: 1 # TODO: set to 0
```
matzew, line 23: :grin:

config/source/multi/500-adapter.yaml
```diff
@@ -0,0 +1,68 @@
...
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
+        - name: POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
+
+        # DO NOT MODIFY: The values below are being filled by the kafka source controller
+        # See 500-controller.yaml
```
matzew, line 47: likely not seen... ?  any chance to "prevent" edits?

config/source/multi/500-controller.yaml
```diff
@@ -0,0 +1,85 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: kafka-controller-manager
```
matzew, line 18: I do think the name is not that great

config/source/single/100-namespace.yaml
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: knative-eventing
+  labels:
+    contrib.eventing.knative.dev/release: devel
```
matzew, line 20: could we do `ln -s` here?

matzew: for all the duplocated ?

cmd/source/mt_receive_adapter/main.go
lionelvillard: fixed.

config/source/multi/500-adapter.yaml
lionelvillard: it's mostly here for documentation purpose to say that these environment variables are modified by the controller. I don't think we can prevent edits.

config/source/multi/500-adapter.yaml
lionelvillard: yup, WIP.

config/source/multi/500-controller.yaml
lionelvillard: I know: https://github.com/knative-sandbox/eventing-kafka/issues/261

config/source/single/100-namespace.yaml
lionelvillard: good idea!

config/source/multi/500-adapter.yaml
```diff
@@ -0,0 +1,68 @@
...
+        - name: K_METRICS_CONFIG
+          value: '{"profiling.enable":"true"}'
+        - name: K_LOGGING_CONFIG
+          value: '{}'
+        - name: K_LEADER_ELECTION_CONFIG
+          value: ''
```
matzew, line 53: /hold

This seems to cause problems when deployed w/ an operator. There are issue w/ that on the eventing-core w/ the `mt pingsource`

Let's figure out a proper fix to get this out in mature way, and let's hold until

matzew, line 53: should the `K_LEADER_ELECTION_CONFIG` be also something else that empty string `' '` ?

matzew:
If possible we should rethink if we really wanna use ENV_VARs

config/source/multi/500-adapter.yaml
matzew: move those to a proper config map?

config/source/multi/500-adapter.yaml
matzew: This cause problems on the "pingsource" in combination with the operator,

see: 
* https://github.com/knative/operator/pull/381
* https://github.com/knative/eventing/issues/4627

let's have that in mind

config/source/multi/500-adapter.yaml
matzew: This is related to issues we had before (e.g. see https://github.com/knative/operator/pull/381 ) 

We must ensure this is not a problem, when being installed by any operator ...

matzew:
@maschmid see my comments on the "config" section - given our recent history w/ the pingsource "fiddling", I think this PR asks for trouble in that area ... :disappointed:

lionelvillard:
We need a more comprehensive solution. Meanwhile this PR does no harm as this is an optional component. We'll fix the config in the follow up PR.

lionelvillard:
Now that 0.20 is cut, can I get stamp on this one? @matzew @devguyio 

/unhold

lionelvillard:
> Hey @lionelvillard I've joined a bit late to this party and I probably lost some answers to stupid questions, sorry for asking these 😆

Thanks for looking at it!

> 
> * Do you plan to move Placeable at some point in eventing, or is it strongly related to mt kafka source only?

The goal is to make it generic enough so that it can be applied to other contexts.   

> * Did I understood correctly that I don't need to implement placeable to implement kafka source, right?

That's right.

> * The controller image remains the same, but depending on the env it acts as mt source or st source controller, correct?

Yup. 

> * My understanding that all the stuff in common is about scheduling the job across different pods... Is this something that can be abstracted and end in eventing? It might be interesting for eventing-kafka-broker too I guess

Yup that's the general idea. A pod is too big of an unit and cannot be dynamically resized (eg. requesting more memory).    

> * Autoscaling and data plane deployment: why the data plane is "owned" by the controller? Why don't we let users manage the data plane deployment, managing the eventual replicas, so we just handle the job scheduling and we let external tools (e.g. keda) to autoscale?

I'm not sure I fully understand what you are proposing here. The "job" scheduling is done by the controller and we are planning to use KEDA for autoscaling. What do you mean by users?

slinkydeveloper:
> The "job" scheduling is done by the controller and we are planning to use KEDA for autoscaling. What do you mean by users?

Ok so, maybe it's my misunderstanding here, but the controller only handles the scheduling of the jobs, not of the pods. If that's the case, i can literally scale up the deployment of the data plane and it should just work, right?

lionelvillard:
/hold

Depends on https://github.com/knative-sandbox/eventing-kafka/pull/321

steven0711dong (approved):
LGTM

steven0711dong (approved):
/lgtm

steven0711dong (approved):
/lgtm

steven0711dong (approved):
/lgtm cancel

PR #187, opened 2020-11-12
Add KafkaChannel WebHook to distributed channel

travis-minke-sap:
It would be ideal for both KafkaChannel implementations (consolidated / distributed) to be able to use the same WebHook for versioning/admission.  Unfortunately the WebHook "config" is all buried in the consolidated channel.  This PR refactors the location of the WebHook config, and some top level source, so that it can be included in both channel configurations.  The intent is to preserve the content and order of the consolidated channel's WebHook configuration, while adding similar config to the distributed channels' configuration.

**Release Note**

```release-note
The "distributed" KafkaChannel configuration YAML now includes the KafkaChannel WebHook which provides conversion.
```


matzew:
>  Unfortunately the WebHook "config" is all buried in the consolidated channel. This PR refactors the location of the WebHook config, and some top level source, so that it can be included in both channel configurations. 

agreed that is good to start

> The intent is to preserve the content and order of the consolidated channel's WebHook configuration, while adding similar config to the distributed channels' configuration.

+1

matzew (approved):
/lgtm
/approve

PR #189, opened 2020-11-12
Allow changing resync period via environment variable

eric-sap:
## Proposed Changes

- 🎁 Added a new optional RESYNC_PERIOD_MINUTES environment variable to the distributed controller (which is propagated to the receiver and dispatcher deployments) which allows convenient setting of a different resync period than the default of 10 hours.


cmd/channel/distributed/controller/main.go
```diff
@@ -34,5 +37,7 @@ func main() {
 	defer kafkasecret.Shutdown()
 
 	// Create The SharedMain Instance With The Various Controllers
-	sharedmain.Main(constants.ControllerComponentName, kafkachannel.NewController, kafkasecret.NewController)
+	ctx := signals.NewContext()
+	sharedmain.MainWithContext(controller.WithResyncPeriod(ctx, env.GetEnvironmentOrDie(ctx).ResyncPeriod),
+		constants.ControllerComponentName, kafkachannel.NewController, kafkasecret.NewController)
```
travis-minke-sap, line 42: Can you pull the controller.WithResyncPeriod() out onto it's own line or combine with ctx line above rather than nest inside MainWithContext() call (readability)?

pkg/channel/distributed/controller/env/environment_test.go
```diff
@@ -94,16 +105,16 @@ func TestGetEnvironment(t *testing.T) {
 	testCase.expectedError = getMissingRequiredEnvironmentVariableError(ReceiverImageEnvVarKey)
 	testCases = append(testCases, testCase)
 
+	testCase = getValidTestCase("Invalid Config - ResyncPeriodMinutes")
+	testCase.resyncPeriodMinutes = "NAN"
+	testCase.expectedError = getInvalidIntEnvironmentVariableError(testCase.resyncPeriodMinutes, env.ResyncPeriodMinutesEnvVarKey)
+	testCases = append(testCases, testCase)
+
```
travis-minke-sap, line 112: Missing an Invalid Int test case ?  (I think there are others to that effect in that test already?)

cmd/channel/distributed/controller/main.go
eric-sap: Sure, no problem

pkg/channel/distributed/controller/env/environment_test.go
eric-sap: I added some more test cases for checking the default value for the optional variable; NAN is already an invalid int test, (and valid int is already checked in every other test case)

cmd/channel/distributed/controller/main.go
eric-sap: Sure, no problem

PR #201, opened 2020-11-18
SASL: Using scram-sha-512 and scram-sha-256 support

matzew:
## Proposed Changes

- adding support for SASL (`scram-sha-512` and `scram-sha-256`)
- _moving_ `client` back to source; out of `common`
- added e2e test for SASL/TLS

## NEXT

- adding documentation

### DIFF
--- a/pkg/source/kafka_scram_client.go
+++ b/pkg/source/kafka_scram_client.go
@@ -1,0 +1,58 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package source
+
+import (
+	"crypto/sha256"
+	"crypto/sha512"
+	"hash"
+
+	"github.com/xdg/scram"
+)
+
+// SHA256 hash generator function for SCRAM conversation
+var SHA256 scram.HashGeneratorFcn = func() hash.Hash { return sha256.New() }
+
+// SHA512 hash generator function for SCRAM conversation
+var SHA512 scram.HashGeneratorFcn = func() hash.Hash { return sha512.New() }
+
+// XDGSCRAMClient struct to perform SCRAM conversation
+type XDGSCRAMClient struct {
+	*scram.Client
+	*scram.ClientConversation
+	scram.HashGeneratorFcn
+}
+
+// Begin starts SCRAM conversation
+func (x *XDGSCRAMClient) Begin(userName, password, authzID string) (err error) {
+	x.Client, err = x.HashGeneratorFcn.NewClient(userName, password, authzID)
+	if err != nil {
+		return err
+	}
+	x.ClientConversation = x.Client.NewConversation()
+	return nil
+}
+
+// Step performs step in SCRAM conversation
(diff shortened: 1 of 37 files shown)


aliok:
Looks good overall. Gonna have a look again once the build is green.

pkg/source/kafka_scram_client.go
```diff
@@ -0,0 +1,58 @@
...
+// Step performs step in SCRAM conversation
+func (x *XDGSCRAMClient) Step(challenge string) (response string, err error) {
+	response, err = x.ClientConversation.Step(challenge)
+	return
+}
+
+// Done completes SCRAM conversation
+func (x *XDGSCRAMClient) Done() bool {
+	return x.ClientConversation.Done()
+}
```
slinkydeveloper, line 58: What is this and why we need it? (sorry, i know the question might be stupid)

test/e2e-tests.sh
```diff
@@ -385,6 +395,20 @@ function test_consolidated_channel_tls() {
...
+  cp ${KAFKA_SASL_CONFIG} "${CONSOLIDATED_TEMPLATE_DIR}/configmaps/kafka-config.yaml"
+  KAFKA_CLUSTER_URL=${KAFKA_SASL_CLUSTER_URL}
+
+  install_consolidated_channel_crds || return 1
+
+  go_test_e2e -tags=e2e -timeout=40m -test.parallel=${TEST_PARALLEL} ./test/e2e -channels=messaging.knative.dev/v1beta1:KafkaChannel  || fail_test
+
+  uninstall_channel_crds || return 1
+}
+
```
slinkydeveloper, line 411: Do we really need to run the whole suite for testing tls and sasl? Could we just run a single test case?

pkg/source/kafka_scram_client.go
matzew: That's what you must implement w/ doing the "none plain" SASL, w/ Sarama. They do here: https://github.com/Shopify/sarama/blob/c53ebd09c72909bb3b7b0bc8d9bfe8fa6b7bde39/examples/sasl_scram_client/scram_client.go

I guess... we could factor that into theirs? If they want, at a later point? I see others do _same_ ... as well ... :shrug:

test/e2e-tests.sh
matzew: I'd really like to make sure the entire tests run, to have some load, and robustness 

Instead of "chaining" them, we can move them to three different jobs, with a different PR? (requires work on test-infra)

slinkydeveloper: yes this sounds like a good idea! This way, we also parallelize and make the build faster overrall

pkg/source/kafka_scram_client.go
slinkydeveloper: Ah, then i guess it's fine

matzew: FWIW: https://github.com/Shopify/sarama/issues/1837

test/e2e-tests.sh
matzew: > yes this sounds like a good idea! This way, we also parallelize and make the build faster overrall

yes! Just wanted it to work first, before getting fancy

aliok (approved):
/lgtm
/approve
/unhold

liangxj8:
Hi all, I'm wondering why we don't make the client code used as the default `SCRAMClientGeneratorFunc` automatically when the user set the `SASL.Mechanism` as `sarama.SASLTypeSCRAMSHA256` or `sarama.SASLTypeSCRAMSHA512` in `config.go`. When I am using the config, I think the client should be automatically generated according to my configuration but not telling me `kafka: invalid configuration (A SCRAMClientGeneratorFunc function must be provided to Net.SASL.SCRAMClientGeneratorFunc)`. As beginner, I spent a lot of time looking for how to generate this client.

Neustradamus:
@matzew: Thanks a lot for all about SCRAM.

Linked to:
- https://github.com/scram-xmpp/info/issues/1

PR #202, opened 2020-11-18
Upconvert kafka errors to sarama TopicError

travis-minke-sap:
It seems that Strimzi can return normal Golang errors instead of Sarama TopicErrors (which are errors).  The distributed channel contains logic for "promoting" these to TopicErrors for consistent handling upstream.  This PR enhances that implementation to String match the normal Golang errors against the Sarama/Kafka error strings.  If a match is found then the promotion is accurately performed to a specific KError instead of just to ErrUnknown.

The specific use case which highlighted this was an edge case and was ultimately handled correctly by the reconciler, but the logging was somewhat confusing.  To reproduce manually do the following...
1. Create a KafkaChannel
2. Manually add some bogus finalizer to prevent deletion.
3. Delete the KafkaChannel
4. Review logs and see an error with the following...
```
"TopicError":"kafka server: Unexpected (unknown?) server error. - kafka server: Request was for a topic or partition that does not exist on this broker.",
```
...which has nested the `does not exist` error string inside an `unknown`.   This will no longer happen and the logging / handling will be more precise.


eric-sap (approved):
Looks like a good change (not sure about what ReviewDog is saying though)

eric-sap (approved):
ReviewDog appears to be appeased

PR #205, opened 2020-11-19
Move AuthConfig code to one place and add tests

matzew:

## Proposed Changes

- If no `saslType` is provided, we should default to "PLAIN" (the source only supports plain)
- moving code to one central place
- adding tests

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/channel/consolidated/dispatcher/dispatcher.go
+++ b/pkg/channel/consolidated/dispatcher/dispatcher.go
@@ -95,35 +95,10 @@ func NewDispatcher(ctx context.Context, args *KafkaDispatcherArgs) (*KafkaDispat
 	conf.Consumer.Return.Errors = true    // Returns the errors in ConsumerGroup#Errors() https://godoc.org/github.com/Shopify/sarama#ConsumerGroup
 	conf.Producer.Return.Successes = true // Must be enabled for sync producer
 
-	// Get the auth info
-	if args.KafkaAuthConfig != nil {
-		// tls
-		if args.KafkaAuthConfig.TLS != nil {
-			conf.Net.TLS.Enable = true
-			tlsConfig, err := source.NewTLSConfig(args.KafkaAuthConfig.TLS.Usercert, args.KafkaAuthConfig.TLS.Userkey, args.KafkaAuthConfig.TLS.Cacert)
-			if err != nil {
-				return nil, err
-			}
-			conf.Net.TLS.Config = tlsConfig
-		}
-		// SASL
-		if args.KafkaAuthConfig.SASL != nil {
-			conf.Net.SASL.Enable = true
-			conf.Net.SASL.Handshake = true
-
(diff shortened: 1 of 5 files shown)


pkg/source/client.go
```diff
@@ -132,13 +132,19 @@ func MakeAdminClient(clientID string, kafkaAuthCfg *utils.KafkaAuthConfig, boots
...
+
+func UpdateSaramaConfigWithKafkaAuthConfig(saramaConf *sarama.Config, kafkaAuthCfg *utils.KafkaAuthConfig) {
 	if kafkaAuthCfg != nil {
 		// tls
 		if kafkaAuthCfg.TLS != nil {
 			saramaConf.Net.TLS.Enable = true
 			tlsConfig, err := NewTLSConfig(kafkaAuthCfg.TLS.Usercert, kafkaAuthCfg.TLS.Userkey, kafkaAuthCfg.TLS.Cacert)
 			if err != nil {
-				return nil, err
+				return
```
lberk, line 147: any particular reason we're no longer passing the error back up?  It would probably be good to at least log it as an error for the user/admin

matzew: it was before call insde a returning func. 

+1 on log - will update

lberk: what if we changed the return type to an error, and returned nil in other cases, but then were able to `return fmt.Errorf("Error creating TLS config: %w, err")` and then the calling function can act accordingly?

matzew: `fmt.Errorf` is not allowed - see the linter

lberk: `fmt.Errorf` isn't allowed when the result isn't used, but if its used as the return value it's proper, no?

pkg/channel/consolidated/dispatcher/dispatcher.go
```diff
@@ -95,36 +95,8 @@ func NewDispatcher(ctx context.Context, args *KafkaDispatcherArgs) (*KafkaDispat
...
-				conf.Net.SASL.SCRAMClientGeneratorFunc = func() sarama.SCRAMClient { return &source.XDGSCRAMClient{HashGeneratorFcn: source.SHA512} }
-				conf.Net.SASL.Mechanism = sarama.SASLTypeSCRAMSHA512
-			}
-
-			conf.Net.SASL.User = args.KafkaAuthConfig.SASL.User
-			conf.Net.SASL.Password = args.KafkaAuthConfig.SASL.Password
-		}
-	}
+	// append the auth info:
+	source.UpdateSaramaConfigWithKafkaAuthConfig(conf, args.KafkaAuthConfig)
```
lberk, line 99: if we return an error from  `UpdateSaramaConfigWithKafkaAuthConfig` below -- then we can catch it here and still `return nil, err` from `NewDispatcher` as well

pkg/source/client.go
matzew: updated - thanks!

lberk:
/lgtm
thanks!

PR #214, opened 2020-11-20
Shared KafkaChannel CRD Config YAML

travis-minke-sap:
Similar to the recent refactor of the KafkaChannel WebHook config YAML (#187) , the actual KafkaChannel CRD YAML should be shared between the two KafkaChannel implementations.  The two CRDs were already 99% identical, and this PR is simply moving the KafkaChannel CRD from `config/channel/consolidated` up a level and adjusting/providing new simlinks.


matzew:
very cool!

/lgtm
/approve

PR #220, opened 2020-11-23
Use SYSTEM_NAMESPACE variable instead of hardcoded value

eric-sap:
## Proposed Changes

- 🧽  Remove references to hard-coded "knative-eventing" namespace, preferring system.Namespace() call
- 🧽  Store the environment, loaded in the controller main(), in the context instead of re-reading it whenever NewController is called


pkg/channel/distributed/common/testing/util.go
```diff
@@ -84,3 +89,13 @@ func RetryGet(url string, pause time.Duration, retryCount int, retryAgain int) (
 	// Request failed too many times; return the error for caller to process
 	return resp, err
 }
+
+// Sets the environment variables that are necessary for common components
+func SetTestEnvironment(t *testing.T) {
+	// The system.Namespace() call panics if the SYSTEM_NAMESPACE variable isn't set, so
+	// this sets "knative-eventing" explicitly for testing purposes
+	assert.Nil(t, os.Setenv(system.NamespaceEnvKey, "knative-eventing"))
+	// The logging.ConfigMapName() has a default if it isn't present, so this just uses that
```
travis-minke-sap, line 98: I'm assuming there's nothing special about the test value of `knative-eventing` in which case I generally prefer to use something non-standard and obviously a "test" value just to be sure?   Say... `test-system-namespace` or something?

travis-minke-sap:
/approve
/hold

* holding for some final manual testing...

pkg/channel/distributed/controller/kafkachannel/controller.go
```diff
@@ -55,7 +55,7 @@ func NewController(ctx context.Context, _ configmap.Watcher) *controller.Impl {
 	serviceInformer := service.Get(ctx)
 
 	// Load The Environment Variables
-	environment, err := env.GetEnvironment(logger)
+	environment, err := env.FromContext(ctx)
 	if err != nil {
 		logger.Fatal("Failed To Load Environment Variables - Terminating!", zap.Error(err))
 	}
```
travis-minke-sap, line 61: Can you change the Fatal message to say `Failed To Get Environment From Context - Terminating!` to match the new usage.

pkg/channel/distributed/controller/kafkasecret/controller.go
```diff
@@ -121,7 +122,7 @@ func enqueueSecretOfKafkaChannel(controller *controller.Impl) func(obj interface
 				secretName := labels[constants.KafkaSecretLabel]
 				if len(secretName) > 0 {
 					controller.EnqueueKey(types.NamespacedName{
-						Namespace: commonconstants.KnativeEventingNamespace,
+						Namespace: system.Namespace(),
 						Name:      secretName,
```
travis-minke-sap, line 126: Since we have the environment.SystemNamespace in the `NewController()` function above which calls this, can we pass it in as a parameter instead of re-loading from the GO environment again?

pkg/channel/distributed/common/testing/util.go
eric-sap: Yep, although it apparently can't be "dummy" because of ReviewDog

pkg/channel/distributed/controller/kafkachannel/controller.go
eric-sap: of course

pkg/channel/distributed/controller/kafkasecret/controller.go
eric-sap: Excellent idea

travis-minke-sap:
@matzew - Can you take a look at the test/e2e files to make sure you're ok with that - I've looked at the rest.

test/config/config-tracing.yaml
```diff
@@ -16,9 +16,9 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: config-tracing
-  namespace: knative-eventing
+  namespace: ${SYSTEM_NAMESPACE}
```
matzew, line 19: with the `sed` that's a good one :smile: 

I will do the same for the consolidated channel too

matzew (approved):
/lgtm
/approve

test/config/config-tracing.yaml
eric-sap: I originally used "envsubst" which worked perfectly locally but is apparently not available on the build image.

PR #224, opened 2020-11-24
Move the Kafka Source to the knative-eventing namespace

matzew:

Fixes #

## Proposed Changes

- move source to knative-eventing
- add old/deprecated controller replica
-remove config-xyz duplicates
- change the leader election cfg name for the source

**Release Note**

```release-note
The KafkaSource will be installed in the knative-eventing namespace, and the old controller in knative-sources is scalled to 0
```

**Docs**


pierDipi (commented):
I like it, we need to make sure that the Kafka source is installable without installing eventing though.

config/source/configmaps/leader-election.yaml
```diff
@@ -15,8 +15,8 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: config-leader-election-kafka
-  namespace: knative-sources
+  name: config-leader-election-kafka-source
```
antoineco, line 18: This will leave duplicates upon the next deployment, do we want to introduce that here?

matzew: Good question, @antoineco

matzew:
@antoineco @pierDipi @lionelvillard I updated this PR


For now we will have `config-leader-election-kafka` CMs, however, I feel this is wrong, as expressed here:
https://github.com/knative-sandbox/eventing-kafka/pull/231

Also related:
https://github.com/knative/eventing/issues/4593

antoineco:
So what's the plan, merge this, hold #231, then un-hold #231 after this is merged?

config/source/configmaps/leader-election.yaml
antoineco: I think we should avoid renaming things in this PR if they are unrelated to the original intention which is "Move the Kafka Source to the knative-eventing namespace".

matzew:
@antoineco 

> I  think we should avoid renaming things in this PR if they are unrelated to the original intention which is "Move the Kafka Source to the knative-eventing namespace"

Sure, I updated earlier today, see: https://github.com/knative-sandbox/eventing-kafka/pull/224/files 

> So what's the plan, merge this, hold #231, then un-hold #231 after this is merged?

I'd think so - the #231 is already on hold ;-)

antoineco (approved):
/lgtm

PR #196, opened 2020-11-16
Common Sarama Clients (admin, producer, consumer)

travis-minke-sap:
Related to #192 and consolidating Sarama client creation (admin, producer, consumer).  

**This is a POC PR at this point!!!**
**Do not merge this yet - still very preliminary (and broken ; )**
/hold

Focus on the API for initial pass...
```
# Moved FROM pkg/channel/distributed/common/kakfa TO pkg/common/kafka
func CreateAdminClient(ctx context.Context, brokers []string, saramaConfig *sarama.Config, clientId string, adminClientType AdminClientType) (AdminClientInterface, error)
func CreateConsumerGroup(brokers []string, config *sarama.Config, groupId string) (sarama.ConsumerGroup, metrics.Registry, error)
func CreateSyncProducer(brokers []string, config *sarama.Config) (sarama.SyncProducer, metrics.Registry, error)
```

## Proposed Changes
- Move brokers string from Secret to ConfigMap in "distributed" KafkaChannel
- Move `pkg/channel/distributed/common/kafka` logic to `pkg/common/kafka` for reuse by both channel implementations.
- Remove all Secret loading/handling from new common API
  - Pass in brokers []string instead of loading from secret
  - For EventHubs pass in connectionStrings
  - Convert EventHubs implementation to use connectionstrings instead of secrets.
  - Keep EventHub namespace pooling ; )
- Change new common API to expect complete/valid `sarama.Config` instead of updating with SASL auth
- Create complete `sarama.Config` higher up main/controller/etc instead.
- Pass Secret name/namespace in ConfigMap & Load in Controller

> TODO - add release notes below once complete.

**Release Note**

```release-note

```

**Docs**

Issue #188, opened 2020-11-12
Kafka client refactoring


matzew:
Refactor to have some more commonly shared "kafka client" code

## Admin client for the channel controller

* https://github.com/knative-sandbox/eventing-kafka/blob/1a46a74b179207ff6217484089cf5970e0716066/pkg/channel/consolidated/reconciler/controller/resources/client.go#L23-L28
* https://github.com/knative-sandbox/eventing-kafka/tree/c3e6f18ffe5dd67b36dcb42b7b3e95838e4ae4f8/pkg/channel/distributed/common/kafka/admin

## Normal dispatcher client:

* https://github.com/knative-sandbox/eventing-kafka/blob/1a46a74b179207ff6217484089cf5970e0716066/pkg/channel/consolidated/dispatcher/dispatcher.go#L85-L90
* https://github.com/knative-sandbox/eventing-kafka/blob/c3e6f18ffe5dd67b36dcb42b7b3e95838e4ae4f8/pkg/channel/distributed/common/kafka/sarama/sarama.go

aliok:
/close

I say this is done via https://github.com/knative-sandbox/eventing-kafka/issues/268 and https://github.com/knative-sandbox/eventing-kafka/issues/298

Reopen if you had something else in your mind

PR #226, opened 2020-11-24
[distributed] Use same configmap name, for the channels

matzew:

Fixes #

## Proposed Changes

- align name of the configmap with the consolidated
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/config/channel/distributed/300-eventing-kafka-configmap.yaml
+++ b/config/channel/distributed/300-eventing-kafka-configmap.yaml
@@ -59,5 +59,5 @@ data:
       adminType: kafka # One of "kafka", "azure", "custom"
 kind: ConfigMap
 metadata:
-  name: config-eventing-kafka
+  name: config-kafka
   namespace: knative-eventing
--- a/pkg/channel/distributed/common/config/configwatcher.go
+++ b/pkg/channel/distributed/common/config/configwatcher.go
@@ -30,7 +30,7 @@ import (
 )
 
 // The EventingKafkaConfig and these EK sub-structs contain our custom configuration settings,
-// stored in the config-eventing-kafka configmap.  The sub-structs are explicitly declared so that they
+// stored in the config-kafka configmap.  The sub-structs are explicitly declared so that they
 // can have their own JSON tags in the overall EventingKafkaConfig
 type EKKubernetesConfig struct {
 	CpuLimit      resource.Quantity `json:"cpuLimit,omitempty"`
--- a/pkg/channel/distributed/common/config/constants.go
+++ b/pkg/channel/distributed/common/config/constants.go
@@ -19,7 +19,7 @@ package config
 // Package Constants
 const (
 	// The name of the configmap used to hold eventing-kafka settings
-	SettingsConfigMapName = "config-eventing-kafka"
+	SettingsConfigMapName = "config-kafka"
 	// The name of the keys in the Data section of the eventing-kafka configmap that holds Sarama and Eventing-Kafka configuration YAML
 	SaramaSettingsConfigKey        = "sarama"
 	EventingKafkaSettingsConfigKey = "eventing-kafka"
--- a/pkg/channel/distributed/common/kafka/sarama/sarama.go
+++ b/pkg/channel/distributed/common/kafka/sarama/sarama.go
@@ -131,7 +131,7 @@ The following shows an example of the expected usage...
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: config-eventing-kafka
+  name: config-kafka
   namespace: knative-eventing
 data:
   sarama: |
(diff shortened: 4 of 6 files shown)


matzew:
/hold
/assign @travis-minke-sap 
/cc @eric-sap 

I was wondering how you'd think about this.

I was also thinking about doing a follow up PR, where I'd like to also change the CM and use the `authSecretName` and `authSecretNamespace` keys, instead of the current label you are using.

Let me know what you think!

travis-minke-sap:
> I was wondering how you'd think about this.

This PR looks fine - changing the name of the ConfigMap for consistency is a good thing ; )
/lgtm
 
> I was also thinking about doing a follow up PR, where I'd like to also change the CM and use the `authSecretName` and `authSecretNamespace` keys, instead of the current label you are using.

This is a bit more problematic and I don't think can be handled that simply.  I worked up a POC PR a week or so ago (didn't see any response so I dropped it for now) to look into converting our implementation to the new pattern.  The new pattern (`authSecretXXX` keys in ConfigMap) is good/fine and we can/should switch to it.  I think it will simplify things for us quite a bit, and my hope was that both Consolidated & Distributed KafkaChannels could use it.  This would add support for EventHubs and the "Custom" sidecar adminclient to Consolidated essentially for free.

The difficulty is in our current implementation's AdminClient's support for EventHubs.  It dynamically looks up secrets based on the label and creates a Pool of connections to EventHub Namespaces.  My POC PR was focused on reworking the creation of the AdminClients so that they didn't have to deal with Auth information at all, but were instead presented with the brokers information and a complete sarama.Config.  This worked fine for the normal 'kafka' use case as well as for our 'custom' use case.  The EventHubs will require a bit more work to get right and we might consider dropping the multi-EventHub-Namespace pooling feature which would simplify things.  Also, we (SAP) are no longer using EventHubs and are only carrying that feature along as a nice-to-have, so maybe we consider dropping it - or rewriting it as a 'custom' sidecar similar.

So, I would hold off for now and we can discuss when we meet with Davy to try to nail down a consolidation roadmap?

matzew:
> This is a bit more problematic and I don't think can be handled that simply. I worked up a POC PR a week or so ago (didn't see any response so I dropped it for now) to look into converting our implementation to the new pattern. The new pattern (authSecretXXX keys in ConfigMap) is good/fine and we can/should switch to it. I think it will simplify things for us quite a bit. My hope was that both Consolidated & Distributed KafkaChannels could use the same adminclient (and produce/consumer creation) from the POC PR. This would add support for EventHubs and the "Custom" sidecar adminclient to Consolidated essentially for free.

Sorry for missing the POC PR - wanna reopen it? But glad the "new pattern" works out !

> both Consolidated & Distributed KafkaChannels could use the same adminclient (and produce/consumer creation) from the POC PR.

+1000 I created this https://github.com/knative-sandbox/eventing-kafka/issues/188  as a "marker" ticket, to track some of that "refactorings".  we are more than happy to help with this: implementation, testing etc


> The difficulty is in our current implementation's AdminClient's support for EventHubs. It dynamically looks up secrets based on the label and creates a Pool of connections to EventHub Namespaces. My POC PR was focused on reworking the creation of the AdminClients so that they didn't have to deal with Auth information at all, but were instead presented with the brokers information and a complete sarama.Config.  This worked fine for the normal 'kafka' use case as well as for our 'custom' use case. The EventHubs will require a bit more work to get right and we might consider dropping the multi-EventHub-Namespace pooling feature which would simplify things. Also, we (SAP) are no longer using EventHubs and are only carrying that feature along as a nice-to-have, so maybe we consider dropping it - or rewriting it as a 'custom' sidecar similar.

I think throwing away all of this is hard :-) Or unfortune - would be good if we could - all together - figure out something

> So, I would hold off for now and we can discuss when we meet with Davy to try to nail down a consolidation roadmap?

Yeah, sounds good!

travis-minke-sap:
> Sorry for missing the POC PR - wanna reopen it? But glad the "new pattern" works out !

The PR is stale and wasn't ever "finished" anyway so I closed it but you can still look at it (#196).  The important question was whether the Consolidated KafkaChannel could use the the same logic for creating the admin / producer / consumer that looked something like this...

```
# Moved FROM pkg/channel/distributed/common/kakfa TO pkg/common/kafka
func CreateAdminClient(ctx context.Context, brokers []string, saramaConfig *sarama.Config, clientId string, adminClientType AdminClientType) (AdminClientInterface, error)
func CreateConsumerGroup(brokers []string, config *sarama.Config, groupId string) (sarama.ConsumerGroup, metrics.Registry, error)
func CreateSyncProducer(brokers []string, config *sarama.Config) (sarama.SyncPro
```

The important change from what we have, and I think what was in consolidated channel has, is the fact that we're separating the logic for creating the `sarama.Config` object.  This allows the two channels to differ until we consolidate the logic around the ConfigMap / Secret.

Once we settle on a plan for consolidating I can resurrect the implementation / ideas and look further into how to address the EventHub Pooling, etc...

PR #231, opened 2020-11-25
Go with the overall default of the leader election

matzew:

## Proposed Changes

- with the global default of the `config-leader-election`, it is still possible to do some fine-tuning, via the `CONFIG_LEADERELECTION_NAME` env-var on the deployments (e.g. controllers/dispatchers)

**Release Note**

```release-note
KafkaSource and KafkaChannel will be default use the `config-leader-election` CM for configs
```

**Docs**

### DIFF
--- a/config/channel/consolidated/configmaps/leader-election.yaml
+++ b/config/channel/consolidated/configmaps/leader-election.yaml
@@ -15,7 +15,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: config-leader-election-kafka
+  name: config-leader-election
   namespace: knative-eventing
 data:
   # An inactive but valid configuration follows; see example.
--- a/config/channel/consolidated/deployments/dispatcher.yaml
+++ b/config/channel/consolidated/deployments/dispatcher.yaml
@@ -49,11 +49,11 @@ spec:
             fieldRef:
               fieldPath: metadata.name
         - name: METRICS_DOMAIN
-          value: "knative.dev/eventing"
+          value: knative.dev/eventing
         - name: CONFIG_LOGGING_NAME
-          value: "config-logging"
+          value: config-logging
         - name: CONFIG_LEADERELECTION_NAME
-          value: "config-leader-election"
+          value: config-leader-election
         - name: CONTAINER_NAME
           value: dispatcher
         ports:
--- a/config/channel/consolidated/deployments/dispatcher.yaml
+++ b/config/channel/consolidated/deployments/dispatcher.yaml
@@ -53,7 +53,7 @@ spec:
         - name: CONFIG_LOGGING_NAME
           value: "config-logging"
         - name: CONFIG_LEADERELECTION_NAME
-          value: "config-leader-election-kafka"
+          value: "config-leader-election"
         - name: CONTAINER_NAME
           value: dispatcher
         ports:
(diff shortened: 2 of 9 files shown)


matzew:
@travis-minke-sap @antoineco how do you feel about this ? 

That way, we can still leverage a global `config-leader-election`, while via the ENV_VAR of the deployment we have still the chance to set it to a different ConfigMap

config/channel/consolidated/deployments/dispatcher.yaml
```diff
@@ -53,7 +53,7 @@ spec:
         - name: CONFIG_LOGGING_NAME
           value: "config-logging"
         - name: CONFIG_LEADERELECTION_NAME
-          value: "config-leader-election-kafka"
+          value: "config-leader-election"
```
antoineco, line 56: nit: No quotes required.

pkg/channel/consolidated/reconciler/controller/resources/dispatcher.go
```diff
@@ -112,7 +112,7 @@ func makeEnv(args DispatcherArgs) []corev1.EnvVar {
 		Value: "config-logging",
 	}, {
 		Name:  "CONFIG_LEADERELECTION_NAME",
-		Value: "config-leader-election-kafka",
+		Value: "config-leader-election",
```
antoineco, line 115: Don't you want to read that from the env?

matzew: yeah, this here "makes" the dispatcher ...

matzew:
@antoineco I think we even should _REMOVE_ the duplicated config for "default" values, all togehter :-) 

https://github.com/knative/eventing/issues/4593

lionelvillard:
note that leader election has been disabled for the dispatcher

pkg/channel/consolidated/reconciler/controller/resources/dispatcher.go
antoineco: But do we want to force it to use the 'global' config? Wouldn't it be wiser to pass whatever the controller is using? (which I guess is a less surprising behaviour)

antoineco:
> 
> 
> @antoineco I think we even should _REMOVE_ the duplicated config for "default" values, all togehter :-)
> 
> [knative/eventing#4593](https://github.com/knative/eventing/issues/4593)

@matzew but then it's a tad harder for users to discover what can be overridden, and expects them to know the `pkg` repo pretty well. I wouldn't set every possible env var in every controller of course, but maybe we can agree on keeping logging + obs + tracing + leader election for discoverability?

matzew:
@antoineco I updated

pkg/channel/consolidated/reconciler/controller/resources/dispatcher.go
matzew: > But do we want to force it to use the 'global' config? Wouldn't it be wiser to pass whatever the controller is configured to use? (which I guess is a less surprising behaviour)

I'd like to address this later - it's generally funky :-) 
This is only executed if the deployment of the dispatcher needs updates...  Perhaps it's cleaner to ignore them in code?
but than it would be a diff the actual deployment cfg we have on file

config/channel/consolidated/configmaps/leader-election.yaml
```diff
@@ -15,7 +15,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: config-leader-election-kafka
+  name: config-leader-election
```
antoineco, line 18: This will override core's version. Are we sure we want to include a copy in this repo?

antoineco (approved):
/lgtm

But I do hope including copies of the leader election configs won't create chaos in core.

PR #235, opened 2020-11-26
Remove configs with no concrete value/override

matzew:

Fixes #

## Proposed Changes

- remove pointless files (they had no real value - but would do some nasty overrides)

**Release Note**

```release-note

```

**Docs**


antoineco (approved):
/lgtm

PR #237, opened 2020-11-27
Add kafkaSource saslType, e2e tests for auth (tls and sasl512)

### LINKED ISSUE #200: Improve SASL for Source
**Problem**
The source only supports SASL plain

**Exit Criteria**

support of `scram-sha-512`

**Time Estimate (optional):**
1-2

lberk:
Fixes #200 

Adds SASL mechanism support for the kafkasource, adds a variety of testing, including e2e testing with plain (no auth), sasl512 and tls

### DIFF
--- a/pkg/apis/bindings/v1alpha1/kafka_types.go
+++ b/pkg/apis/bindings/v1alpha1/kafka_types.go
@@ -56,6 +56,10 @@ type KafkaSASLSpec struct {
 	// Password is the Kubernetes secret containing the SASL password.
 	// +optional
 	Password SecretValueFromSource `json:"password,omitempty"`
+
+	// Type of saslType, defaults to plain (vs SCRAM-SHA-512 or SCRAM-SHA-256)
+	// +optional
+	Type SecretValueFromSource `json:"saslType,omitempty"`
 }
 
 type KafkaTLSSpec struct {
--- a/pkg/apis/bindings/v1alpha1/kafka_types.go
+++ b/pkg/apis/bindings/v1alpha1/kafka_types.go
@@ -59,7 +59,7 @@ type KafkaSASLSpec struct {
 
 	// Type of saslType, defaults to plain (vs SCRAM-SHA-512 or SCRAM-SHA-256)
 	// +optional
-	Type SecretValueFromSource `json:"saslType,omitempty"`
+	Type SecretValueFromSource `json:"type,omitempty"`
 }
 
 type KafkaTLSSpec struct {
--- a/pkg/source/client.go
+++ b/pkg/source/client.go
@@ -33,6 +33,7 @@ type AdapterSASL struct {
 	Enable   bool   `envconfig:"KAFKA_NET_SASL_ENABLE" required:"false"`
 	User     string `envconfig:"KAFKA_NET_SASL_USER" required:"false"`
 	Password string `envconfig:"KAFKA_NET_SASL_PASSWORD" required:"false"`
+	Type     string `envconfig:"KAFKA_NET_SASL_TYPE" required:"false"`
 }
 
 type AdapterTLS struct {
@@ -65,8 +66,22 @@ func NewConfig(ctx context.Context) ([]string, *sarama.Config, error) {
 
 	if env.Net.SASL.Enable {
 		cfg.Net.SASL.Enable = true
+		cfg.Net.SASL.Handshake = true
 		cfg.Net.SASL.User = env.Net.SASL.User
 		cfg.Net.SASL.Password = env.Net.SASL.Password
+
+		// We default to plain sasl type
+		cfg.Net.SASL.Mechanism = sarama.SASLTypePlaintext
+
+		if env.Net.SASL.Type == sarama.SASLTypeSCRAMSHA256 {
+			cfg.Net.SASL.SCRAMClientGeneratorFunc = func() sarama.SCRAMClient { return &XDGSCRAMClient{HashGeneratorFcn: SHA256} }
+			cfg.Net.SASL.Mechanism = sarama.SASLTypeSCRAMSHA256
+		}
+
+		if env.Net.SASL.Type == sarama.SASLTypeSCRAMSHA512 {
+			cfg.Net.SASL.SCRAMClientGeneratorFunc = func() sarama.SCRAMClient { return &XDGSCRAMClient{HashGeneratorFcn: SHA512} }
+			cfg.Net.SASL.Mechanism = sarama.SASLTypeSCRAMSHA512
+		}
 	}
 
 	if env.Net.TLS.Enable {
--- a/pkg/source/client_test.go
+++ b/pkg/source/client_test.go
@@ -334,16 +334,160 @@ func generateCert(t *testing.T) (string, string) {
 }
 
 func TestNewConfig(t *testing.T) {
-	ctx := context.Background()
-
 	// Increasing coverage
-	_ = os.Setenv("KAFKA_BOOTSTRAP_SERVERS", "my-cluster-kafka-bootstrap.my-kafka-namespace:9092")
+	defaultBootstrapServer := "my-cluster-kafka-bootstrap.my-kafka-namespace:9092"
+	defaultSASLUser := "secret-user"
+	defaultSASLPassword := "super-seekrit-password"
+	testCases := map[string]struct {
+		env             map[string]string
+		enabledTLS      bool
+		enabledSASL     bool
+		wantErr         bool
+		saslMechanism   string
+		bootstrapServer string
+		saslUser        string
+		saslPassword    string
+	}{
+		"Just bootstrap Server": {
+			env: map[string]string{
+				"KAFKA_BOOTSTRAP_SERVERS": defaultBootstrapServer,
+			},
+			bootstrapServer: defaultBootstrapServer,
+		},
+		"Incorrect bootstrap Server": {
+			env: map[string]string{
+				"KAFKA_BOOTSTRAP_SERVERS": defaultBootstrapServer,
+			},
+			bootstrapServer: "ImADoctorNotABootstrapServerJim!",
+			wantErr:         true,
+		},
+		/*
+			TODO
+			"Multiple bootstrap servers": {
+				env: map[string]string{
 
-	servers, config, err := NewConfig(ctx)
+				},
 
-	require.NoError(t, err)
-	require.NotNil(t, config)
-	require.Equal(t, []string{"my-cluster-kafka-bootstrap.my-kafka-namespace:9092"}, servers)
+			},*/
+		"No Auth": {
+			env: map[string]string{
+				"KAFKA_BOOTSTRAP_SERVERS": defaultBootstrapServer,
+			},
+			enabledTLS:      false,
+			enabledSASL:     false,
+			bootstrapServer: defaultBootstrapServer,
+		},
+		"Defaulting to SASL-Plain Auth (none specified)": {
+			env: map[string]string{
+				"KAFKA_BOOTSTRAP_SERVERS": defaultBootstrapServer,
+				"KAFKA_NET_SASL_ENABLE":   "true",
+			},
+			enabledSASL:     true,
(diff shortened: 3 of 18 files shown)


matzew:
I tested the source, e2e, locally - like:




1) Setup strimzi (with only SASL/TLS), like:

```yaml
apiVersion: kafka.strimzi.io/v1beta1
kind: Kafka
metadata:
  name: my-cluster
spec:
  kafka:
    version: 2.6.0
    replicas: 3
    listeners:
      - name: sasl
        port: 9094
        type: internal
        tls: true
        authentication:
          type: scram-sha-512
    config:
      offsets.topic.replication.factor: 3
      transaction.state.log.replication.factor: 3
      transaction.state.log.min.isr: 2
      log.message.format.version: "2.6"
    storage:
      type: jbod
      volumes:
      - id: 0
        type: persistent-claim
        size: 100Gi
        deleteClaim: false
  zookeeper:
    replicas: 3
    storage:
      type: persistent-claim
      size: 100Gi
      deleteClaim: false
  entityOperator:
    topicOperator: {}
    userOperator: {}
```

2) Setup a Kafka User via Strimzi, like: 

```yaml
apiVersion: kafka.strimzi.io/v1beta1
kind: KafkaUser
metadata:
  name: my-sasl-user
  labels:
    strimzi.io/cluster: my-cluster
spec:
  authentication:
    type: scram-sha-512
  authorization:
    type: simple
    acls:
      # Example ACL rules for consuming from knative-messaging-kafka using consumer group my-group
      - resource:
          type: topic
          name: "*"
        operation: Read
        host: "*"
      - resource:
          type: topic
          name: "*"
        operation: Describe
        host: "*"
      - resource:
          type: group
          name: "*"
        operation: Read
        host: "*"
      # Example ACL rules for producing to topic knative-messaging-kafka
      - resource:
          type: topic
          name: "*"
        operation: Write
        host: "*"
      - resource:
          type: topic
          name: "*"
        operation: Create
        host: "*"
      - resource:
          type: topic
          name: "*"
        operation: Describe
        host: "*"
```


3) Create the secret with the SASL and TLS info, like (note: using `default` namespace):

```shell
STRIMZI_CRT=$(kubectl -n kafka get secret my-cluster-cluster-ca-cert --template='{{index .data "ca.crt"}}' | base64 --decode )
SASL_PASSWD=$(kubectl -n kafka get secret my-sasl-user --template='{{index .data "password"}}' | base64 --decode )

kubectl create secret --namespace default generic strimzi-sasl-secret \
    --from-literal=ca.crt="$STRIMZI_CRT" \
    --from-literal=password="$SASL_PASSWD" \
    --from-literal=saslType="SCRAM-SHA-512" \
    --from-literal=user="my-sasl-user"
```

5) Define some source:

```yaml
apiVersion: sources.knative.dev/v1beta1
kind: KafkaSource
metadata:
  name: kafka-source2
spec:
  bootstrapServers:
   - my-cluster-kafka-bootstrap.kafka.svc.cluster.local:9094
  topics:
   - my-topic
  net:
    sasl:
      enable: true
      user:
        secretKeyRef:
          name: strimzi-sasl-secret
          key: user
      password:
        secretKeyRef:
          name: strimzi-sasl-secret
          key: password
      saslType:
        secretKeyRef:
          name: strimzi-sasl-secret
          key: saslType
    tls:
      enable: true
      caCert:
        secretKeyRef:
          name: strimzi-sasl-secret
          key: ca.crt
  sink:
    ref:
      apiVersion: serving.knative.dev/v1
      kind: Service
      name: event-display
```

**SOURCE POD CAME UP**

All in `default` namespace ...

pkg/apis/bindings/v1alpha1/kafka_types.go
```diff
@@ -56,6 +56,10 @@ type KafkaSASLSpec struct {
 	// Password is the Kubernetes secret containing the SASL password.
 	// +optional
 	Password SecretValueFromSource `json:"password,omitempty"`
+
+	// Type of saslType, defaults to plain (vs SCRAM-SHA-512 or SCRAM-SHA-256)
+	// +optional
+	Type SecretValueFromSource `json:"saslType,omitempty"`
```
matzew, line 62: should we name it `type` instead of `saslType` ?  On the source it's more strongly typed / structured - since you are on the `.SASL` path here already

test/e2e/kafka_source_test.go
```diff
@@ -97,14 +214,80 @@ func testKafkaSource(t *testing.T, name string, version string, messageKey strin
...
+				bootStrapServer: kafkaBootstrapURLSASL,
+				Net: AdapterNet{
+					AdapterSASL{
+						Enable:   true,
+						User:     kafkaClusterSASLUsername,
+						Password: kafkaClusterPassword,
+						Type:     "SCRAM-SHA-512",
+					},
+					AdapterTLS{
+						Enable: false,
```
matzew, line 252: we have SASL only enabled w TLS:
https://github.com/knative-sandbox/eventing-kafka/blob/master/test/config/100-kafka-ephemeral-triple-2.6.0.yaml#L20-L25

test/e2e/kafka_source_test.go
```diff
@@ -97,14 +214,80 @@ func testKafkaSource(t *testing.T, name string, version string, messageKey strin
...
+					},
+				},
+			},
+		},
+		/*"sasl-256": {
```
matzew, line 257: we can not test this - strimzi does not support that

test/e2e/kafka_source_test.go
```diff
@@ -97,14 +214,80 @@ func testKafkaSource(t *testing.T, name string, version string, messageKey strin
...
+						Password: kafkaClusterPassword,
+						Type:     sarama.SASLTypeSCRAMSHA256,
+					},
+					AdapterTLS{
+						Enable: false,
+					},
+				},
+			},
+		},
+		"sasl-p": {
```
matzew, line 273: lets not do this :-)   it's plain text w/o TLS ...

test/e2e/kafka_source_test.go
```diff
@@ -42,26 +45,138 @@ import (
...
-	kafkaClusterName      = "my-cluster"
-	kafkaClusterNamespace = "kafka"
+	kafkaBootstrapUrlPlain   = "my-cluster-kafka-bootstrap.kafka.svc:9092"
+	kafkaBootstrapUrlTLS     = "my-cluster-kafka-bootstrap.kafka.svc:9093"
+	kafkaBootstrapURLSASL    = "my-cluster-kafka-bootstrap.kafka.svc:9094"
+	kafkaClusterName         = "my-cluster"
+	kafkaClusterNamespace    = "kafka"
+	kafkaClusterSASLUsername = "my-sasl-user"
+	kafkaClusterTLSUsername  = "my-tls-user"
+	kafkaClusterPassword     = "my-seekrit-password"
```
matzew, line 55: In the `consolidated-sasl` test run, we create a secrte before the actual run:
https://github.com/knative-sandbox/eventing-kafka/blob/master/test/e2e-tests.sh#L494

And the secret gets all the required settings for SASL/TLS, from Strimzi (CA CERT and the SASL USER):
https://github.com/knative-sandbox/eventing-kafka/blob/master/test/e2e-tests.sh#L369-L372

matzew: How about:

1) we create a new `kafka_source_sasl_test.go` file, 
https://github.com/knative-sandbox/eventing-kafka/blob/192986e75e1db91bf45fd9dd45fba9176ce9bfe4/test/e2e/kafka_source_test.go#L1-L2

2) add a `sasl` build-tag after https://github.com/knative-sandbox/eventing-kafka/blob/192986e75e1db91bf45fd9dd45fba9176ce9bfe4/test/e2e/kafka_source_test.go#L1-L2 

e.g.:
```
//+build sasl
```

3) Run that test only on the SASL test run:
https://github.com/knative-sandbox/eventing-kafka/blob/master/test/e2e-tests.sh#L412
e.g. change that line to:

```
...
  go_test_e2e -tags=e2e,sasl -timeout=40m -test.parallel=${TEST_PARALLEL} ./test/e2e -channels=messaging.knative.dev/v1beta1:KafkaChannel  || fail_test

```

lberk: I added TLS support for the tests before seeing this comment.  Because there are already both -- I just opted to move the auth creation before the channel tests and they all pass.  Please let me know if you'd prefer that reworked (but figure, a couple extra secrets isn't the worst tradeoff? I could be wrong :)

matzew (approved):
/lgtm
/approve
/hold

@aliok mind giving this a final look ?

pkg/source/client.go
```diff
@@ -65,8 +66,22 @@ func NewConfig(ctx context.Context) ([]string, *sarama.Config, error) {
...
+
+		if env.Net.SASL.Type == sarama.SASLTypeSCRAMSHA256 {
+			cfg.Net.SASL.SCRAMClientGeneratorFunc = func() sarama.SCRAMClient { return &XDGSCRAMClient{HashGeneratorFcn: SHA256} }
+			cfg.Net.SASL.Mechanism = sarama.SASLTypeSCRAMSHA256
+		}
+
+		if env.Net.SASL.Type == sarama.SASLTypeSCRAMSHA512 {
+			cfg.Net.SASL.SCRAMClientGeneratorFunc = func() sarama.SCRAMClient { return &XDGSCRAMClient{HashGeneratorFcn: SHA512} }
+			cfg.Net.SASL.Mechanism = sarama.SASLTypeSCRAMSHA512
+		}
```
aliok, line 84: Seeing this made me realize we need validation for the type.

If the user makes a typo in the YAML, the code will start using `sarama.SASLTypePlaintext` but user won't notice that.

What do you think? Just validate 3 possible options in the webhook and reject the CR if the value is anything else?
I wouldn't perhaps create the webhook for this, but the webhook is already there.

In the webhook,

pkg/source/client_test.go
```diff
@@ -334,16 +334,160 @@ func generateCert(t *testing.T) (string, string) {
...
+			env: map[string]string{
+				"KAFKA_BOOTSTRAP_SERVERS": defaultBootstrapServer,
+			},
+			bootstrapServer: "ImADoctorNotABootstrapServerJim!",
+			wantErr:         true,
+		},
+		/*
+			TODO
+			"Multiple bootstrap servers": {
+				env: map[string]string{
```
aliok, line 367: Maybe delete this comment?

aliok:
I just have a comment about the validation of the new `type` field but other than that it looks good to me.

pkg/source/client.go
matzew: oh, yes - good catch !

pkg/source/client_test.go
lberk: could you elaborate on why? is it not a valid use case that we should be testing? (was going to make the tests more difficult to run/outside the scope of this PR, but I think it should still be tracked for future improvements)

aliok: Nah, not because of that. It is because it is a TODO. If this is a valid thing and you want to do it as a follow up, maybe create a ticket?

lberk: https://github.com/knative-sandbox/eventing-kafka/issues/240

pkg/source/client.go
lberk: Perhaps I'm misreading the code, but, do we have a validating webhook for the source bits?
https://github.com/knative-sandbox/eventing-kafka/blob/master/cmd/webhook/main.go#L49

lets do this in a later PR please

matzew: @lberk feel free to "unhold" when ready

lberk:
I believe all the comments are addressed, unholding
/unhold

pkg/source/client.go
aliok: Nah, it is here: https://github.com/knative-sandbox/eventing-kafka/blob/82c096577d36be37a12624367b0ebc96b9d9506e/cmd/source/controller/main.go#L194

The source has the controller and the webhook in the same pod.

But, +1 on merging w/o it and doing as follow up though.

PR #239, opened 2020-11-27
Close request body for tcp connection reuse

tangyanhan:
Fixes #

Missing Close() for unused response body blocked tcp connections from being reused within process scope. This significantly reduces performance.

## Proposed Changes

- 🧽 Improved performance of delivering messages

### DIFF
--- a/pkg/source/adapter/adapter.go
+++ b/pkg/source/adapter/adapter.go
@@ -136,6 +136,10 @@ func (a *Adapter) Handle(ctx context.Context, msg *sarama.ConsumerMessage) (bool
 		a.logger.Debug("Error while sending the message", zap.Error(err))
 		return false, err // Error while sending, don't commit offset
 	}
+	// Always try to close body so the connection can be reused afterwards
+	if res.Body != nil {
+		res.Body.Close()
+	}
 
 	if res.StatusCode/100 != 2 {
 		a.logger.Debug("Unexpected status code", zap.Int("status code", res.StatusCode))


tangyanhan:
> @googlebot I signed it!

pkg/source/adapter/adapter.go
```diff
@@ -136,6 +136,7 @@ func (a *Adapter) Handle(ctx context.Context, msg *sarama.ConsumerMessage) (bool
 		a.logger.Debug("Error while sending the message", zap.Error(err))
 		return false, err // Error while sending, don't commit offset
 	}
+	res.Body.Close()
```
matzew, line 139: how about `nil` check ? 

and move that to a function, eg:

```
func safeCloseHTTPResponseBody(response *http.Response) 
  if response != nil && response.Body != nil {
    err := response.Body.Close()
    if err != nil {
      c.logger.Error("Failed To Close HTTP Response Body", zap.Error(err))
    }
  }
}
```

than `	defer safeCloseHTTPResponseBody(response)`  instead of the direct ? 


@slinkydeveloper have we smth like that already in the kncloudevents ? :thinking:

slinkydeveloper: Yep you miss a nil check, but i don't think we need to do a check on the body.Close error. If there's a broken pipe we don't care anyway, we don't read that body.

> @slinkydeveloper have we smth like that already in the kncloudevents ?

We don't have such function

matzew:
@tangyanhan wanna follow up on the nil comment?

tangyanhan:
> @tangyanhan wanna follow up on the nil comment?

In comment of go source `net/http/response.go`, the response body is always non-nil, and checking of close errors seems not essential as well, for we just take use of the status code.
```
	// The http Client and Transport guarantee that Body is always
	// non-nil, even on responses without a body or responses with
	// a zero-length body. It is the caller's responsibility to
	// close Body. The default HTTP client's Transport may not
	// reuse HTTP/1.x "keep-alive" TCP connections if the Body is
	// not read to completion and closed.
```

slinkydeveloper:
> the response body is always non-nil

@tangyanhan I would rather prefer not depend on this implementation detail, since it may change in future and break this code

tangyanhan:
> > the response body is always non-nil
> 
> @tangyanhan I would rather prefer not depend on this implementation detail, since it may change in future and break this code

That make sense. Code updated.

PR #242, opened 2020-11-30
Fix label consistency

skonto:
## Proposed Changes

- Label source name has a different spelling in [eventing repo](https://github.com/knative/eventing/blob/241af67e1844c9363869ca4141b910d0bc9819c6/pkg/reconciler/apiserversource/resources/labels.go#L28)
- We should keep things consistent because tooling may rely on these labels etc. In my case this makes hard to parse metric labels.
- Ping source btw seems to have its [own special label](https://github.com/knative/eventing/blob/241af67e1844c9363869ca4141b910d0bc9819c6/pkg/reconciler/pingsource/resources/labels.go#L28).

**Release Note**

```release-note
NONE
```

### DIFF
--- a/pkg/source/reconciler/source/resources/labels.go
+++ b/pkg/source/reconciler/source/resources/labels.go
@@ -25,6 +25,6 @@ const (
 func GetLabels(name string) map[string]string {
 	return map[string]string{
 		"eventing.knative.dev/source":     controllerAgentName,
-		"eventing.knative.dev/SourceName": name,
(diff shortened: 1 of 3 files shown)


lionelvillard:
The pingsource labels are not used anymore. Instead the labels are directly set on the deployment config: https://github.com/knative/eventing/blob/master/config/core/deployments/pingsource-mt-adapter.yaml

/ok-to-test
/lgtm

matzew (approved):
/lgtm
/approve

PR #249, opened 2020-12-02
Backport: Set subscription readiness based on consumer group status (#182)

### LINKED ISSUE #98: [consolidated] Dispatcher deployment readonly

### LINKED ISSUE #134: Consolidated KafkaChannel marks Subscriptions Ready too early

devguyio:
Backport of #182 
Fixes #134
Fixes #98 
Fixes knative/eventing-contrib#1556

## Proposed Changes

- Set subscription readiness based on consumer group status in the consolidated Kafka
- Set's KafkaChannel consolidated dispatcher to read-only
- Enables HA in KafkaChannel consolidated dispatcher

**Release Note**

```release-note
- A bug was fixed in the consolidated KafkaChannel where subscriptions would show up in the channel's `status.subscribers` before the dispatcher becomes ready to dispatch messages for those subscribers.
- Consolidated KafkaChannel dispatcher's horizontal scalability works now seamlessly with reconciler leader election.
```

### DIFF
--- a/cmd/channel/consolidated/dispatcher/main.go
+++ b/cmd/channel/consolidated/dispatcher/main.go
@@ -35,8 +35,5 @@ func main() {
 		ctx = injection.WithNamespaceScope(ctx, ns)
 	}
 
-	// Do not run the dispatcher in leader-election mode
-	ctx = sharedmain.WithHADisabled(ctx)
-
 	sharedmain.MainWithContext(ctx, component, controller.NewController)
 }
--- a/pkg/channel/consolidated/kafka/admin.go
+++ b/pkg/channel/consolidated/kafka/admin.go
@@ -1,0 +1,109 @@
+/*
+Copyright 2020 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kafka
+
+import (
+	"context"
+	"fmt"
+	"math"
+	"sync"
+	"time"
+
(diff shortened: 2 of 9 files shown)


matzew:
Nice seeing this backport, @devguyio 

/lgtm

/assign @pierDipi 
/assign @lionelvillard 

for final reviews/approvalz

pierDipi (approved):
/lgtm

PR #253, opened 2020-12-03
Increase health check timeouts to allow Sarama timeouts

eric-sap:
The health (liveness and readiness) checks in the distributed channel caused a restart of the pod in 25 seconds (10 + 3*5) if Sarama was not ready by that point.  The timeout for the TCP dialer defaults to 30 seconds (in sarama/config.go) so the pod would restart (and take any logging information with it) without the administration being able to see the error response in the logs.

## Proposed Changes

- 🧽  Increase initial health check timeouts to cover the default Net.Dialer.Timeout that sarama uses when connecting via the broker string to 45 seconds (30 + 5*3)

### DIFF
--- a/pkg/channel/distributed/controller/constants/constants.go
+++ b/pkg/channel/distributed/controller/constants/constants.go
@@ -88,12 +88,12 @@ const (
 
 	// Health Configuration
 	HealthPort                = 8082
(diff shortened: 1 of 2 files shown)


pierDipi (commented):
/ok-to-test

PR #265, opened 2020-12-09
Fix error format in adapter

antoineco:
## Proposed Changes

Someone reported that error in Slack, and the repetition of "occurred" caught my attention.

I also shortened the message while I was at it. Less is more.

**Release Note**

```release-note
```

**Docs**


matzew (approved):
/lgtm
/approve

Thanks dude!

PR #273, opened 2020-12-11
Read resp body for connection reuse

tangyanhan:
Fixes #

This PR works as a complement for [Pull 239](https://github.com/knative-sandbox/eventing-kafka/pull/239). 
Inorder to reuse TCP connection for HTTP requests with `KeepAlive: true`, request body MUST be read and closed manually. Otherwise, lots of tcp connections will be created and stays TIME_WAIT, resulting errors like "cannot assign requested address".

Without this code change, users may adjust sysctl parameters below as workaround:
```
net.ipv4.tcp_tw_reuse=1
net.ipv4.tcp_fin_timeout=15
```

### DIFF
--- a/pkg/source/adapter/adapter.go
+++ b/pkg/source/adapter/adapter.go
@@ -18,6 +18,8 @@ package kafka
 
 import (
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


matzew:
/ok-to-test
/lgtm
perhaps we could enhance the comment a bit?

/assign @slinkydeveloper

slinkydeveloper:
@tangyanhan isn't closing enough to discard all the incoming body? Is there any doc to look at?

lionelvillard:
There is a blog about it: http://tleyden.github.io/blog/2016/11/21/tuning-the-go-http-client-library-for-load-testing/

slinkydeveloper:
Is this still valid? post is from 2016 and to me this clearly sounds like a bug of golang std lib

lionelvillard:
It still valid. I could reproduce with go 1.14.2.

PR #275, opened 2020-12-11
Remove default limits

eric-sap:
## Proposed Changes

- 🎁 Allow not having the cpu/memory resource requests and/or limits in the config-kafka configmap for eventing-kafka.receiver and eventing-kafka.dispatcher
- 🧽 Default to no limit for receiver and dispatcher (but still allow a limit if defined by the admin in the config-kafka configmap) and a reasonable (100m / 50Mi) initial request size based on empirical small-cluster and small-message-size testing.

### DIFF
--- a/pkg/channel/distributed/controller/kafkachannel/reconciler_test.go
+++ b/pkg/channel/distributed/controller/kafkachannel/reconciler_test.go
@@ -237,7 +237,7 @@ func TestReconcile(t *testing.T) {
 				controllertesting.NewKafkaChannelSuccessfulReconciliationEvent(),
 			},
 			OtherTestData: map[string]interface{}{
-				"configOptions": []controllertesting.KafkaConfigOption { controllertesting.WithNoDispatcherResources },
+				"configOptions": []controllertesting.KafkaConfigOption{controllertesting.WithNoDispatcherResources},
 			},
 		},
 
--- a/pkg/channel/distributed/controller/kafkachannel/reconciler_test.go
+++ b/pkg/channel/distributed/controller/kafkachannel/reconciler_test.go
@@ -192,6 +192,56 @@ func TestReconcile(t *testing.T) {
 		},
 
 		//
+		// Full Reconciliation (No Dispatcher Resource Requests Or Limits)
+		//
+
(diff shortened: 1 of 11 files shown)


pkg/channel/distributed/controller/kafkachannel/reconciler_test.go
```diff
@@ -191,6 +191,56 @@ func TestReconcile(t *testing.T) {
 			},
 		},
 
+		//
+		// Full Reconciliation (No Dispatcher Resource Requests Or Limits)
+		//
+
```
travis-minke-sap, line 197: This shouldn't be a new "section" of tests with the triple comment separator - it's just another Complete Reconciliation test instance - remove comment and change the two test names to be different.

pkg/channel/distributed/controller/kafkasecret/reconciler_test.go
```diff
@@ -119,7 +119,34 @@ func TestReconcile(t *testing.T) {
...
+			WantPatches: []clientgotesting.PatchActionImpl{controllertesting.NewKafkaSecretFinalizerPatchActionImpl()},
+			WantEvents: []string{
+				controllertesting.NewKafkaSecretFinalizerUpdateEvent(),
+				controllertesting.NewKafkaSecretSuccessfulReconciliationEvent(),
+			},
+			OtherTestData: map[string]interface{}{
+				"configOptions": []controllertesting.KafkaConfigOption{controllertesting.WithNoReceiverResources},
+			},
+		},
 		//
```
travis-minke-sap, line 150: maintain linefeed separators (please ; )

pkg/channel/distributed/controller/kafkachannel/reconciler_test.go
eric-sap: Done; not sure why I did that in the first place.

pkg/channel/distributed/controller/kafkasecret/reconciler_test.go
eric-sap: certainly

PR #278, opened 2020-12-15
Move Sarama YAML related functions into common

### LINKED ISSUE #268: Common util for SaramaConfig
**Problem**

the `distributed` channel has a good way to configure Sarama via values from a Configmap.

We want to move that to a common util, to be used in both: `distributed` and `consolidated`

It will provide a common way of handling (watching/parsing) of ConfigMap with full Sarama.Config. (and ability to merge in TLS/SASL info from `consolidated`)
common code for the sarama config

aliok:
Related to #268

## Proposed Changes

- Move some Sarama config related functions to a common package
- Make the `MergeSaramaSettings` function work with YAML-string instead of configmap
- Will follow up with more PRs that unify the behavior on both channels and the source around Sarama config by modifying `MergeSaramaSettings` as necessary and using it as the central place for Sarama config building

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/channel/distributed/controller/kafkachannel/reconciler.go
+++ b/pkg/channel/distributed/controller/kafkachannel/reconciler.go
@@ -32,6 +32,7 @@ import (
 	"knative.dev/eventing-kafka/pkg/channel/distributed/common/config"
 	kafkaadmin "knative.dev/eventing-kafka/pkg/channel/distributed/common/kafka/admin"
 	kafkasarama "knative.dev/eventing-kafka/pkg/channel/distributed/common/kafka/sarama"
+	"knative.dev/eventing-kafka/pkg/channel/distributed/common/testing"
 	"knative.dev/eventing-kafka/pkg/channel/distributed/controller/constants"
 	"knative.dev/eventing-kafka/pkg/channel/distributed/controller/env"
 	"knative.dev/eventing-kafka/pkg/channel/distributed/controller/event"
@@ -39,6 +40,7 @@ import (
 	kafkaclientset "knative.dev/eventing-kafka/pkg/client/clientset/versioned"
 	"knative.dev/eventing-kafka/pkg/client/injection/reconciler/messaging/v1beta1/kafkachannel"
 	kafkalisters "knative.dev/eventing-kafka/pkg/client/listers/messaging/v1beta1"
+	"knative.dev/eventing-kafka/pkg/common/client"
 	kubeclient "knative.dev/pkg/client/injection/kube/client"
 	"knative.dev/pkg/reconciler"
 )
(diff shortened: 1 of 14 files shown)


pkg/common/client/config.go
```diff
@@ -0,0 +1,73 @@
+package client
```
matzew, line 1: license header

matzew:
> Will follow up with more PRs that unify the behavior on both channels 

that includes also using it on the `consolidated` channel, right ?

aliok:
> that includes also using it on the consolidated channel, right ?
Yes

pkg/common/client/config.go
aliok: added them now

aliok:
@matzew  ready for another go. please ignore coverage failures

aliok:
cc @travis-minke-sap

pkg/channel/distributed/controller/kafkachannel/reconciler.go
```diff
@@ -242,7 +244,17 @@ func (r *Reconciler) configMapObserver(configMap *corev1.ConfigMap) {
 	// environment will also need to be re-parsed here.
 
 	// Load the Sarama settings from our configmap, ignoring the eventing-kafka result.
-	saramaConfig, err := kafkasarama.MergeSaramaSettings(nil, configMap)
+
```
travis-minke-sap, line 247: Minor nit - looks like the comment on line 246 was abandoned - probably belongs down below on line 257

pkg/channel/distributed/receiver/producer/producer.go
```diff
@@ -21,6 +21,8 @@ import (
 	"errors"
 	"time"
 
+	"knative.dev/eventing-kafka/pkg/channel/distributed/common/testing"
+	"knative.dev/eventing-kafka/pkg/common/client"
 	"knative.dev/eventing-kafka/pkg/common/tracing"
```
travis-minke-sap, line 26: Why is the `distributed/common/testing` import being added?  Doesn't appear to be used and would be bad to include test code in production if it was.  Assuming it can be removed?

travis-minke-sap:
Looks good to me (thanks Ali!) I'm adding a hold just so that the testing import can be removed - I am fine with someone removing the hold once that's resolved or ping me and I'll do it if I'm around ; )

/lgtm
/hold

matzew:
@aliok mind addressing the comments and we merge this?

pkg/channel/distributed/receiver/producer/producer.go
aliok: I also noticed that. But, it is already used in head's `sarama.go` in here: https://github.com/knative-sandbox/eventing-kafka/blob/master/pkg/channel/distributed/common/kafka/sarama/sarama.go#L37

My refactor is simply moving some stuff out that file to another place. However, I also made the function's scope smaller and made it operate on YAML-string instead of reading stuff from configmap. The configmap reading now is done by the caller and in this case, in this `producer.go` file. That's why the import is added.

IMO, let's tackle that separately, because there's more of this import in files also not changed in this PR.

pkg/channel/distributed/controller/kafkachannel/reconciler.go
aliok: +1

pkg/channel/distributed/receiver/producer/producer.go
travis-minke-sap: OK, I didn't realize that - bad form on our part and as you suggest we can correct it ourselves - thanks!

aliok: >  bad form on our part and as you suggest we can correct it ourselves 

@travis-minke-sap sorry, I didn't mean that. I meant, WE can do it together, but not part of this PR ;) Wasn't blaming anybody OR wasn't saying "you fix it"

PR #286, opened 2020-12-18
Unifiy Sarama config building

### LINKED ISSUE #268: Common util for SaramaConfig

aliok:
Fixes #268

Follow up for https://github.com/knative-sandbox/eventing-kafka/pull/278

## Proposed Changes

- Move Sarama config building into `common` package
- Made the functions that modify the config more generic so that we can use them in both channels and the source
- Get rid of too specific functions for a module (like `UpdateSaramaConfig` function only made sense for the distributed channel but it could be replaced with a function that is more generic and reusable)
- Added more unit tests for the Sarama config building

**Release Note**

```release-note

```

**Docs**


aliok:
Gonna add PR desc later. Let me see first if things are green.
Lacking some unit tests anyway.

cmd/channel/distributed/dispatcher/main.go
```diff
@@ -77,8 +77,21 @@ func main() {
 		logger.Fatal("Failed To Load Sarama Settings", zap.Error(err))
 	}
 
+	client.InitSaramaConfig(saramaConfig)
+	saramaConfig.ClientID = constants.Component
+
 	// Update The Sarama Config - Username/Password Overrides (EnvVars From Secret Take Precedence Over ConfigMap)
-	client.UpdateSaramaConfig(saramaConfig, constants.Component, environment.KafkaUsername, environment.KafkaPassword)
```
aliok, line 81: This function was too specific. Got rid of it and changed its usages with using more generic `UpdateSaramaConfigWithKafkaAuthConfig`

cmd/channel/distributed/receiver/main.go
```diff
@@ -80,8 +80,21 @@ func main() {
 		logger.Fatal("Failed To Load Sarama Settings", zap.Error(err))
 	}
 
-	// Update The Sarama Config - Username/Password Overrides (EnvVars From Secret Take Precedence Over ConfigMap)
-	client.UpdateSaramaConfig(saramaConfig, constants.Component, environment.KafkaUsername, environment.KafkaPassword)
```
aliok, line 84: Same story here

pkg/channel/consolidated/dispatcher/dispatcher.go
```diff
@@ -90,13 +89,15 @@ func (sub Subscription) String() string {
...
-	conf.Consumer.Return.Errors = true    // Returns the errors in ConsumerGroup#Errors() https://godoc.org/github.com/Shopify/sarama#ConsumerGroup
-	conf.Producer.Return.Successes = true // Must be enabled for sync producer
+	// explicit version
+	conf.Version = sarama.V2_0_0_0
+
+	client.InitSaramaConfig(conf)
 
 	// append the auth info:
-	err := source.UpdateSaramaConfigWithKafkaAuthConfig(conf, args.KafkaAuthConfig)
+	conf, err := client.BuildSaramaConfig(conf, "", args.KafkaAuthConfig)
```
aliok, line 100: Once we decide to use the Sarama settings from `config-kafka`, we can simply pass it here (2nd arg).

pkg/channel/consolidated/utils/util.go
```diff
@@ -62,24 +63,7 @@ type KafkaConfig struct {
...
-	Usercert string
-	Userkey  string
-}
-
-type KafkaSaslConfig struct {
-	User     string
-	Password string
-	SaslType string
-}
-
```
aliok, line 81: Moved these into `common` to be able to reuse them in 2 channels and the source.

pkg/channel/distributed/common/kafka/sarama/sarama.go
```diff
@@ -47,40 +43,6 @@ func EnableSaramaLogging(enable bool) {
 	}
 }
 
-// ConfigEqual is a convenience function to determine if two given sarama.Config structs are identical aside
-// from unserializable fields (e.g. function pointers).  To ignore parts of the sarama.Config struct, pass
-// them in as the "ignore" parameter.
-func ConfigEqual(config1, config2 *sarama.Config, ignore ...interface{}) bool {
```
aliok, line 53: moved to `common`

pkg/channel/distributed/dispatcher/dispatcher/dispatcher.go
```diff
@@ -283,7 +283,7 @@ func (d *DispatcherImpl) ConfigChanged(configMap *v1.ConfigMap) Dispatcher {
 	// Merge The ConfigMap Settings Into The Provided Config
 	saramaSettingsYamlString := configMap.Data[testing.SaramaSettingsConfigKey]
 
-	newConfig, err := client.MergeSaramaSettings(nil, saramaSettingsYamlString)
+	newConfig, err := client.BuildSaramaConfig(nil, saramaSettingsYamlString, nil)
```
aliok, line 286: If we have the Kafka auth in channel CRD for the distributed channel in the future, we can just pass it here.

aliok:
@travis-minke-sap @eric-sap 
May I ask your eyes on this please?
Ready for review.

travis-minke-sap:
/lgtm
/hold

* Holding in case @devguyio wants more time to review?

pkg/common/client/config.go
```diff
@@ -38,52 +39,202 @@ func MergeSaramaSettings(config *sarama.Config, saramaSettingsYamlString string)
 		config.Version = constants.ConfigKafkaVersionDefault
 
 		// Add Any Required Settings
-		UpdateSaramaConfig(config, config.ClientID, "", "")
+		InitSaramaConfig(config)
```
devguyio, line 42: I'd really avoid having functions with strong side effects like this. Again, unless I know  InitSaramaConfig internals, I'd be surprised that I'm passing an object to a function so that it sets a couple of default settings on the object I'm passing. I think we should avoid this style of functions with heavy side effects. 

alternatives I can think of is setting the default settings explicitly,  or my least favorite, call it something that describes its intent (e.g `UpdateConfigWithDefaults` or `SetConfigDefaults` )

pkg/common/client/config.go
```diff
@@ -38,52 +39,202 @@ func MergeSaramaSettings(config *sarama.Config, saramaSettingsYamlString string)
...
-	if err != nil {
-		return nil, fmt.Errorf("ConfigMap's sarama value could not be converted to a Sarama.Config struct: %s : %v", err, saramaSettingsYamlString)
-	}
+	// Return Success
+	return config, nil
+}
 
-	// Override The Custom Parsed KafkaVersion
-	config.Version = kafkaVersion
+func InitSaramaConfig(config *sarama.Config) {
```
devguyio, line 84: 1. I think the name isn't reflecting what's happening here as mentioned on the comment above. 
2. Would be good to have a comment on the function.

pkg/common/client/config.go
```diff
@@ -38,52 +39,202 @@ func MergeSaramaSettings(config *sarama.Config, saramaSettingsYamlString string)
...
+
+	ignoredUnexported := cmpopts.IgnoreUnexported(config1.Version, x509.CertPool{}, tls.Config{})
+
+	// Compare the two sarama config structs, ignoring types and unexported fields as specified
+	return cmp.Equal(config1, config2, ignoredTypes, ignoredUnexported)
 }
 
-// Utility Function For Configuring Common Settings For Admin/Producer/Consumer
-func UpdateSaramaConfig(config *sarama.Config, clientId string, username string, password string) {
+type KafkaAuthConfig struct {
```
devguyio, line 161: It would be great to keep the type declaration at the top of the file

pkg/channel/consolidated/dispatcher/dispatcher.go
```diff
@@ -90,13 +89,15 @@ func (sub Subscription) String() string {
 
 func NewDispatcher(ctx context.Context, args *KafkaDispatcherArgs) (*KafkaDispatcher, error) {
 	conf := sarama.NewConfig()
-	conf.Version = sarama.V2_0_0_0
+	// don't get the clientID cleared by when it is built later
```
devguyio, line 92: I'm not sure I understand this comment

devguyio (changes requested):
Nice work @aliok ! I can imagine it wasn't easy to figure out a unified approach. 

My 2 cents, I think your mind is in the right place, from what I can see you're trying to satisfy the different config needs of the different Kafka components but maybe we can take another stab at designing that config package and making sure it's well documented and easily understandable? cause it's used in three different components by different groups of maintainers. So check out my comments and lmk if they make sense. 

(e.g I can very clearly see that this is a typical builder pattern use case `...NewConfig().WithXYZ()` etc )

pkg/common/client/config.go
```diff
@@ -18,17 +18,18 @@ package client
...
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"knative.dev/eventing-kafka/pkg/channel/distributed/common/kafka/constants"
 )
 
-// Merge existing Sarama config with provided YAML string.
-// Values in YAML string will override the values in the config.
-// If config Is nil, A New sarama.Config Struct Will Be Created With Default Values
-func MergeSaramaSettings(config *sarama.Config, saramaSettingsYamlString string) (*sarama.Config, error) {
+func BuildSaramaConfig(config *sarama.Config, saramaSettingsYamlString string, kafkaAuthCfg *KafkaAuthConfig) (*sarama.Config, error) {
```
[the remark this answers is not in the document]
aliok: Hey, thanks for the review. You have a lot of good points.

It was kind of hard to refactor this. I initially when with a builder pattern locally, then saw it ended up too much duplicated code in the callers of the builder and changed my mind. I still think we shouldn't convert this into a builder.

To discuss individually:

1. This is about Sarama's, the Kafka client's, config. So, I named it as `common/client/config` but I don't have a strong opinion. I can move stuff around.
2. Ok, let me check that.
3. Gonna rework on the docs and also the naming of the functions.

pkg/common/client/config.go
aliok: I will go with `UpdateConfigWithDefaults` then.
Sounds good.

pkg/common/client/config.go
aliok: As above, I will rename this to `UpdateConfigWithDefaults` and add comments.

pkg/common/client/config.go
aliok: Sure

pkg/channel/consolidated/dispatcher/dispatcher.go
aliok: Let me add more explanation here.
But, it is about an explicit setting of the clientId here in the caller of the `BuildSaramaConfig`.

aliok:
@devguyio switched to builder pattern. please have a new look

matzew (approved):
/lgtm

/hold

@devguyio please take a look and unhold if your comments are all addressed

pkg/common/client/config.go
```diff
@@ -18,72 +18,306 @@ package client
...
+	// Build builds the Sarama config with the given builder config
+	Build() (*sarama.Config, error)
+}
+
+func NewConfigBuilder() ConfigBuilder {
+	return &configBuilder{}
+}
+
+type configBuilder struct {
+	existing *sarama.Config
```
devguyio, line 92: is there a reason of using a pointer here?

pkg/common/client/config.go
```diff
@@ -18,72 +18,306 @@ package client
...
+	defaults bool
+	version  *sarama.KafkaVersion
```
devguyio, line 94: same pointer question

pkg/common/client/config.go
```diff
@@ -18,72 +18,306 @@ package client
...
+	clientId string
+	yaml     string
+	auth     *KafkaAuthConfig
```
devguyio, line 97: same pointer question

pkg/common/client/yaml.go
```diff
@@ -36,10 +35,11 @@ var regexRootPEMs = regexp.MustCompile(`(?s)\s*RootPEMs:.*-----END CERTIFICATE--
...
 // the yaml into.  Therefore, we instead support the user providing a 'Version' string of the
 // format "major.minor.patch" (e.g. "2.3.0") which we will "extract" out of the YAML string
-// and return as an actual sarama.KafkaVersion.  In the case where the user has NOT specified
-// a Version string we will return a default version.
+// and return as an actual sarama.KafkaVersion.
 //
-func extractKafkaVersion(saramaConfigYamlString string) (string, sarama.KafkaVersion, error) {
+// In the case where the user has NOT specified, nil will be returned.
+//
+func extractKafkaVersion(saramaConfigYamlString string) (string, *sarama.KafkaVersion, error) {
```
devguyio, line 42: why return a mix of value and pointer? probably the pointer isn't needed

devguyio (changes requested):
Great job! I like the design! Just some cleanup regarding pointer-value semantics and it'd be ready to go!

devguyio (approved):
/lgtm
/unhold

PR #290, opened 2020-12-29
Add distributed traceid

eric-sap:
## Proposed Changes

-🎁  Use loggers from context in controller reconciler code (includes a traceid already provided by eventing)

### DIFF
--- a/cmd/channel/distributed/receiver/main.go
+++ b/cmd/channel/distributed/receiver/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"context"
 	"flag"
+	"knative.dev/pkg/logging/logkey"
 	nethttp "net/http"
 	"strconv"
 	"strings"
@@ -160,11 +161,15 @@ func flush(logger *zap.Logger) {
 // CloudEvent Message Handler - Converts To KafkaMessage And Produces To Channel's Kafka Topic
 func handleMessage(ctx context.Context, channelReference eventingchannel.ChannelReference, message binding.Message, transformers []binding.Transformer, _ nethttp.Header) error {
 
+	// Add a traceId to the logger so that we can more easily follow a particular reconciliation in the log
+	traceLogger := logger.With(zap.String(logkey.TraceID, uuid.New().String()))
+	ctx = logging.WithLogger(ctx, traceLogger.Sugar())
+
 	// Note - The context provided here is a different context from the one created in main() and does not have our logger instance.
-	if logger.Core().Enabled(zap.DebugLevel) {
+	if traceLogger.Core().Enabled(zap.DebugLevel) {
 		// Checked Logging Level First To Avoid Calling zap.Any In Production
-		logger.Debug("~~~~~~~~~~~~~~~~~~~~  Processing Request  ~~~~~~~~~~~~~~~~~~~~")
-		logger.Debug("Received Message", zap.Any("ChannelReference", channelReference))
+		traceLogger.Debug("~~~~~~~~~~~~~~~~~~~~  Processing Request  ~~~~~~~~~~~~~~~~~~~~")
+		traceLogger.Debug("Received Message", zap.Any("ChannelReference", channelReference))
 	}
 
 	// Trim The "-kn-channel" Suffix From The Service Name
@@ -173,14 +178,14 @@ func handleMessage(ctx context.Context, channelReference eventingchannel.Channel
 	// Validate The KafkaChannel Prior To Producing Kafka Message
 	err := channel.ValidateKafkaChannel(channelReference)
 	if err != nil {
-		logger.Warn("Unable To Validate ChannelReference", zap.Any("ChannelReference", channelReference), zap.Error(err))
+		traceLogger.Warn("Unable To Validate ChannelReference", zap.Any("ChannelReference", channelReference), zap.Error(err))
 		return err
 	}
 
 	// Produce The CloudEvent Binding Message (Send To The Appropriate Kafka Topic)
 	err = kafkaProducer.ProduceKafkaMessage(ctx, channelReference, message, transformers...)
 	if err != nil {
-		logger.Error("Failed To Produce Kafka Message", zap.Error(err))
+		traceLogger.Error("Failed To Produce Kafka Message", zap.Error(err))
 		return err
 	}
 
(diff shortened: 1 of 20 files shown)


cmd/channel/distributed/receiver/main.go
```diff
@@ -160,11 +162,15 @@ func flush(logger *zap.Logger) {
 // CloudEvent Message Handler - Converts To KafkaMessage And Produces To Channel's Kafka Topic
 func handleMessage(ctx context.Context, channelReference eventingchannel.ChannelReference, message binding.Message, transformers []binding.Transformer, _ nethttp.Header) error {
 
+	// Add a traceId to the logger so that we can more easily follow a particular message/failure in the log
+	traceLogger := logger.With(zap.String(logkey.TraceID, uuid.New().String()))
+	ctx = logging.WithLogger(ctx, traceLogger.Sugar())
+
```
travis-minke-sap, line 168: Minor nit - I wouldn't decorate the name `logger` as `tracelogger` just because we added the traceID to it.  Inconsistent with other instances below as well.

travis-minke-sap:
I don't think these changes will result in all/most of the actual reconciliation logging using the logger from ctx with the traceid right?   Meaning we should (at least in the controller) remove the logger from the reconciler struct and always get it from the ctx?

cmd/channel/distributed/receiver/main.go
eric-sap: I can name it something different, though I don't want to overwrite the package-global "logger" with the new logger containing this particular traceid

eric-sap:
> I don't think these changes will result in all/most of the actual reconciliation logging using the logger from ctx with the traceid right? Meaning we should (at least in the controller) remove the logger from the reconciler struct and always get it from the ctx?

That's true; I've now gone through and removed references to "r.logger" and the like where possible.  I wasn't able to completely remove the logger field from the kafkachannel reconciler because the configObserver function needs it (there is now a comment to that effect in the reconciler.go) but most other places could extract the proper logger from the provided Context.

pkg/channel/distributed/controller/kafkachannel/kafkachannel.go
```diff
@@ -49,6 +51,9 @@ func (r *Reconciler) reconcileKafkaChannel(ctx context.Context, channel *kafkav1
 // Reconcile The KafkaChannels MetaData (Annotations, Labels, etc...)
 func (r *Reconciler) reconcileMetaData(ctx context.Context, channel *kafkav1beta1.KafkaChannel) error {
 
+	// Get The Logger From The Context
+	logger := logging.FromContext(ctx)
+
```
travis-minke-sap, line 56: Should this be a "wrapped" ChannelLogger as is done above in `reconcileKafkaChannel` ?

pkg/channel/distributed/controller/kafkachannel/reconciler.go
```diff
@@ -49,6 +49,9 @@ import (
 
 // Reconciler Implements controller.Reconciler for KafkaChannel Resources
 type Reconciler struct {
+	// Note:  The only purpose this logger field serves is to be available for the configObserver function
+	//        which has a predefined definition that contains no context or logger parameters.  Other reconciler
+	//        logging output should be done via the Logger contained in the given Context.
 	logger               *zap.Logger
```
travis-minke-sap, line 55: Let's chat about this one - I'd rather not leave this reference here (too tempting for accidental future use) and we can possibly create a new one in the `configObserver()` code?

pkg/channel/distributed/controller/kafkachannel/reconciler.go
```diff
@@ -221,22 +228,26 @@ func (r *Reconciler) reconcile(ctx context.Context, channel *kafkav1beta1.KafkaC
...
+
 	// Enable Sarama Logging If Specified In ConfigMap
 	if ekConfig, err := kafkasarama.LoadEventingKafkaSettings(configMap); err == nil && ekConfig != nil {
 		kafkasarama.EnableSaramaLogging(ekConfig.Kafka.EnableSaramaLogging)
-		r.logger.Debug("Updated Sarama logging", zap.Bool("Kafka.EnableSaramaLogging", ekConfig.Kafka.EnableSaramaLogging))
+		logger.Debug("Updated Sarama logging", zap.Bool("Kafka.EnableSaramaLogging", ekConfig.Kafka.EnableSaramaLogging))
 	} else {
-		r.logger.Error("Could Not Extract Eventing-Kafka Setting From Updated ConfigMap", zap.Error(err))
+		logger.Error("Could Not Extract Eventing-Kafka Setting From Updated ConfigMap", zap.Error(err))
 	}
```
travis-minke-sap, line 252: Is this if/else correct?  Seems like it will ALWAY log an error?

eric-sap: I haven't changed the logic here; it should only log an error if either "err != nil" or "ekConfig == nil" according to the if condition.  Enabling and disabling the sarama logging via the configmap definitely works (or at least it worked when I tested that feature in a previous PR).

pkg/channel/distributed/controller/kafkachannel/reconciler.go
eric-sap: The only reason I didn't create a new logger in the configMapObserver is that the logger put into the Reconciler struct here is the one that comes from the context in the controller.go's NewController() function.  I don't know if there's anything in that particular logger that's important (the traceId, if present, is probably not all that useful in the configMapObserver).  If not, making a new logger in the configMapObserver function (and dropping this field) is certainly easy enough.

pkg/channel/distributed/controller/kafkachannel/kafkachannel.go
eric-sap: That's a good idea; I'll see if it works

pkg/channel/distributed/controller/kafkachannel/reconciler.go
travis-minke-sap: 👍  I read it wrong before - looks ok - thx/sorry ; )

pkg/channel/distributed/controller/kafkachannel/kafkachannel.go
eric-sap: It worked pretty well; I've moved the util.ChannelLogger call as far up the stack as is reasonable.

PR #305, opened 2021-01-13
Use Sarama config from config-kafka in consolidated channel

### LINKED ISSUE #298: Use Sarama field of config-kafka in consolidated channel and in source
**Problem**
With https://github.com/knative-sandbox/eventing-kafka/pull/286, the Sarama config building is unified.

The builder builds the config from defaults, config-kafka and KafkaAuthConfig.

However, we don't leverage the builder fully in consolidated channel and in source since we don't pass the Sarama YAML in config-kafka.

**Exit Criteria**
1. Sarama config defaults defined in config-kafka is used in the consolidated channel and the source
2. Documents updated

**Time Estimate (optional):**

aliok:
Related: #298

## Proposed Changes

- Use Sarama config from config-kafka in consolidated channel: see the added docs in the PR
- Do not do that for source yet, gonna do it as a follow up
- Some refactoring (moving common stuff around)

**Release Note**

```release-note
- 🎁 Add new feature
It is not possible to define Sarama config defaults for consolidated channel in `config-kafka` configmap with a `sarama` field.
```

**Docs**


pkg/channel/consolidated/dispatcher/dispatcher.go
```diff
@@ -163,12 +164,13 @@ func NewDispatcher(ctx context.Context, args *KafkaDispatcherArgs) (*KafkaDispat
 type TopicFunc func(separator, namespace, name string) string
 
 type KafkaDispatcherArgs struct {
-	KnCEConnectionArgs *kncloudevents.ConnectionArgs
-	ClientID           string
-	Brokers            []string
-	KafkaAuthConfig    *client.KafkaAuthConfig
-	TopicFunc          TopicFunc
-	Logger             *zap.SugaredLogger
+	KnCEConnectionArgs       *kncloudevents.ConnectionArgs
```
matzew, line 167: are these really used (via ConfigMap?)

matzew: used ... or set (via CM?)

pkg/source/client.go
```diff
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 
+	"knative.dev/eventing-kafka/pkg/channel/consolidated/utils"
```
matzew, line 23: should `utils` be in "common", eventually too ?

aliok: yeah, I have a few TODOs around that.

aliok:
I will only update the consolidated channel in this PR. I will do the same on the source in a follow up PR

pkg/channel/consolidated/dispatcher/dispatcher.go
aliok: Just checked. They're not used. Gonna remove them. Don't look like they would break anything.

matzew: before removing: Please add a ticket, that we do that 

See recent/related change on IMC for these values. IMO we should use that here too (but different PR)...
/cc @slinkydeveloper

aliok: Ok, sounds good.

aliok: https://github.com/knative-sandbox/eventing-kafka/issues/312

pkg/source/client.go
aliok: Won't do this now. I checked and there's nothing shared between the consolidated and the distributed in that package.

I moved some other stuff to a common place though.

aliok:
cc @slinkydeveloper @travis-minke-sap @eric-sap

aliok:
/hold
I think the config watching is missing. Gonna check and add if missing.

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
```diff
@@ -165,7 +165,7 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, kc *v1beta1.KafkaChannel
 		return r.kafkaConfigError
 	}
 
-	kafkaClusterAdmin, err := r.createClient()
+	kafkaClusterAdmin, err := r.createClient(logger)
```
slinkydeveloper, line 168: More than passing logger, can you pass the context? Same below, it's generally nicer to pass context, since it can contain other stuff you have in the signature.

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
```diff
@@ -592,7 +592,7 @@ func (r *Reconciler) updateKafkaConfig(ctx context.Context, configMap *corev1.Co
 	r.kafkaConfig = kafkaConfig
 	r.kafkaConfigError = err
 	ac, err := kafka.NewAdminClient(ctx, func() (sarama.ClusterAdmin, error) {
-		return source.MakeAdminClient(controllerAgentName, r.kafkaAuthConfig, kafkaConfig.Brokers)
+		return source.MakeAdminClient(controllerAgentName, r.kafkaAuthConfig, kafkaConfig, logger)
```
slinkydeveloper, line 595: Same here, pass the context as first parameter

pkg/channel/distributed/common/kafka/sarama/sarama.go
```diff
@@ -74,7 +75,7 @@ func LoadSettings(ctx context.Context, clientId string, kafkaAuthConfig *client.
 		FromYaml(saramaSettingsYamlString).
 		WithAuth(kafkaAuthConfig).
 		WithClientId(clientId).
-		Build()
+		Build(logging.FromContext(ctx))
```
slinkydeveloper, line 78: Pass the context

slinkydeveloper (commented):
Just some comments about the loggers around, otherwise lgtm

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
aliok: Sounds good, will do

aliok:
> /hold
> I think the config watching is missing. Gonna check and add if missing.

Yeah, it is missing. When I change anything in the `config-kafka` configmap, I don't see the dispatcher getting the changes OR getting recreated.

However, I will create a ticket for that. Doesn't look that related to my task I am working on here.

/unhold

pkg/common/client/config.go
```diff
@@ -221,6 +225,11 @@ func (b *configBuilder) Build() (*sarama.Config, error) {
 		config.ClientID = b.clientId
 	}
 
+	if ctx != nil {
+		logger := logging.FromContext(ctx)
+		logger.Debugf("Built Sarama config: %+v", config)
```
eric-sap, line 230: Usually with these zap loggers we would use something like this, I think.  I'm not sure if it makes any practical difference though.
```suggestion
		logger.Debug("Built Sarama config", zap.Any("config", config))
```

aliok: I never learnt the Zap logger properly, maybe there's a way to do it better but suggested change doesn't work.

That results in this:
```
Built Sarama config{config 23 0  0xc0000bb180}
```

eric-sap: Apologies; I didn't test this specific suggestion.  I haven't noticed that much of a difference when I code one way vs. the other and was only noting it for consistency across the eventing codebase.  If it doesn't work here I wouldn't push for it.

In my quick test with a different value, the mechanisms look very similar; I'm not sure why it only printed a pointer value with the sarama config.

Using configMap *corev1.ConfigMap :

Infof("zap test: %v", configMap)  ==>   "zap test: &ConfigMap{ObjectMeta:{config-kafka  knative-eventing /api/v1/namespaces/knative-eventing/..."
Info("zap test", zap.Any("test", configMap))  ==>  "zap test: {configMap 25 0  &ConfigMap{ObjectMeta:{config-kafka  knative-eventing /api/v1/namespaces/knative-eventing/..."

eric-sap: Just for curiosity's sake, I noticed this:
```
any := zap.Any("test", saramaConfig)
fmt.Print(any)   // Outputs "{saramaConfig 23 0  0xc00049ea80}"
fmt.Print(any.Interface)   // Outputs "&{{{5 100ms} 3s} {5 30s 30s 30s {false <nil>} ...."
```
Apparently the reflection calls (in fmt/print.go) don't like the zap.Field with a zapcore.ReflectType in it (which zap.Any returns) -- for whatever reason it doesn't delve into unpacking the entire zap.Field interface to get to the zap.Field.Interface where the saramaConfig is actually stashed.

When using zap.Any with a *corev1.ConfigMap, the zap.Field.Type is "StringerType" and not "ReflectType" which causes it to go down a different path.

aliok: Any changes needed in the code? Sorry wasn't clear in the comment :)

eric-sap: No changes; I was just expressing my confusion at the way the zap.Any() call works (or doesn't work, in this case).  Thanks!

PR #345, opened 2021-01-29
No retry on 400 response

eric-sap:
## Proposed Changes

- 🧽 Removed 400 ("Bad Request") from the list of responses in the distributed channel dispatcher's handler that indicate retries.

**Reasoning**

Retrying on a 400 response was a legacy behavior that was required due to the way the filter handler was consuming 5xx responses and returning 400 in their place.  This is no longer the case and "Bad Request" responses are not generally conditions that warrant retries, so that code has been removed.

Note that there are still several 4xx responses that do cause checkRetry to return true (namely 404, 409, and 429).  409 is required unless the e2e tests are changed, and 404/429 are left for their original reasons (noted in the comments for the checkRetry function).


travis-minke-sap:
/approve
/lgtm
/hold

*holding briefly for other reviewers to have a chance ; )

PR #359, opened 2021-02-02
Adding check that requires us to just enable TLS, no need for custom certs

### LINKED ISSUE #356: Consolidated channel: SASL/PLAIN does not work with public TLS certificate
**Describe the bug**

When using a channel against a public service, with proper TLS certificates, the channel is not working, since it expects there CA Cert
https://github.com/knative-sandbox/eventing-kafka/blob/master/pkg/channel/consolidated/utils/util.go#L80

 **Expected behavior**
Works without the ca cert

 **Knative release version**
0.19+

matzew:

Fixes #356

## Proposed Changes

- with public CA Certs we just need to "enable" TLS .... and we don't have to bother with the certs like for self-signed ones

**Release Note**

```release-note
Adding tls.enabled flag for public cert usage and allowing skipping CA/User certs and key
```

**Docs**

### DIFF
--- a/pkg/channel/consolidated/utils/util.go
+++ b/pkg/channel/consolidated/utils/util.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strconv"
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -41,7 +42,7 @@ const (
 	MaxIdleConnectionsKey        = "maxIdleConns"
 	MaxIdleConnectionsPerHostKey = "maxIdleConnsPerHost"
 
-	Tls          = "tls"
+	TlsEnabled  = "tls.enabled"
 	TlsCacert    = "ca.crt"
 	TlsUsercert  = "user.crt"
 	TlsUserkey   = "user.key"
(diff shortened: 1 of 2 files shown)


pkg/channel/consolidated/utils/util.go
```diff
@@ -76,16 +77,24 @@ func GetKafkaAuthData(ctx context.Context, secretname string, secretNS string) *
...
+	// TLS
 	if string(secret.Data[TlsCacert]) != "" {
+		// We have a self-signed TLS cert
 		tls := &client.KafkaTlsConfig{
 			Cacert:   string(secret.Data[TlsCacert]),
 			Usercert: string(secret.Data[TlsUsercert]),
 			Userkey:  string(secret.Data[TlsUserkey]),
 		}
 		kafkaAuthConfig.TLS = tls
+	} else if strings.ToLower(string(secret.Data[Tls])) == strings.ToLower("enabled") {
```
aliok, line 90: Maybe change this to a boolean named `tlsEnabled`?

/lgtm
/hold

Feel free to unhold if you think it is ok as a string.

aliok:
@matzew can you add something in the PR desc for release notes?

aliok:
/lgtm

Thanks for the changes!

pkg/channel/consolidated/utils/util.go
```diff
@@ -65,27 +67,32 @@ type KafkaConfig struct {
...
-	if err != nil || secret == nil {
-		logging.FromContext(ctx).Errorf("Referenced Auth Secret not found")
-		return nil
-	}
+func parseTls(secret *corev1.Secret, kafkaAuthConfig *client.KafkaAuthConfig) {
 
-	kafkaAuthConfig := &client.KafkaAuthConfig{}
-	// check for TLS
+	// self-signed CERTs we need CA CERT, USER CERT and KEy
 	if string(secret.Data[TlsCacert]) != "" {
```
mgencur, line 73: Curious. Does this allow for client authentication (mTLS) when the CA CERT is not specified (when we use proper CA)? This code seems to ignore the TlsUsercert and TlsUserkey in this case.

matzew: we dont need them for real CAs.  tested w/ our managed kafka

maschmid: But here, https://github.com/knative-sandbox/eventing-kafka/pull/359/files#diff-0be4a9cfeff3c7254a2121d8dab22f6607fbac51700d2678da775bd841121d33R90  

we're not loading the *client* certificate ( Usercert / Userkey )

matzew: > we're not loading the client certificate ( Usercert / Userkey )

@maschmid if you need, than in the SECRET, just don't set the `tls.enabled` thing. use all three arguments, like herE:

https://github.com/knative-sandbox/eventing-kafka/blob/main/test/e2e-common.sh#L467-L471

Perhaps the name is not the best ? LOL

matzew:
@mgencur @aliok I added some doc for this now:
https://github.com/knative/docs/pull/3493

Issue #357, opened 2021-02-02
Configure MaxIdleConnection via CM ? 


matzew:
**Problem**
in IMC the `MaxIdleConection` settings can be configured via Config Map. But not for the kafka channel

**Exit Criteria**
would be good to have the same kind of tweak for this?

**Time Estimate (optional):**
How many developer-days do you think this may take to resolve?

**Additional context (optional)**
Add any other context about the feature request here.

matzew:
looking here: https://github.com/knative-sandbox/eventing-kafka/pull/353/files and wondering about that "feature"

slinkydeveloper:
We removed that from IMC for this reason: https://github.com/knative/eventing/issues/4469 :smile: If we need it, we can allow the configuration through envs

matzew:
Something like `k -n knative-eventing patch configmap/config-kafka --type=merge --patch='{"data": {"maxIdleConns": "1000", "maxIdleConnsPerHost": "1000"}}'` is possible now 

/close

PR #366, opened 2021-02-03
Increase idle connection limits from 100/2 to 1000/100

eric-sap:
## Proposed Changes

-🧽 Increase the idle connection limits to the values used by the consolidated and in-memory channels

### DIFF
--- a/cmd/channel/distributed/dispatcher/main.go
+++ b/cmd/channel/distributed/dispatcher/main.go
@@ -19,9 +19,10 @@ package main
 import (
 	"context"
 	"flag"
-	"knative.dev/eventing/pkg/kncloudevents"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


matzew (approved):
/ok-to-test

While this does
/lgtm

I'd also like to see these things being alinged, see a _similar_ PR:
https://github.com/knative-sandbox/eventing-kafka/commit/ffd2397e2552050bdbcb39f90e355ee791e5b1b9


@aliok is that something you can add to the "roadmap" for more alignement of the channels ? 

Let's capture all so we have a lot of tasks for moving this forward

matzew:
Related: https://github.com/knative-sandbox/eventing-kafka/issues/357 

I think we should likely also think about ENV_VARs ?

Issue #363, opened 2021-02-03
Auth Secret not actively watched 


matzew:
**Describe the bug**

Providing the `config-kafka` ConfigMap with a `Secret` that does not exist, starting to cause the `KafkaChannel` problems, and prevents it from working:

```
[...]

Issue #318, opened 2021-01-19
Consolidated channel dispatcher is not getting the changes in config-kafka during runtime


aliok:
**Describe the bug**
Consolidated channel dispatcher is not getting the changes in config-kafka during runtime

Main problem: channel dispatcher doesn't get the changes in `config-kafka`
- Create a consolidated channel
- Change stuff in `config-kafka`
- Changes are read by the controller (besides the problems in https://github.com/knative-sandbox/eventing-kafka/issues/393), but not reflected in the channel dispatcher

There might be the same story for KafkaSource.

 **Expected behavior**
Similar to distributed channel, watch the changes, and apply them.

2 alternatives for channel:
- Watch the configmap in the controller, recreate the dispatcher
- Watch the configmap in the dispatcher

Same story for KafkaSource.

 **To Reproduce**
- Create a consolidated channel
- Change stuff in `config-kafka`
- Changes are not reflected in the channel dispatcher

 **Knative release version**
0.20

 **Additional context**
Noticed during https://github.com/knative-sandbox/eventing-kafka/pull/305

aliok:
@slinkydeveloper @lionelvillard 

We currently mount the `config-kafka` to the dispatcher pod but when the configmap is modified, the pod doesn't know about it.

A. Watch the configmap in the controller, recreate the dispatcher (by deleting and recreating OR updating the dispatcher deployment?)
B. Watch and read the configmap in the dispatcher instead of mounting it as a volume
C. Pass the configmap as an env to the dispatcher


Which one makes more sense?

I think I would prefer C since we already do that in other places like source adapters. However, B might be the simplest with least effects.

slinkydeveloper:
What about mounting the CM as volume and then use a file watcher? That's what we do in eventing-kafka-broker: no need for role bindings and friends and easy to implement

aliok:
Sounds better. Let me check StackOverflow on how to do it :D

aliok:
Similar to https://github.com/knative-sandbox/eventing-kafka/issues/399, I will wait until I hear more opinions on ML if we should implement this or not.

At least I fixed the problems in the controller regarding watching the configmaps.

aliok:
For a unified channel implementation, I want to implement this also in consolidated channel if it exists in distributed channel.

ML discussion here: https://groups.google.com/g/knative-dev/c/VQuCXfAl8JQ

aliok:
/reopen

[...]

PR #373, opened 2021-02-09
Watch kafka secret update auth

### LINKED ISSUE #370: Distributed receiver and dispatcher do not react to auth changes in secret
**Describe the bug**
The distributed channel's receiver and dispatcher deployments are generated from values in the kafka-cluster secret once and never updated, because these components do not watch the secret for changes.

 **Expected behavior**
Changes to the kafka-cluster secret should result in recreation of the relevant objects (Producer, Dispatcher, Consumer Groups) with the new auth settings.

 **To Reproduce**
Change the auth settings in the kafka-cluster secret and note that the current receiver and dispatcher take no action based on this.

 **Knative release version**
0.19+

 **Additional context**
The configMapObserver and related ConfigChanged functions for the distributed receiver and dispatcher already have logic to perform the required component restarts, but they currently always use whatever the pod startup variables were for auth settings.  Fixing this bug requires (code locations need not be exactly these suggestions but should be a good starting point):

- Adding a mechanism for watching the secret (probably in distributed/common/config)
- New secretObserver functions in the receiver and dispatcher's main.go (or other high-level code)
- New SecretChanged() functions in the distributed/receiver/producer/producer.go and distributed/dispatcher/dispatcher/dispatcher.go
- The ConfigChanged and SecretChanged functions will share a fair bit of logic; moving it to helper functions may help prevent too much duplicated code

Not entirely necessary but would be nice to do while working this issue would be to change the variables created in the deployments for the dispatcher and receiver to include the name/namespace of the auth secret (if not completely changing to the "secret is identified in the configmap" mechanism that is eventually desired) instead of passing the initial brokers, username, password, and SASL type as separate environment variables.

eric-sap:
Fixes #370

## Proposed Changes

- 🎁  The brokers string has been moved from the kafka-cluster secret to the kafka.brokers section of the eventing-kafka field in the config-kafka configmap.  A placeholder value has been added to config/channel/distributed/300-eventing-kafka-configmap.yaml that should be replaced by the administrator.
- 🎁  The Distributed dispatcher and receiver now watch the auth secret by name/namespace and reconfigure themselves when necessary (i.e. when the username, password, or SASL type changes).  The configmap is already being watched, and changes to the broker field in it will also trigger the same reconfiguration.
- 🧽  Brokers/Username/Password/SaslType are no longer passed to the dispatcher and receiver deployments by environment variable.  Instead the secret and configmap are always used, both for initial pod startup and for watched changes.

**Release Note**

```release-note
Action Required:  If using the distributed channel, you must place your brokers string into
the config-kafka configmap in the "kafka.brokers" subsection of the eventing-kafka field.
This field is present in the default configmap provided (config/channel/distributed/300-eventing-kafka-configmap.yaml) 
with a default value of REPLACE_WITH_CLUSTER_URL.  The broker field should also be
removed from the kafka-cluster secret after the upgrade is complete.
```


matzew:
Isn't this also related to https://github.com/knative-sandbox/eventing-kafka/issues/363  that @aliok is looking at ? 

Worth to push for more "unification", so work on these "related" topics brings more value for all of us?
/cc @eric-sap @travis-minke-sap

aliok:
This is similar to https://github.com/knative-sandbox/eventing-kafka/issues/363. I think we can merge this PR that does changes only for distributed channel and then in #363 we can do the same for consolidated channel.

aliok:
> Isn't this also related to #363 that @aliok is looking at ?

I haven't started yet BTW. Currently busy with https://github.com/knative-sandbox/eventing-kafka/issues/318, yet another missing "watch" problem

pkg/channel/distributed/common/config/secretwatcher_test.go
```diff
@@ -0,0 +1,155 @@
...
+func TestGetAuthConfigFromKubernetes_Valid(t *testing.T) {
+	ctx := context.WithValue(context.TODO(), injectionclient.Key{}, clientWithTestSecret(t, commontesting.SecretName))
+	kafkaAuth, err := GetAuthConfigFromKubernetes(ctx, commontesting.SecretName, commontesting.SystemNamespace)
+	assert.Nil(t, err)
+	assert.NotNil(t, kafkaAuth)
+	assert.Equal(t, kafkaAuth.Brokers, commontesting.OldAuthBrokers)
+	assert.Equal(t, kafkaAuth.SASL.User, commontesting.OldAuthUsername)
+	assert.Equal(t, kafkaAuth.SASL.Password, commontesting.OldAuthPassword)
+	assert.Equal(t, kafkaAuth.SASL.SaslType, commontesting.OldAuthSaslType)
+}
```
travis-minke-sap, line 51: Slight nit but to improve test "robust-ness"...

There's an assumption being made that the commontesting.OldAuthXXX values ARE in fact what the clientWithTestSecret() is using.  Might be good to decompose that a bit and get the secret directly in the test and pass that to clientWithTestSecret() and then you can validate agains the values from the testsecret object without having to "know" what those might be.

cmd/channel/distributed/dispatcher/main.go
```diff
@@ -224,3 +223,27 @@ func configMapObserver(ctx context.Context, configMap *corev1.ConfigMap) {
...
+		logger.Debug("Dispatcher is nil during call to secretObserver; ignoring changes")
+		return
+	}
+
+	// Toss the new secret to the dispatcher for inspection and action
+	newDispatcher := dispatcher.SecretChanged(ctx, secret)
+	if newDispatcher != nil {
+		// The configuration change caused a new dispatcher to be created, so switch to that one
+		logger.Info("Producer Reconfigured; Switching To New Producer")
+		dispatcher = newDispatcher
```
travis-minke-sap, line 247: Copy/paste issue?  Should "Producer" be "Dispatcher" ?  Also is this the right terminology?  Typically Receiver/Dispatcher go together OR Producer/ConsumerGroup depending on whether it's k8s components or kafka resources - make sure this and receiver version are correct please ; )

pkg/channel/distributed/common/config/secretwatcher.go
```diff
@@ -0,0 +1,100 @@
+/*
+Copyright 2020 The Knative Authors
+
```
travis-minke-sap, line 3: I think they want these to be current 2021 - check all new files.

pkg/channel/distributed/common/env/constants.go
```diff
@@ -29,10 +29,8 @@ const (
 	ResyncPeriodMinutesEnvVarKey = "RESYNC_PERIOD_MINUTES"
 
 	// Kafka Authorization
-	KafkaBrokerEnvVarKey   = "KAFKA_BROKERS"
-	KafkaUsernameEnvVarKey = "KAFKA_USERNAME"
-	KafkaPasswordEnvVarKey = "KAFKA_PASSWORD"
-	KafkaSaslTypeEnvVarKey = "KAKFA_SASL_TYPE"
+	KafkaSecretNamespaceEnvVarKey = "KAFKA_SECRET_NAMESPACE"
+	KafkaSecretNameEnvVarKey      = "KAFKA_SECRET_NAME"
```
travis-minke-sap, line 34: Are these used anywhere?

pkg/channel/distributed/dispatcher/dispatcher/dispatcher.go
```diff
@@ -339,3 +348,39 @@ func (d *DispatcherImpl) ConfigChanged(ctx context.Context, configMap *corev1.Co
...
+		kafkaAuthCfg = nil
+	}
+	newConfig, err := client.NewConfigBuilder().WithExisting(d.SaramaConfig).WithAuth(kafkaAuthCfg).Build(ctx)
+	if err != nil {
+		d.Logger.Error("Unable to merge new auth into sarama settings", zap.Error(err))
+		return nil
+	}
+
+	// Create A New Producer With The New Configuration (Reusing All Other Existing Config)
+	d.Logger.Info("Changes Detected In New Secret - Closing & Recreating Dispatcher")
```
travis-minke-sap, line 384: Producer -> Dispatcher

pkg/channel/distributed/common/env/constants.go
travis-minke-sap: never mind - i see them now - searched for the wrong thing ; )

pkg/channel/distributed/dispatcher/dispatcher/dispatcher.go
```diff
@@ -329,6 +333,11 @@ func (d *DispatcherImpl) ConfigChanged(ctx context.Context, configMap *corev1.Co
 
 	// Create A New Dispatcher With The New Configuration (Reusing All Other Existing Config)
 	d.Logger.Info("Consumer Changes Detected In New Configuration - Recreating Dispatcher")
+	return d.reconfigure(newConfig)
+}
+
+// Shut down the current dispatcher and recreate it with new settings
+func (d *DispatcherImpl) reconfigure(newConfig *sarama.Config) Dispatcher {
 	d.Shutdown()
```
travis-minke-sap, line 341: Super nit picky but... can you move this function down below SecretChanged() - keeps the two public "changed" functions together and improves (imo) readability - thanks!

pkg/common/client/config.go
```diff
@@ -32,8 +33,9 @@ import (
 )
 
 type KafkaAuthConfig struct {
-	TLS  *KafkaTlsConfig
-	SASL *KafkaSaslConfig
+	Brokers string
+	TLS     *KafkaTlsConfig
```
travis-minke-sap, line 37: Pretty certain we don't want to do this.  Our long term direction is to put the brokers in the configmap instead of the secret (as the consolidated channel implementation does) and this is a temporary one-off for our own needs - let's try to find another workaround (or maybe just move the brokers if that's easier?)

travis-minke-sap:
> This is similar to #363. I think we can merge this PR that does changes only for distributed channel and then in #363 we can do the same for consolidated channel.

Thanks for being flexible - apologies for not being aware of #363 - sounds like a good plan!

pkg/channel/distributed/common/config/secretwatcher_test.go
eric-sap: Agreed; I put the two tests into a small table and now explicitly set the secret values.

cmd/channel/distributed/dispatcher/main.go
eric-sap: Changed to Dispatcher.  Also modified the internal messages to refer to closing and restarting Producer/Consumer Groups (the ones from the observer are still Receiver/Dispatcher).  I'm not completely certain those are the best places to use those terms but I attempted some more consistency.

pkg/channel/distributed/common/config/secretwatcher.go
eric-sap: Updated to 2021 for the current fileset.

pkg/channel/distributed/dispatcher/dispatcher/dispatcher.go
eric-sap: Adjusted

pkg/channel/distributed/dispatcher/dispatcher/dispatcher.go
eric-sap: absolutely

pkg/common/client/config.go
eric-sap: Since we're eventually planning to move the brokers out of the secret, I changed the GetAuthConfigFromKubernetes and GetAuthConfigFromSecret functions to return the broker string as part of the return tuple, but with a broker-free KafkaAuthConfig struct.  This should make it moderately easy to extricate and reposition later (though I would rather not do the move-to-configmap work as part of this particular issue).

PR #376, opened 2021-02-10
Allow v1beta1 stkafkasource topics & bootstrapservers to be mutable

lberk:
Change the validation of the kafkasource api to allow for changing the
topics/servers after creation

### DIFF
--- a/pkg/apis/sources/v1beta1/kafka_validation.go
+++ b/pkg/apis/sources/v1beta1/kafka_validation.go
@@ -19,6 +19,8 @@ package v1beta1
 import (
 	"context"
 
+	"github.com/google/go-cmp/cmp/cmpopts"
+
 	"knative.dev/pkg/apis"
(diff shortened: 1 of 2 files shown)


matzew:
/lgtm
I am fine for just doing this for `v1beta1`.

/hold


/assign @lionelvillard for another set of review

lionelvillard:
/lgtm

It'd be nice to have e2e tests stressing this new feature (a follow up PR is fine)
Can you add a release note? (even though there is no release notes, maybe time to change this bad behavior?)
The doc should be updated as well.


I let @lberk unhold it.

lberk:
Thanks @lionelvillard

I've got a follow-up for a couple other fields that'll (hopefully)
make our validation a bit more consistent with the norms of other
sources (instead of just blanket immutability for all the fields).

I'll add a release note, e2e tests, docs with the follow up PR so we can be
sure to encompass all the changes.

/unhold

PR #380, opened 2021-02-11
[mtsource] don't kill pods with placed vpods

lionelvillard:
## Proposed Changes

- Limit scaling down the adapter to the last pod with placed "vpods" 
- Temporarily disable cancelling single-tenant adapters. See https://github.com/knative-sandbox/eventing-kafka/issues/382
- 

**Release Note**

```release-note

```

**Docs**

source e2e tests are passing ... locally. Let's see if Prow is happy.

### DIFF
--- a/pkg/common/scheduler/statefulset/autoscaler.go
+++ b/pkg/common/scheduler/statefulset/autoscaler.go
@@ -46,7 +46,8 @@ func (s *StatefulSetScheduler) autoscale(ctx context.Context) {
 				zap.Int32("free", s.freeCapacity(snapshot)),
 				zap.Int32("used", s.usedCapacity(snapshot)),
 				zap.Int32("pending", s.pendingVReplicas()),
-				zap.Int32("replicas", s.replicas))
+				zap.Int32("replicas", s.replicas),
+				zap.Int32("last ordinal", snapshot.lastOrdinal))
 
 			var ratio float64
 			if s.replicas == 0 {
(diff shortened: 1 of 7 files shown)


matzew:
I still see a failure on the MT source job

lionelvillard:
I know :-) Not an easy fix. There are at least 2 other issues I'm currently working on.

pkg/common/scheduler/statefulset/snapshot.go
```diff
@@ -22,6 +22,10 @@ import "go.uber.org/zap"
 type Snapshot struct {
 	// free tracks the free capacity of each pod.
 	free map[string]int32
+
+	// lastOrdinal is the ordinal index corresponding to the last statefulset replica
+	// with placed vpods.
+	lastOrdinal int32
```
steven0711dong, line 28: Is the adding of lastOrdinal only for snapshot test?

pkg/common/scheduler/statefulset/autoscaler.go
```diff
@@ -72,13 +73,21 @@ func (s *StatefulSetScheduler) autoscale(ctx context.Context) {
...
-				s.logger.Infow("updating adapter replicas", zap.Int32("replicas", scale.Spec.Replicas))
+				// Make sure not to scale down past the last pod with placed vpods
+				if new < snapshot.lastOrdinal+1 {
+					new = snapshot.lastOrdinal + 1
+				}
 
-				_, err = s.statefulSetClient.UpdateScale(ctx, s.statefulSetName, scale, metav1.UpdateOptions{})
-				if err != nil {
-					s.logger.Errorw("updating scale subresource failed", zap.Error(err))
+				if new != scale.Spec.Replicas {
```
steven0711dong, line 83: Before this change, we always update the statefulset scale, even when new calculated replica number equals the current replicas. What's the drawback of this always update? Why is this update needed?

steven0711dong (approved):
/lgtm

pkg/common/scheduler/statefulset/snapshot.go
lionelvillard: It's is used below, to make sure no pods with placed vpods are killed.

pkg/common/scheduler/statefulset/autoscaler.go
lionelvillard: see comment above.

PR #410, opened 2021-02-19
Refactor distributed component mains to handle Kubernetes config better

aliok:
Refactoring distributed component `main`s to make them use more of the functionality in sharedmain.

I wasn't able to [debug](https://github.com/knative/eventing/blob/master/DEVELOPMENT.md#debugging-knative-controllers-and-friends-locally-with-intellij-idea) the receiver and dispatcher locally because they required me to provide additional stuff like Kube URL etc. Once I provided these, this time the debugging failed because of missing service account token.

Fixes #

## Proposed Changes

- Use `ParseAndGetRESTConfigOrDie` of `knative/pkg` which handles in-cluster or out-of-cluster situations better than `BuildConfigFromFlags` of the same package
- Change above also made these `main`s more similar to other Knative controllers with injection (more effort needed in this area)

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/cmd/channel/distributed/dispatcher/main.go
+++ b/cmd/channel/distributed/dispatcher/main.go
@@ -18,11 +18,11 @@ package main
 
 import (
 	"context"
-	"flag"
 	"strconv"
 	"strings"
 
 	"knative.dev/eventing/pkg/kncloudevents"
+	"knative.dev/pkg/injection"
 
 	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
@@ -40,6 +40,7 @@ import (
 	dispatcherhealth "knative.dev/eventing-kafka/pkg/channel/distributed/dispatcher/health"
 	"knative.dev/eventing-kafka/pkg/client/clientset/versioned"
 	"knative.dev/eventing-kafka/pkg/client/informers/externalversions"
+	injectionclient "knative.dev/pkg/client/injection/kube/client"
 	kncontroller "knative.dev/pkg/controller"
 	"knative.dev/pkg/logging"
 	eventingmetrics "knative.dev/pkg/metrics"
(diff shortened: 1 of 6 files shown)


cmd/channel/distributed/dispatcher/main.go
```diff
@@ -50,18 +50,32 @@ import (
...
 func main() {
 
-	// Parse The Flags For Local Development Usage
-	flag.Parse()
+	ctx := signals.NewContext()
+
+	// Create The K8S Configuration (In-Cluster By Default / Cmd Line Flags For Out-Of-Cluster Usage)
+	k8sConfig := injection.ParseAndGetRESTConfigOrDie()
+
+	// TODO: do we really need these? I moved them up
```
aliok, line 63: The QPS and burst: do we really need these? I moved them up since before the Kube client was created twice (once w/ QPS and burst, once w/o them) and now we create it once.

cmd/channel/distributed/dispatcher/main.go
```diff
@@ -141,18 +155,9 @@ func main() {
...
-	}
-
-	const numControllers = 1
-	config.QPS = numControllers * rest.DefaultQPS
-	config.Burst = numControllers * rest.DefaultBurst
-	kafkaClientSet := versioned.NewForConfigOrDie(config)
-	kubeClient := kubernetes.NewForConfigOrDie(config)
+	kafkaClient := kafkaclientset.NewForConfigOrDie(k8sConfig)
 
-	kafkaInformerFactory := externalversions.NewSharedInformerFactory(kafkaClientSet, environment.ResyncPeriod)
```
aliok, line 155: Recreation of the Kube clients is gone now. Only created once at the top.

cmd/channel/distributed/dispatcher/main.go
aliok: > The QPS and burst: do we really need these?

I mean, it is totally fine to have these :) My main intention here is to understand why :)

travis-minke-sap:
Hi Ali - Thanks for taking the time to improve this - sorry it caused you problems.  I think it was just this way because some of it predated the injected kubeconfig stuff and/or ignorance of it on our part, etc...   I think the changes look good (just did a quick pass) but I'll have @eric-sap take a closer look to be sure and then I can merge it.

cmd/channel/distributed/dispatcher/main.go
eric-sap: I'm not sure why these were put in place originally but since it's a constant setting of "1" times the default values of 5.0/10 for the QPS/Burst, and originally they were (re-)set after the BuildConfigFromFlags() function, I suspect it was more "ensuring that these values are the defaults" instead of anything more explicit.  I think setting the values to "0.0" and "0" with a comment that this forces the default values instead of loading them from the Kubernetes config would accomplish the same effect and possibly be less confusing.  Or maybe it would be more confusing :)

eric-sap: Actually now that I look at it the code that sets the defaults for those two settings is in the ParseAndGetRESTConfigOrDie() function, so if the intention is to force the defaults the code likely needs to be the way it currently is.

eric-sap: It looks like the code that translates "zero" into those defaults is in MainWithConfig->EnableInjectionOrDie, which is not called by this non-shared main.go function, so my suspicion is that this enforcing of the default values was taken from there and dropped here (meaning that this code should stay how it is, I think).

eric-sap:
This looks good; after looking into the QPS/Burst settings, I am convinced that the setting of defaults here is just because the non-shared main.go code doesn't call knative.dev/pkg/injection/injection.go::EnableInjectionOrDie, which would do the same thing.

eric-sap:
I have built and deployed this branch locally and it starts up and receives/dispatches messages

aliok:
thanks for the reviews!

PR #418, opened 2021-02-26
Stringify the sarama RecordHeader slices for easier log reading

eric-sap:
This will change the way the debug log output formats messages for the distributed dispatcher and receiver.  They used to be at the whim of zap.Any() which produced a lot of base64 strings for the headers and the message, which is obnoxious to decode when looking through a log and provided no benefit.

## Proposed Changes

- 🧽 Added helper functions to convert []sarama.RecordHeader and []*sarama.RecordHeader inputs into basic string maps so that the logs of these are cleaner to human eyes, and called these from the ProduceKafkaMessage and consumeMessage functions, respectively.

### DIFF
--- a/pkg/channel/distributed/common/util/util.go
+++ b/pkg/channel/distributed/common/util/util.go
@@ -20,8 +20,6 @@ import (
 	"os"
 	"os/signal"
 
-	"github.com/Shopify/sarama"
-
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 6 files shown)


pkg/channel/distributed/common/util/util.go
```diff
@@ -36,3 +38,23 @@ func WaitForSignal(logger *zap.Logger, signals ...os.Signal) {
...
+
+// Pointer-version of the StringifyHeaders function
+func StringifyHeaderPtrs(headers []*sarama.RecordHeader) map[string][]string {
+	stringHeaders := make(map[string][]string)
+	for _, header := range headers {
+		key := string(header.Key)
+		stringHeaders[key] = append(stringHeaders[key], string(header.Value))
+	}
+	return stringHeaders
+}
```
travis-minke-sap, line 60: My kingdom for an interface!

should these be in `distributed/common/kafka/sarama` instead?  or maybe in the `eventing-kafka/pkg/common` ?

eric-sap: kafka/sarama probably is a more relevant location.  We can move it out to the higher pkg/common if it becomes something we want in consolidated at that time.

aliok:
/lgtm
/approve

Thanks!

aliok (approved):

PR #428, opened 2021-03-03
Use Oldest offset as the initial one when creating the consumer group

### LINKED ISSUE #420: After kafka-ch-dispatcher is restarted, undelivered events in Kafkatopic are not delivered.
**Describe the bug**
After kafka-ch-dispatcher is restarted, undelivered events in Kafkatopic are not delivered.

 **Expected behavior**
After kafka-ch-dispatcherh is restarted, it will deliver the undelivered events in Kafkatopic.

 **To Reproduce**
- Create a Kafkachannel.
- Send an event to the created Kafka channel.
- Execute 'docker kill' command to stop the kafka-ch-dispatcher container before the event is delivered.The kafka-ch-dispatcher will be launched by the deployment resource.
- KafkaTopic still has undistributed events.However, kafka-ch-dispatcher will not deliver this event.

 **Knative release version**
0.20

slinkydeveloper:
Signed-off-by: Francesco Guardiani <francescoguard@gmail.com>

Fixes #420

## Proposed Changes

- Use OffsetOldest as default, but let user change this value through the config-kafka CM

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/channel/consolidated/dispatcher/dispatcher.go
+++ b/pkg/channel/consolidated/dispatcher/dispatcher.go
@@ -141,7 +141,11 @@ func (d *KafkaDispatcher) ServeHTTP(w nethttp.ResponseWriter, r *nethttp.Request
 }
 
 func NewDispatcher(ctx context.Context, args *KafkaDispatcherArgs) (*KafkaDispatcher, error) {
+	confTemplate := sarama.NewConfig()
+	confTemplate.Consumer.Offsets.Initial = sarama.OffsetOldest
+
 	conf, err := client.NewConfigBuilder().
+		WithExisting(confTemplate).
 		WithClientId(args.ClientID).
 		WithDefaults().
 		FromYaml(args.SaramaSettingsYamlString).


aliok:
- Should we also do it for distributed channel?
- Sorry, I don't have 100% knowledge on this: why not keeping the default as is but documenting that this behavior can be done with the CM change?

slinkydeveloper:
> Should we also do it for distributed channel?

I honestly don't know cc @travis-minke-sap 

> Sorry, I don't have 100% knowledge on this: why not keeping the default as is but documenting that this behavior can be done with the CM change?

Because the default behaviour doesn't actually reflect the user expectations, which is: if no offset is committed, the consumer group should start reading from the beginning of the partition. In other words, this should be the default behaviour.

travis-minke-sap:
> > Should we also do it for distributed channel?
> 
> I honestly don't know cc @travis-minke-sap

I think it would be good for the distributed and consolidated channels to support the same configuration, and I support the idea of making it a config setting.  I prefer that we not change the distributed implementation to default to the oldest offset UNTIL the configuration option has been added so that current users will have the option to maintain the current behavior.

> > Sorry, I don't have 100% knowledge on this: why not keeping the default as is but documenting that this behavior can be done with the CM change?
> 
> Because the default behaviour doesn't actually reflect the user expectations, which is: if no offset is committed, the consumer group should start reading from the beginning of the partition. In other words, this should be the default behaviour.

I'm curious as to why the oldest offset should be the default?  How do you know that is the user's assumption / expectation?  I'm not saying you're wrong - just that both assumptions seem equally valid.  Have we heard otherwise from users?  I can imagine high-volume use cases where historical data (from days/weeks ago) is not of any use to a subscriber and they just want to start receiving new events.

travis-minke-sap:
After reading associated Issue #420   ; )

This seems like a startup bootstrap edge case that we're just using the initial offset as an easy fix for?   Should the dispatcher maybe commit the previous offset (not sure if that's possible) upon startup to put a marker in place as to where it started in case the first event fails?

Again (regardless of whatever the default value might be) I still think making it configurable is a win if we have the time ; )

matzew (approved):
/lgtm
/approve

pkg/channel/consolidated/dispatcher/dispatcher.go
```diff
@@ -141,7 +141,11 @@ func (d *KafkaDispatcher) ServeHTTP(w nethttp.ResponseWriter, r *nethttp.Request
 }
 
 func NewDispatcher(ctx context.Context, args *KafkaDispatcherArgs) (*KafkaDispatcher, error) {
+	confTemplate := sarama.NewConfig()
+	confTemplate.Consumer.Offsets.Initial = sarama.OffsetOldest
```
matzew, line 145: what happens on upgrades (e.g. from 0.21 to 0.22), when this changes?

Does it have side effects.

I shot too quick w/ the LGTM

slinkydeveloper:
> This seems like a startup bootstrap edge case that we're just using the initial offset as an easy fix for? 

~TBH I don't think it's an easy fix, IMO is logically correct: the channel topic is a topic _we_ control and the dispatcher consumer group always starts from the beginning: when you start consuming there might be messages that are already pushed in there.~

Ok I'm thinking about this again. Because our consumer groups are per subscription, when a new subscription starts, using `OldestOffset` will start re-sending from the beginning all the messages in the channel... I completely neglected this... Is that at least once anyway right?

>  Should the dispatcher maybe commit the previous offset (not sure if that's possible) upon startup to put a marker in place as to where it started in case the first event fails?

I'm pretty sure you can't commit -1 as offset.

> Again (regardless of whatever the default value might be) I still think making it configurable is a win if we have the time ; )

It's still configurable this way, you can change it in `config-kafka`, mine is setting a default.

> we we're able to determine the offset from a timestamp - maybe could do something similar to determine the baseline offset to commit upon startup before processing events?

That sounds like a good idea

travis-minke-sap:
> Ok I'm thinking about this again. Because our consumer groups are per subscription, when a new subscription starts, using `OldestOffset` will start re-sending from the beginning all the messages in the channel... I completely neglected this... Is that at least once anyway right?

right - it would start every new subscription from the beginning of the channel (kafka topic/log)

> I'm pretty sure you can't commit -1 as offset.

Yeah, sorry - I didn't mean commit `-1` explicitly - I meant determine what the current/next offset would be and then subtract one from that to get the point in the log/topic prior to starting to send events...
 
> It's still configurable this way, you can change it in `config-kafka`, mine is setting a default.

Yeah, agreed - I only meant that we need to implement the configurability before we change the "default" (at least for distributed implementation) rather than change the default and maybe later get around to making it configurable ; ).  cart/horse vs horse/cart is all - haha
 
> > we we're able to determine the offset from a timestamp - maybe could do something similar to determine the baseline offset to commit upon startup before processing events?
> 
> That sounds like a good idea

👍

slinkydeveloper:
> Yeah, sorry - I didn't mean commit -1 explicitly - I meant determine what the current/next offset would be and then subtract one from that to get the point in the log/topic prior to starting to send events...

This would be -1 when the topic is new (like in the example of #420)

> rather than change the default and maybe later get around to making it configurable

This change doesn't affect distributed impl, right? Are you worried it might cause inconsistencies between the 2 impls?

pkg/channel/consolidated/dispatcher/dispatcher.go
slinkydeveloper: https://github.com/knative-sandbox/eventing-kafka/pull/428#issuecomment-789802302

travis-minke-sap:
> This would be -1 when the topic is new (like in the example of #420)

Yeah, I assume we'd be able to handle the 0 case to not subtract 1 - just an idea - same issue arises if we determine the current offset from the timestamp - it might be 0 as well.
 
> This change doesn't affect distributed impl, right? Are you worried it might cause inconsistencies between the 2 impls?

Correct - this PR is not affecting the distributed implementation and I (as a user of the distributed channel) am not affected by this change and am fine with it proceeding.  The comments are just around @aliok 's question of how to do this in a consistent way for both channels.  If there is urgency for making this change in the consolidated channel, and then later we want to add configurability to distributed/both and change the distributed default value, - thats ok with me ; ). The only downside (as you mention) is the difference between the two channels for that interim time period.

devguyio (changes requested):
I think this is not correct. @slinkydeveloper as you've said, this means that a subscription can end up receiving events that are days before its creation and that's not correct.

slinkydeveloper:
Ok then I'll close it and we can tell the user to use the config-kafka map to fix its corner case

Issue #426, opened 2021-03-03
Remove usage of eventing duck v1alpha1 shapes


vaikas:
**Describe the bug**
The v1alpha1 channelable shape has been deprecated forever and we need to remove it. 
https://github.com/knative/eventing/pull/5005

Looks like kafka still uses it and it's still storing channels as v1alpha1:
https://github.com/knative-sandbox/eventing-kafka/blob/master/config/channel/resources/kafkachannel-crd.yaml#L30

So, these will stop being reconciled once 5005 goes in. We should fix this asap so we can unblock the upstream work.

@matzew 

 **Expected behavior**
Using supported ducks for Channelable. 

 **To Reproduce**
Steps to reproduce the behavior.

 **Knative release version**
head

 **Additional context**
Add any other context about the problem here such as proposed priority

vaikas:
We will also need something like this to migrate storage versions from v1alpha1 to v1beta1:
https://github.com/knative/eventing/pull/3262

matzew:
@vaikas :see_no_evil:   that pre-job triggers bad memories

vaikas:
So here's couple of PRs. Note that you must start using the v1beta1 channelable like in this PR so that any channels created during the migration will get the proper duck type and they will get reconciled properly.

https://github.com/knative/eventing/pull/3169

Here's post install script:
https://github.com/knative/eventing/pull/3168/files

For Brokers, we moved from v1alpha1 -> v1beta1 by installing this CRD first:
https://github.com/knative/eventing/pull/3282

So you could do something like that for kafka, apply first that flips the bits while the conversion from v1alpha1<->v1beta1 still works.

Then run the pre-install storage migrator, then upgrade.
https://github.com/knative/eventing/pull/3110

slinkydeveloper:
Thanks for the pointers. I'm going to tackle it now.

Since this issue is a duplicated of #71, I'm going to close this

vaikas:
I do not think there's a way to do this smoothly without preinstall or adding some special handling for the kafka channel for the subscription controller. The problem is that the Subscription reconciler will treat the Channel differently (because the shapes are different) based on the label on the Channel. After #5005 in eventing went in, it will not have any understanding of the old channel types so it will not see the subscribers and it will produce wrong patches that will fail.
[...]

PR #430, opened 2021-03-03
move eventing/pkg/apis/duck/v1alpha1 here

vaikas:
Addresses #426 

## Proposed Changes

- This moves the ducks from knative.dev/eventing/pkg/apis/duck/v1alpha1 here

**Release Note**

```release-note
 * Note that after the upstream PR: https://github.com/knative/eventing/pull/5005 goes in, old v1alpha1 channels will not be reconciled.

```

**Docs**


vaikas:
So, as expected the v1alpha1 tests are now failing.

pierDipi:
I can't think of any other solution.

I guess during a "pre-upgrade step" we have to:
- change the stored version to v1beta1 
- stop serving v1alpha1

_and there no way to downgrade_

Such pre-upgrade has to happen before upgrading eventing.

/cc @matzew @travis-minke-sap

pierDipi:
In this PR we should remove v1alpha1.

vaikas:
@pierDipi would you like me to make the PR that removes the v1alpha1 as part of this PR? If that's the case, upgrade will not be easily done (not even sure how) since the storage upgrade scripts reads them in at v1alpha1 and converts. I could stop serving v1alpha1 if that's what you'd like. But I think I personally would split the PRs into this, then the next and upgrade separate ones for smaller chunks. Thoughts?

pierDipi:
If you can make tests work in smaller chunks, I prefer smaller chunks, but at this point I guess v1alpha1 with the latest eventing just doesn't work, that's the reason I was suggesting to remove it, or am I wrong?

For the upgrade part, I'm thinking to a multi step pre-upgrade with multiple jobs, or do you a better solution to give users an upgrade path?

vaikas:
@pierDipi ok, I disabled tests with v1alpha1 and everything passes. So, I think we should get this in, and start working on the upgrade path, thoughts?

travis-minke-sap (approved):
looks like a good stepping stone ; )

PR #445, opened 2021-03-09
[source] Set default number of consumers

lionelvillard:
## Proposed Changes

- Set the default number of consumers to 1. A default value is needed for KEDA.
-
-

**Release Note**

```release-note
- 🐛The default number of KafkaSource consumers is now set to 1. 
```

**Docs**


matzew (approved):
/lgtm
/approve

PR #446, opened 2021-03-10
Upgraded storage version of kafkachannel

slinkydeveloper:
Signed-off-by: Francesco Guardiani <francescoguard@gmail.com>

Fixes #73

## Proposed Changes

- 🧽 Upgraded storage version of kafka channel

**Release Note**

```release-note
🧽 Update or clean up current behavior
```

**Docs**

PR #448, opened 2021-03-10
Added post-install script to migrate KafkaChannel to v1beta1

slinkydeveloper:
Signed-off-by: Francesco Guardiani <francescoguard@gmail.com>

Followup of #446

## Proposed Changes

- 🧽 Added a post-install migration job to migrate KafkaChannel to v1beta1

**Release Note**

```release-note
🧽 Added a post-install migration job to migrate KafkaChannel to v1beta1. You must run it after installing eventing-kafka 0.22
```

**Docs**

### DIFF
--- a/config/post-install/clusterrole.yaml
+++ b/config/post-install/clusterrole.yaml
@@ -1,0 +1,46 @@
+# Copyright 2021 The Knative Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: knative-kafka-eventing-post-install-job-role
+  labels:
+    eventing.knative.dev/release: devel
+rules:
+  # Storage version upgrader needs to be able to patch CRDs.
+  - apiGroups:
+      - "apiextensions.k8s.io"
+    resources:
+      - "customresourcedefinitions"
+      - "customresourcedefinitions/status"
+    verbs:
+      - "get"
+      - "list"
+      - "update"
+      - "patch"
+      - "watch"
+  # Our own resources we care about.
+  - apiGroups:
+      - "messaging.knative.dev"
+    resources:
+      - "kafkachannels"
+    verbs:
+      - "get"
+      - "list"
+      - "create"
+      - "update"
+      - "delete"
+      - "patch"
+      - "watch"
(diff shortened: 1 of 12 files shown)


aliok:
Can you rebase this once https://github.com/knative-sandbox/eventing-kafka/pull/446 is merged?

slinkydeveloper:
@aliok this is ready for review

aliok (dismissed):
/lgtm
/approve

Looks good!

devguyio (dismissed):
I believe this must be a pre-install job. v0.22 must end up with a KafkaChannel CRD that says
```yaml
- name: v1alpha1
  served: false
  storage: false
- name: v1beta1 
  served: true
  storage: true
```

config/post-install/clusterrole.yaml
```diff
@@ -0,0 +1,46 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: knative-kafka-eventing-post-install-job-role
+  labels:
+    eventing.knative.dev/release: devel
```
matzew, line 20: same regarding the label value

config/post-install/serviceaccount.yaml
```diff
@@ -0,0 +1,38 @@
...
+    eventing.knative.dev/release: devel
+
+---
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: knative-kafka-eventing-post-install-job-role-binding
+  labels:
+    eventing.knative.dev/release: devel
```
matzew, line 30: same... you get it @slinkydeveloper  :smile:

devguyio:
I was hoping we'd not end up in a state where we are the only object in `messaging.knative.dev` that's still serving v1alpha1. But if we want to play it 100% safe we should have a release that has storage on v1beta, and still serves v1alpha1 in order to allow for safe storage migration in case the user forgot to run the migration job. TL;DR that's how I understand our options:
- Option1: Have storage and serving migration as pre-install
    - Pros:
        - User ends up with no v1alpha1 objects in messaging.knative.dev 
        - Faster cut / getting rid of dangling v1alpha1 work
    - Cons:
        - Upgrading users might run into issue if they forgot to run the pre-install job
- Option2: Have storage migration as post-install , and stop serving in 0.23
    - Pros:
        - Upgrading users have a way to fix things in case they forgot to run the migration job
    - Cons:
        - We will end up being the only channel in the sandbox that still serves v1alpha1 
        - We still need to plan the removal work if v1alpha1 in following versions

I feel maybe we should cater for upgrading user safety and go for Option 2 indeed.  Let's un-hold once we get an [answer here](https://github.com/knative-sandbox/eventing-kafka/issues/426#issuecomment-795162530)

matzew:
A real problem happens if both storage/serve go from true/true to false/false in just one release.

Context for one of the most painful things I noticed with this so far: 
https://github.com/knative/eventing/issues/4455

matzew:
> We will end up being the only channel in the sandbox that still serves v1alpha1

To me that is not a contra point. It is what it is. Better than some `pre-job` nightmare. AND the removal work of the _duck v1alpha1 shapes_ is needed anyways, right  ? 


Sounds like option 2 is the way to go. for 0.22 and on .23 we do the switch to serve: false

slinkydeveloper:
> Option2: Have storage migration as post-install , and stop serving in 0.23 

ATM I'm pursuing this solution :smile:

aliok (approved):
/lgtm
/approve
/unhold

The discussion is here: https://github.com/knative-sandbox/eventing-kafka/issues/426

PR #449, opened 2021-03-10
Move setting CE connection args setting one level above in consolidated dispatcher

aliok:
Fixes #

## Proposed Changes

- Move setting CE connection args setting one level above in consolidated dispatcher so that it is more clear that this is to be done exactly once per process

**Release Note**

```release-note

```

**Docs**


slinkydeveloper (approved):
/lgtm

PR #433, opened 2021-03-05
Add Subscription prober

devguyio:
## Proposed Changes

- Adds a prober for the dispatcher subscriptions status

### TODO
- [x] Proper manual testing
- [x] Lots of cleanup
- [ ] Adding tests
- [ ] Add hash support

PR #451, opened 2021-03-10
Separate consolidated dispatcher subscription endpoint

aliok:
Fixes #

## Proposed Changes

- Separate consolidated dispatcher subscription endpoint into its own type
-
-

**Release Note**

```release-note

```

**Docs**


lberk:
/hold
/lgtm
as we chatted, holding to make sure #433 lands smoothly before adding this.

pkg/channel/consolidated/dispatcher/subscription_endpoint.go
```diff
@@ -0,0 +1,58 @@
...
+		return
+	}
+	_, err = w.Write(jsonResult)
+	if err != nil {
+		d.logger.Errorf("Error writing jsonResult to serveHTTP writer: %w", err)
+	}
+}
+
+func (d *subscriptionEndpoint) start() {
+	d.logger.Fatal(nethttp.ListenAndServe(":8081", d))
```
slinkydeveloper, line 57: I think this goes fatal if err == nil, so split it with a if err != nil

aliok: From https://golang.org/pkg/net/http/#ListenAndServe

```
ListenAndServe always returns a non-nil error.
```

aliok:
@slinkydeveloper this is rebased.

I guess it is fine if don't ask for others' review since this is only a refactoring w/o any business logic change.

pkg/channel/consolidated/dispatcher/subscription_endpoint.go
aliok: :shrug:

slinkydeveloper (approved):

PR #456, opened 2021-03-11
reduce apiserver requests for dispatcher

xinydev:
Signed-off-by: XinYang <xinydev@gmail.com>

The dispatcher now only gets information for the channel it cares about. reducing the GET requests to apiserver in a cluster with lots of channels.

## Proposed Changes

-  return before GET requests for channels dispatcher doesn’t need to care about

**Release Note**

```release-note

```

**Docs**


travis-minke-sap:
looks like a good change (thanks!) - I'm doing a quick smoke test now and then will approve!

travis-minke-sap (approved):
Thanks for this fix - nice efficiency win !

PR #464, opened 2021-03-15
Support arbitrary sarama metrics

### LINKED ISSUE #222: Expose Sarama metrics through Prometheus
**Problem**
Sarama metrics are not exposed.

**Exit Criteria**

**Time Estimate (optional):**

Metrics are documented there: https://godoc.org/github.com/Shopify/sarama

eric-sap:
Fixes #222

**Proposed Changes**

- Moved stats_reporter.go to common/metrics and updated distributed dispatcher/receiver main functions to use it (note: did not alter consolidated channel to use the new common functionality yet but this should make it easier if desired)
- Added capability for the StatsReporter.Report() function to record all incoming Sarama metrics instead of just a single message count.
- If the old single-metric is desired, this an example of that count value in the new collection:
```
HELP eventing_kafka_record_send_rate_for_topic_stage_sample_kafka_channel_1_count Record Send Rate for Topic stage_sample-kafka-channel-1: Count
TYPE eventing_kafka_record_send_rate_for_topic_stage_sample_kafka_channel_1_count gauge
eventing_kafka_record_send_rate_for_topic_stage_sample_kafka_channel_1_count 12345
```

**Release Note**

```
Distributed KafkaChannel now exposes all Sarama metrics, previous single count metric has been renamed.
```


eric-sap:
This is a duplicate of PR #460, closed due to complications with it being opened by "not-the-owner-of-the-repo "

pkg/common/metrics/stats_reporter.go
```diff
@@ -0,0 +1,225 @@
...
+
+	newView := &view.View{
+		Name:        name,
+		Description: description,
+		Measure:     measure,
+		Aggregation: view.LastValue(),
+	}
+	err = view.Register(newView)
+	if err != nil {
+		r.logger.Error("failed to register opencensus views", zap.Error(err))
```
matzew, line 133: nit: `Failed...` ?

aliok: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pkg/channel/distributed/common/metrics/stats_reporter.go
```diff
@@ -28,69 +27,39 @@ import (
...
+		measurements: make(map[string]stats.Measure)}
 }
 
 // Our RecordWrapper, which defaults to the knative metrics.Record()
 // This wrapper function facilitates minimally-invasive unit testing of the
 // Report functionality without requiring live servers to be started.
 var RecordWrapper = metrics.Record
 
+type ReportingItem = map[string]interface{}
+type ReportingList = map[string]ReportingItem
```
aliok: Thank you for these. Map of maps was not very easy to read

aliok (dismissed):
/lgtm
/approve
/hold

Leaving the final approve to @travis-minke-sap as he has more context

pkg/common/metrics/stats_reporter.go
eric-sap: I've seen in this repo that error structs created via mechanisms like errors.New() and fmt.Errorf follow the style guideline of "no initial capital, no trailing punctuation" but not necessarily the direct logger output (fmt.Print, zap.Info, etc).  I don't know if there's a specific style guide for the zap output -- in this particular case the error was copied from eventing code (knative.dev/eventing/pkg/channel/stats_reporter.go) and I didn't think to change it to match the file style.  I would be happy to make things at least consistent (or better, consistently correct) if there is a recommended style.

aliok: IMO, according to that code conventions page I posted above, your code is right

aliok (dismissed):
/lgtm
/approve

travis-minke-sap:
Did you verify that the Dispatcher is producing sarama metrics?  I don't see where that was enabled here?

eric-sap:
After some experimentation with my original code, I discovered that the individually-exported metrics were very unwieldy to use.  For example, having the 75% and 95% (and other percentile) values listed as completely distinct metrics with their own names and descriptions led to it being very cumbersome to make Prometheus queries that combined them for easy viewing.

As such, I have reworked the code to make those percentile metrics into single-TimeSeries-arrays (i.e. one name with multiple tags for the different percentile sections).  Examples are given in the comments of the recordMetric() function of stats_reporter.go.  This required creating our own Producer as the OpenCensus "view" paradigm wouldn't allow anything other than cumulative histograms, which aren't what Sarama gives us.  Fortunately that was not overly complicated, requiring only a Read() function and local storage of the latest snapshot of Sarama metrics in a map so that the Producer interface has something to return when Read is called.

I also considered putting the meters into TimeSeries arrays the same way, but that wasn't as much of an improvement (e.g. the single metric of "Responses/second received from broker 0: 5-Minute Rate" is actually useful by itself without needing to be coupled with the 15-minute and 1-minute rates.   However, it wouldn't be difficult to add "isMeterMetric" and corresponding "recordMeterMetric" functions if we want that behavior now or in the future.

@aliok and @travis-minke-sap if you wouldn't mind taking another look at this refactor, I think it will be a good bit more useful and more worth of being in the common code.

travis-minke-sap:
Looks ok to me - but I'm no opencensus expert ; )

Will remove hold and approve this tomorrow if no further comments from @aliok or @matzew (or others)

aliok (approved):
/lgtm
/approve

Thanks!

PR #470, opened 2021-03-16
[mtsource] Automatically adjust Sarama resource usage to fit within limits

lionelvillard:
## Proposed Changes

- Compute the maximum fetch buffer size to not exceed the configured memory limit. Take into account the number of partitions handled by the sources and inevitable pod outages. 
-  Set default pod memory request to 100MB, capacity of 50, so each vreplica memory limit is 2MB. With these parameters, the maximum number of partitions a source can handle to process 64k messages is about 16.  
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/common/scheduler/placement.go
+++ b/pkg/common/scheduler/placement.go
@@ -38,3 +38,18 @@ func GetPlacementForPod(placements []duckv1alpha1.Placement, podName string) *du
 	}
 	return nil
 }
+
+// GetPodCount returns the number of pods with the given placements
+func GetPodCount(placements []duckv1alpha1.Placement) int {
+	set := map[string]int{}
+	for _, p := range placements {
+		if p.VReplicas > 0 {
+			if _, ok := set[p.PodName]; !ok {
+				set[p.PodName] = 1
+				continue
+			}
+			set[p.PodName] += 1
+		}
+	}
+	return len(set)
+}
--- a/pkg/common/scheduler/placement.go
+++ b/pkg/common/scheduler/placement.go
@@ -17,6 +17,7 @@ limitations under the License.
 package scheduler
 
 import (
+	"k8s.io/apimachinery/pkg/util/sets"
 	duckv1alpha1 "knative.dev/eventing-kafka/pkg/apis/duck/v1alpha1"
 )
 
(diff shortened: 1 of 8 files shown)


pkg/source/client/client.go
```diff
@@ -119,3 +119,12 @@ func NewProducer(ctx context.Context) (sarama.Client, error) {
 
 	return sarama.NewClient(bs, cfg)
 }
+
+// NewProducer is a helper method for constructing an admin client
+func MakeAdminClient(ctx context.Context, env *KafkaEnvConfig) (sarama.ClusterAdmin, error) {
```
matzew, line 124: why not aligning with here: https://github.com/knative-sandbox/eventing-kafka/blob/ee7edfdd50d750db3b984c17f0d5cf835dcc3910/pkg/common/client/admin_client.go ?

lionelvillard: because the heavy-lifting is done in NewConfigWithEnv.

pkg/source/mtadapter/adapter.go
```diff
@@ -87,55 +95,127 @@ func (a *Adapter) Start(ctx context.Context) error {
...
 
+// bufferSize returns the maximum fetch buffer size (in bytes)
+// so that the st adapter memory consumption does not exceed
+// the allocated memory per vreplica (see MemoryLimit).
+// Account pod (consumer) partial outage.
+func (a *Adapter) bufferSize(ctx context.Context,
+	logger *zap.SugaredLogger,
+	kafkaEnvConfig *client.KafkaEnvConfig,
+	topics []string,
+	consumerCount int) (int, error) {
```
steven0711dong, line 106: should really be podCount?

pkg/common/scheduler/placement.go
```diff
@@ -38,3 +38,18 @@ func GetPlacementForPod(placements []duckv1alpha1.Placement, podName string) *du
...
+// GetPodCount returns the number of pods with the given placements
+func GetPodCount(placements []duckv1alpha1.Placement) int {
+	set := map[string]int{}
+	for _, p := range placements {
+		if p.VReplicas > 0 {
+			if _, ok := set[p.PodName]; !ok {
+				set[p.PodName] = 1
+				continue
+			}
+			set[p.PodName] += 1
```
steven0711dong, line 51: Since we don't really use the map value here, no need to increment the count. Consider using a set instead of a map map[string]interface{}

steven0711dong: Actually I can improve this after we merge this PR.

pkg/source/mtadapter/adapter.go
lionelvillard: yes good catch, thanks!

pkg/common/scheduler/placement.go
lionelvillard: I did it. Thx

steven0711dong (dismissed):
LGTM

steven0711dong (approved):
LGTM

PR #471, opened 2021-03-17
Do not mark messages as consumed if the context was canceled

eric-sap:
## Proposed Changes

- 🐛 During blue/green upgrade testing of the distributed channel, we found that when a continuous stream of messages is being sent during a rollout restart, the dispatcher would often miss a single message when the new pod became ready.  This was tracked down to MarkMessage being called when consumeMessage() returns a "context was canceled" error, which is not appropriate.  This change skips the MarkMessage call when that particular error is returned, and adds a new unit test to verify that.

### DIFF
--- a/pkg/channel/distributed/dispatcher/dispatcher/handler_test.go
+++ b/pkg/channel/distributed/dispatcher/dispatcher/handler_test.go
@@ -93,7 +93,9 @@ func TestHandlerConsumeClaim(t *testing.T) {
 		destinationUri *apis.URL
 		replyUri       *apis.URL
 		deadLetterUri  *apis.URL
+		dispatchErr    error
 		retry          bool
+		expectSkipMark bool
 	}
 
 	// Define The TestCases
@@ -134,6 +136,15 @@ func TestHandlerConsumeClaim(t *testing.T) {
 			name:  "Empty Subscriber Configuration",
 			retry: false,
 		},
+		{
+			name:           "Context Canceled",
+			destinationUri: testSubscriberURI,
+			replyUri:       testReplyURI,
+			deadLetterUri:  testDeadLetterURI,
+			retry:          true,
+			dispatchErr:    context.Canceled,
+			expectSkipMark: true,
+		},
 	}
 
 	// Filter To Those With "only" Flag (If Any Specified)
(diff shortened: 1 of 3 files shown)


pkg/channel/distributed/dispatcher/dispatcher/handler_test.go
```diff
@@ -150,13 +161,26 @@ func TestHandlerConsumeClaim(t *testing.T) {
...
-			performHandlerConsumeClaimTest(t, testCase.destinationUri, testCase.replyUri, testCase.deadLetterUri, testCase.retry)
+			performHandlerConsumeClaimTest(t,
+				testCase.destinationUri,
+				testCase.replyUri,
+				testCase.deadLetterUri,
+				testCase.retry,
+				testCase.dispatchErr,
+				testCase.expectSkipMark,
+			)
 		})
```
travis-minke-sap, line 172: maybe it's time we just pass the testCase to the performHandlerConsumeClaimTest() function ?

pkg/channel/distributed/dispatcher/dispatcher/handler_test.go
```diff
@@ -203,26 +227,34 @@ func performHandlerConsumeClaimTest(t *testing.T, destinationUri, replyUri, dead
...
+		assert.NotNil(t, mockMessageDispatcher.Message())
+		verifyDispatchedMessage(t, mockMessageDispatcher.Message())
+	}
+	assert.Equal(t, !expectSkipMark, mockConsumerGroupSession.MarkMessageCalled())
 
 	// Close The Mock ConsumerGroupClaim Message Channel To Complete/Exit Handler's ConsumeClaim()
 	close(mockConsumerGroupClaim.MessageChan)
 
-	// Verify The Results (CloudEvent Was Dispatched & ConsumerMessage Was Marked)
-	assert.Equal(t, consumerMessage, markedMessage)
```
travis-minke-sap, line 223: can you put this `// Verify The Results` comment up above new section line 242?

travis-minke-sap (commented):
looks good - couple minor suggestions which are up to you...

pkg/channel/distributed/dispatcher/dispatcher/handler_test.go
eric-sap: sounds good

pkg/channel/distributed/dispatcher/dispatcher/handler_test.go
eric-sap: et voilà

travis-minke-sap (approved):
/lgtm

PR #486, opened 2021-03-29
Global resync Kafka channels when config-kafka is changed

aliok:
Fixes #

## Proposed Changes

- Global resync Kafka channels when config-kafka is changed. So, everything will be resynced, including namespaced channels. Channel controller doesn't do anything that's not repeatable per process, so, this is fine.
- Dispatcher can't work the same though. My idea is to kill the dispatcher when there's a change in the configmap so a new process starts

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/channel/consolidated/reconciler/controller/controller.go
+++ b/pkg/channel/consolidated/reconciler/controller/controller.go
@@ -107,8 +107,21 @@ func NewController(
 	)
 	r.statusManager = statusProber
 	statusProber.Start(ctx.Done())
+
+	// Call GlobalResync on kafkachannels.
+	grCh := func(obj interface{}) {
+		logger.Info("Changes detected, doing global resync")
+		impl.GlobalResync(kafkaChannelInformer.Informer())
+	}
+
+	handleKafkaConfigMapChange := func(ctx context.Context, configMap *corev1.ConfigMap) {
+		logger.Info("Configmap is updated or, it is being read for the first time")
+		r.updateKafkaConfig(ctx, configMap)
+		grCh(configMap)
+	}
+
 	// Get and Watch the Kakfa config map and dynamically update Kafka configuration.
-	err := commonconfig.InitializeKafkaConfigMapWatcher(ctx, cmw, logger, r.updateKafkaConfig, system.Namespace())
+	err := commonconfig.InitializeKafkaConfigMapWatcher(ctx, cmw, logger, handleKafkaConfigMapChange, system.Namespace())
 	if err != nil {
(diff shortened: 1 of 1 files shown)


travis-minke-sap:
I'm confused... Is this adding the dynamic watching of the configmap to consolidated channel?  I thought that was not workable and we were going to remove that capability from the distributed channel?

aliok:
@travis-minke-sap

The problem with on-the-fly reloading is around the dispatcher/receiver and not on the controller.

My current thinking is:
- User changes config-kafka
- Controller picks it up, it does a global resync
- At this stage, we also need to reload the dispatchers/receivers. Either by getting the controller to recreate them entirely, or by watching the configmap directly committing suicide inside distributers/receivers.

What I would like to **disable** is **reloading** config-kafka **without restarting** the dispatcher/receiver pod.

1. That doesn't work properly because there are some parameters in config-kafka that can only be set once in a process
2. Just killing the entire process and starting from scratch makes the code much simpler

slinkydeveloper (commented):
Sounds like a good approach

travis-minke-sap:
ok, thanks for clarifying - makes sense ; )

aliok:
I am just thinking how can I test this change properly at the moment.

aliok:
I can't add automated e2e tests for this. The old behavior didn't have a test either.

I wrote some YAML + some tools here: https://github.com/aliok/kafka-ch-reload-tests

I did some manual tests and I can see everything works fine when `config-kafka` changes during message ingestion by KafkaChannel.

Reload logic in KafkaChannel controller is only affecting the controller itself anyway. It doesn't do anything about dispatcher. (it was this way and I want to keep it that way).

aliok:
cc @eric-sap mind having a look?

eric-sap:
The moving of the GlobalResync into a function that is called when the configmap changes seems like an overall useful change.  Just to confirm - all that does is force a re-reconcile of the KafkaChannels, and has nothing to do with any restarting of deployments/pods (because the consolidated KafkaChannel reconciler doesn't do that unless the image changes or the replicas were somehow set to zero).  My knowledge of the consolidated behavior is somewhat limited, though, and there may be behavior I've missed.

aliok:
@eric-sap 

> Just to confirm - all that does is force a re-reconcile of the KafkaChannels, and has nothing to do with any restarting of deployments/pods

Yes, that's true. 

I am working on another PR that does the restart of the dispatcher pods.

aliok:
/hold

Let's wait for the release cut for getting this in.

eric-sap:
/lgtm
(though I don't believe I'm in the group with permissions to use that command)

matzew (approved):
/lgtm
/approve

PR #487, opened 2021-03-29
[Consolidated Channel] Update target pod IPs when dispatcher pods scale before the subscription is marked ready 

devguyio:
## Background
The existing dispatcher pods Prober acts like this:
1. `IsReady(sub)` -> get list of target pods, cache the state (including the list of pods), and start probing those pods.
3. Probing collects ready partitions and stores them in the cached target state
4. After each dispatcher pod partitions are cached, check if we collected all partitions we are looking for and if so, mark the cached targetState as ready.  

There's a corner case the manifests itself as follows:
1. `IsReady(sub)` -> get list of target pods, cache the state (including the list of pods), and start probing those pods.
2.  Dispatcher pods scale up before the cached targetState becomes ready. 
3.  No new probes are created for the new pods.
4. The existing probing returns only a subset of the partitions, because not all dispatcher pods are probed for this targetState.

This is a fix for the behavior [described here](https://github.com/knative-sandbox/eventing-kafka/pull/476#issuecomment-809156584)

### Kafka Controller 
1. Refresh prober pods when dispatcher pods are added or deleted.

### Prober
1. Refactor the code to be more readible by breaking down `IsReady` and getting rid of non-readable anonymous function and logic.
2. Consider an in-flight targetState outdated if the list of initial probed pods change from the current, forget the state and reprobe new pods.
3. Add `RefreshPodProbing()` that will eject outdated in-flight target states, and reprobe the new pods instead.

### DIFF
--- a/pkg/channel/consolidated/reconciler/controller/controller.go
+++ b/pkg/channel/consolidated/reconciler/controller/controller.go
@@ -19,16 +19,15 @@ package controller
 import (
 	"context"
 
-	corev1 "k8s.io/api/core/v1"
-	knativeReconciler "knative.dev/pkg/reconciler"
-
 	"github.com/kelseyhightower/envconfig"
 	"go.uber.org/zap"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/cache"
 
 	"knative.dev/eventing-kafka/pkg/apis/messaging/v1beta1"
 	"knative.dev/eventing-kafka/pkg/channel/consolidated/status"
+	kafkavmessaging1beta1 "knative.dev/eventing-kafka/pkg/client/informers/externalversions/messaging/v1beta1"
 	kafkaChannelClient "knative.dev/eventing-kafka/pkg/client/injection/client"
 	"knative.dev/eventing-kafka/pkg/client/injection/informers/messaging/v1beta1/kafkachannel"
 	kafkaChannelReconciler "knative.dev/eventing-kafka/pkg/client/injection/reconciler/messaging/v1beta1/kafkachannel"
@@ -45,6 +44,7 @@ import (
 	"knative.dev/pkg/configmap"
 	"knative.dev/pkg/controller"
 	"knative.dev/pkg/logging"
+	knativeReconciler "knative.dev/pkg/reconciler"
 	"knative.dev/pkg/system"
 )
 
@@ -154,26 +154,22 @@ func NewController(
 		),
 		Handler: cache.ResourceEventHandlerFuncs{
 			// Cancel probing when a Pod is deleted
-			DeleteFunc: func(obj interface{}) {
-				pod, ok := obj.(*corev1.Pod)
-				if ok && pod != nil {
-					logger.Debugw("Dispatcher pod deleted. Refreshing pod probing.",
-						zap.String("pod", pod.GetName()))
-					statusProber.RefreshPodProbing(ctx)
-					impl.GlobalResync(kafkaChannelInformer.Informer())
-				}
-			},
-			AddFunc: func(obj interface{}) {
-				pod, ok := obj.(*corev1.Pod)
-				if ok && pod != nil {
-					logger.Debugw("Dispatcher pod added. Refreshing pod probing.",
-						zap.String("pod", pod.GetName()))
-					statusProber.RefreshPodProbing(ctx)
-					impl.GlobalResync(kafkaChannelInformer.Informer())
-				}
-			},
+			DeleteFunc: getPodInformerEventHandler(ctx, logger, statusProber, impl, kafkaChannelInformer, "Delete"),
+			AddFunc:    getPodInformerEventHandler(ctx, logger, statusProber, impl, kafkaChannelInformer, "Add"),
 		},
 	})
 
 	return impl
+}
+
+func getPodInformerEventHandler(ctx context.Context, logger *zap.SugaredLogger, statusProber *status.Prober, impl *controller.Impl, kafkaChannelInformer kafkavmessaging1beta1.KafkaChannelInformer, handlerType string) func(obj interface{}) {
+	return func(obj interface{}) {
+		pod, ok := obj.(*corev1.Pod)
+		if ok && pod != nil {
+			logger.Debugw("%s rods. Refreshing pod probing.", handlerType,
+				zap.String("pod", pod.GetName()))
+			statusProber.RefreshPodProbing(ctx)
+			impl.GlobalResync(kafkaChannelInformer.Informer())
+		}
+	}
 }
--- a/pkg/channel/consolidated/reconciler/controller/controller.go
+++ b/pkg/channel/consolidated/reconciler/controller/controller.go
@@ -157,9 +157,18 @@ func NewController(
 			DeleteFunc: func(obj interface{}) {
 				pod, ok := obj.(*corev1.Pod)
 				if ok && pod != nil {
-					logger.Debugw("Dispatcher pod deleted. Canceling pod probing.",
+					logger.Debugw("Dispatcher pod deleted. Refreshing pod probing.",
 						zap.String("pod", pod.GetName()))
-					statusProber.CancelPodProbing(*pod)
+					statusProber.RefreshPodProbing(ctx)
+					impl.GlobalResync(kafkaChannelInformer.Informer())
+				}
+			},
+			AddFunc: func(obj interface{}) {
+				pod, ok := obj.(*corev1.Pod)
+				if ok && pod != nil {
+					logger.Debugw("Dispatcher pod added. Refreshing pod probing.",
+						zap.String("pod", pod.GetName()))
+					statusProber.RefreshPodProbing(ctx)
 					impl.GlobalResync(kafkaChannelInformer.Informer())
 				}
 			},
--- a/pkg/channel/consolidated/reconciler/controller/controller.go
+++ b/pkg/channel/consolidated/reconciler/controller/controller.go
@@ -27,7 +27,7 @@ import (
 
 	"knative.dev/eventing-kafka/pkg/apis/messaging/v1beta1"
 	"knative.dev/eventing-kafka/pkg/channel/consolidated/status"
-	kafkavmessaging1beta1 "knative.dev/eventing-kafka/pkg/client/informers/externalversions/messaging/v1beta1"
+	kafkamessagingv1beta1 "knative.dev/eventing-kafka/pkg/client/informers/externalversions/messaging/v1beta1"
 	kafkaChannelClient "knative.dev/eventing-kafka/pkg/client/injection/client"
 	"knative.dev/eventing-kafka/pkg/client/injection/informers/messaging/v1beta1/kafkachannel"
 	kafkaChannelReconciler "knative.dev/eventing-kafka/pkg/client/injection/reconciler/messaging/v1beta1/kafkachannel"
@@ -162,11 +162,11 @@ func NewController(
 	return impl
 }
 
-func getPodInformerEventHandler(ctx context.Context, logger *zap.SugaredLogger, statusProber *status.Prober, impl *controller.Impl, kafkaChannelInformer kafkavmessaging1beta1.KafkaChannelInformer, handlerType string) func(obj interface{}) {
+func getPodInformerEventHandler(ctx context.Context, logger *zap.SugaredLogger, statusProber *status.Prober, impl *controller.Impl, kafkaChannelInformer kafkamessagingv1beta1.KafkaChannelInformer, handlerType string) func(obj interface{}) {
 	return func(obj interface{}) {
 		pod, ok := obj.(*corev1.Pod)
 		if ok && pod != nil {
-			logger.Debugw("%s rods. Refreshing pod probing.", handlerType,
+			logger.Debugw("%s pods. Refreshing pod probing.", handlerType,
 				zap.String("pod", pod.GetName()))
 			statusProber.RefreshPodProbing(ctx)
 			impl.GlobalResync(kafkaChannelInformer.Informer())
--- a/pkg/channel/consolidated/reconciler/controller/kafkachannel.go
+++ b/pkg/channel/consolidated/reconciler/controller/kafkachannel.go
@@ -100,10 +100,6 @@ func newRoleBindingWarn(err error) pkgreconciler.Event {
 	return pkgreconciler.NewEvent(corev1.EventTypeWarning, "DispatcherRoleBindingFailed", "Reconciling dispatcher RoleBinding failed: %s", err)
 }
 
-func newSubscribersNotReadyWarn() pkgreconciler.Event {
-	return pkgreconciler.NewEvent(corev1.EventTypeWarning, "SubscribersNotReady", "Subscribers not ready.")
-}
-
 func init() {
 	// Add run types to the default Kubernetes Scheme so Events can be
 	// logged for run types.
@@ -237,12 +233,9 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, kc *v1beta1.KafkaChannel
 		Scheme: "http",
 		Host:   network.GetServiceHostname(svc.Name, svc.Namespace),
 	})
-	reconciled, err := r.reconcileSubscribers(ctx, kc)
+	err = r.reconcileSubscribers(ctx, kc)
 	if err != nil {
 		return fmt.Errorf("error reconciling subscribers %v", err)
-	}
-	if !reconciled {
-		return newSubscribersNotReadyWarn()
 	}
 
 	// Ok, so now the Dispatcher Deployment & Service have been created, we're golden since the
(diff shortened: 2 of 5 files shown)


pkg/channel/consolidated/reconciler/controller/kafkachannel.go
```diff
@@ -225,34 +230,30 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, kc *v1beta1.KafkaChannel
...
 	if err != nil {
-		logger.Errorw("error setting up some subscription status watchers", zap.Error(err))
+		return fmt.Errorf("error reconciling subscribers %v", err)
 	}
-	// close the connection
-	err = kafkaClusterAdmin.Close()
-	if err != nil {
-		logger.Errorw("Error closing the connection", zap.Error(err))
-		return err
+	if !reconciled {
```
lberk, line 244: why wouldn't this be considered a new error type and then we can warn on that?

devguyio: can you explain more? as in consider the case where not all subscribers were reconciled an error and check the error returned by `reconcileSubscribers` instead of a bool and error?

lberk: right.  We're mixing the return value across multiple variables at this point.  For example, the fact we sometimes return `reconciled` and othertimes elect to return a constant (`false`) indicates to me, that something is amiss with the logic as it currently stands.

let's keep the single return value (of type `error`) and have the control flow dependant on the actual error.

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
```diff
@@ -261,29 +262,31 @@ func (r *Reconciler) setupSubscriptionStatusWatcher(ctx context.Context, ch *v1b
 				ObservedGeneration: s.Generation,
 				Ready:              corev1.ConditionTrue,
 			})
+		} else {
+			reconciled = false
```
lberk, line 266: realistically, if we know that we haven't properly reconciled here, why are we even continuing down the rest of the function?

lberk:
https://github.com/knative-sandbox/eventing-kafka/blob/3784fe05d38212764c6130f3db0b7d61b8bb4c29/pkg/channel/consolidated/status/status.go#L186-L221
tbh I'm finding this lambda block really hard to review now.  I'd appreciate it if you could abstract this into its own function with named return variables so I can understand the logic a bit better.  Especially because now we're dealing with prematurely unlocking and re-locking.

pkg/channel/consolidated/status/status.go
```diff
@@ -182,9 +191,26 @@ func (m *Prober) IsReady(ctx context.Context, ch messagingv1beta1.KafkaChannel,
...
 					zap.Any("subscription", sub.UID))
-				return m.checkReadiness(state), true
+				if state.ready {
+					/* This subscription became ready once, that's sufficient for us for now.
+					Ideally we'd have a way to know if each partition already have at least one offset commit, but that's
+					another problem for another day
+					*/
+					return true, true
+				} else {
+					if target.PodIPs.Len() == state.probedPods.Len() && target.PodIPs.HasAll(state.probedPods.List()...) {
```
lberk, line 201: would checking if the set is `Equal()` work?

pkg/channel/consolidated/status/status.go
```diff
@@ -182,9 +191,26 @@ func (m *Prober) IsReady(ctx context.Context, ch messagingv1beta1.KafkaChannel,
...
+						return false, true
+					} else {
+						logger.Debugw("Dispatcher pods changed. Canceling old probes and restarting",
+							zap.Any("subscription", sub.UID))
+						// the dispatcher pods changed, we need to cancel old probing and start again
+						m.mu.Unlock()
```
lberk, line 207: why're we doing the unlocking, cancelprobing, re-lock when it seems like CancelProbing does the same thing as lines 214-216?

I'd prefer an approach where we're not unlocking, calling another function and then relocking like this tbh

pkg/channel/consolidated/status/status.go
```diff
@@ -182,44 +186,25 @@ func (m *Prober) IsReady(ctx context.Context, ch messagingv1beta1.KafkaChannel,
...
-						m.CancelProbing(state.sub)
-						m.mu.Lock()
-						return false, false
-					}
-				}
-			}
-			// Cancel the polling for the outdated version
-			state.cancel()
-			delete(m.targetStates, subscriptionKey)
+		s, o := m.targetStates[sub.UID]
```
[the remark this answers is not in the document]
devguyio: using single letter variable names for such extremely short lived variables is pretty idiomatic in go. I understand it's not readable if the variables would've had a bigger scope, but this is literally a single line scope where you'd know immediately what they from the assignment 3 lines above.

pkg/channel/consolidated/status/status.go
```diff
@@ -467,6 +509,15 @@ func (m *Prober) probeVerifier(item *workItem) prober.Verifier {
 	}
 }
 
+// A target state is outdated if the generation is different or if the target IPs change before it becomes
+// ready.
+func isOutdatedTargetState(s *targetState, sub eventingduckv1.SubscriberSpec, podIPs sets.String) bool {
+	s.readyLock.RLock()
+	defer s.readyLock.RUnlock()
+	r := s.ready
```
lberk, line 517: is adding a single use, single char variable like this for just the next line really more readable than just using `s.ready` ?

pkg/channel/consolidated/status/status.go
```diff
@@ -182,44 +186,25 @@ func (m *Prober) IsReady(ctx context.Context, ch messagingv1beta1.KafkaChannel,
...
-						return false, false
-					}
-				}
-			}
-			// Cancel the polling for the outdated version
-			state.cancel()
-			delete(m.targetStates, subscriptionKey)
+		s, o := m.targetStates[sub.UID]
+		return s, o
+	}()
```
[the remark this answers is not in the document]
devguyio: because of the lock, accessing the map needs to happen under lock and this construct of lock, defer ulock, inside an anonymous function is both idiomatic and less bug prone (since unlocking is deferred)

lberk:
there's quite a few new code paths that unit testing isn't covering.  were there any planned additions to unit tests?

pkg/channel/consolidated/status/status.go
```diff
@@ -354,6 +343,37 @@ func (m *Prober) CancelPodProbing(pod corev1.Pod) {
...
+// RefreshPodProbing lists probe targets and invalidates any in-flight (non-ready) states whose initial probed targets changed from the
+// current ones.
+func (m *Prober) RefreshPodProbing(ctx context.Context) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	logger := logging.FromContext(ctx)
+	for _, state := range m.targetStates {
+		if !m.checkReadiness(state) {
+			// This is an in-flight state
+			sub := state.sub
```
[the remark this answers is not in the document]
devguyio: the one on line 365 was overlooked.

pkg/channel/consolidated/status/status.go
devguyio: nope, that function changed a lot and that one was overlooked

devguyio:
@lberk yes and I created this issue to increase the unit tests coverage of the prober #491

pkg/channel/consolidated/status/status.go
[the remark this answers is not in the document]
devguyio: the whole idea is to rely on go's defer reliability which the go runtime will ensure to take place even if the call panics, something that the normal call to 'unlock' doesn't guarantee. Check the example by the end here https://blog.golang.org/defer-panic-and-recover

pkg/channel/consolidated/status/status.go
```diff
@@ -168,11 +170,13 @@ func (m *Prober) checkReadiness(state *targetState) bool {
 		zap.Any("channel partitions", partitions),
 		zap.Any("ready partitions", state.readyPartitions.List()),
 	)
-	return state.readyPartitions.Len() == int(partitions)
+	if !state.ready {
+		state.ready = state.readyPartitions.Len() == int(partitions)
```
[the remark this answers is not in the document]
devguyio: `Len()` return type is `int`

pkg/channel/consolidated/status/status.go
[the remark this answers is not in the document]
devguyio: I changed the one on 365.

devguyio:
@lberk if you've done reviewing, do you mind giving me a summary and what does this PR needs to be merged from your POV?

slinkydeveloper (commented):
The idea sounds correct to me, but I'll let somebody else approve it :)

pkg/channel/consolidated/status/status.go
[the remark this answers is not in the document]
matzew: perhaps adding a comment, on why?

the debate here shows, there is reasons, but also questions.  Comments usually help to get larger context

lberk:
> @lberk yes and I created this issue to increase the unit tests coverage of the prober #491

So basically, no.  There will be no futher unit test additions with this PR.

pkg/channel/consolidated/reconciler/controller/controller.go
```diff
@@ -157,9 +157,18 @@ func NewController(
 			DeleteFunc: func(obj interface{}) {
 				pod, ok := obj.(*corev1.Pod)
 				if ok && pod != nil {
-					logger.Debugw("Dispatcher pod deleted. Canceling pod probing.",
+					logger.Debugw("Dispatcher pod deleted. Refreshing pod probing.",
```
lberk, line 160: If these two functions are essentially copy and pasted (save for a debug line).  Abstract the identical code into a different function and pass it with the debug string.

matzew: I agree here, that `DeleteFunc` and `Addfunc` are very very identical - let's not start copy/paste here for no reason

matzew:
> So basically, no. There will be no futher unit test additions with this PR.

@lberk I see where you coming from, and I think we should always try have coverage of code via unit tests. 

I lack context of why the unit tests for the `status.go` are part of a different issue/PR and can not be part of this PR...

But I am fine in adding that afterwards, FWIW.

pkg/channel/consolidated/status/status.go
```diff
@@ -317,6 +320,11 @@ func (m *Prober) Start(done <-chan struct{}) chan struct{} {
 func (m *Prober) CancelProbing(sub eventingduckv1.SubscriberSpec) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
+	m.forgetState(sub)
+}
+
+// This is meant to be called while m.mu is locked
+func (m *Prober) forgetState(sub eventingduckv1.SubscriberSpec) {
```
lberk, line 327: could you elaborate to me why it's a sound design decision to have:
`CancelProbing` (which is part of the public Manager interface)
`CancelProbing`'s single job is to aquire the mutex lock, and call `forgetState(sub)`
`forgetState(sub)` has a comment a comment that indicates that we *must* have the lock before calling it
we then have other ways of calling `forgetState` elsewhere in the file.

Why aren't we using `CancelProbing` as the proper entrypoint my consistently?

devguyio:
> So basically, no. There will be no futher unit test additions with this PR.

@lberk I misunderstood what you meant earlier by "were there any planned additions to unit tests?" I thought you meant "do you plan on adding unit tests later after this PR". And indeed I did, so I went ahead and created the issue to confirm that there's an issue and planned work for adding unit tests.

> I lack context of why the unit tests for the status.go are part of a different issue/PR and can not be part of this PR...

@matzew Reasons are: 

1. Time pressure.
2. We have some e2e tests in place that're giving us some level confidence.

lberk:
> @matzew Reasons are:
> 
>     1. Time pressure.
>     2. We have some e2e tests in place that're giving us some level confidence.

Neither of these are valid criterion.

devguyio:
@lberk I put it on hold to add unit tests and increase the coverage. Please let me know what else is needed for the PR to be approved? I'm addressing some comments too and responded to the others.

pkg/channel/consolidated/status/status.go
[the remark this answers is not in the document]
devguyio: sure, will add a comment

pkg/channel/consolidated/status/status.go
matzew: OK, so the issue is:

for outside invocation he have the `CancelProbe`, that basically does the lock and calls the private one.

Internally we call the `forget` since we have already taken care of the lock.

Perhaps the different names of `CancelProbe` versus `froget...` is confusing?

I think I understand the way of decorating the invocation of the `forget` with a lock, inside of the `CancelProbe`.

So...  Let's perhaps change the name?  `internalCancelProbe`, and be more explicit that it's only used w/in a mutex, and outside users should be instead calling `CancelProbe` ? 

And perhaps add a TODO... to have that sorted later

lberk:
https://github.com/knative-sandbox/eventing-kafka/blob/9128624c5e5a7d0f4f684585bb25e98b8a909bb8/pkg/channel/consolidated/status/status.go#L110
is that ingress line still valid here?

devguyio:
> is that ingress line still valid here?
Nope, it was overlooked

devguyio:
@lberk @matzew I increased the test coverage , please check.
/unhold

devguyio:
Checking the failed tests

pkg/channel/consolidated/status/status.go
```diff
@@ -361,15 +361,13 @@ func (m *Prober) RefreshPodProbing(ctx context.Context) {
 					zap.Any("subscription", sub.UID))
 				return
 			}
-			if !state.probedPods.Equal(target.PodIPs) {
```
lberk: just so I understand, why was this condition no longer needed?

devguyio: To force refreshing the Pods regardless of whether the IPs changed or not. We can't for sure know if pods with the same IPs are the same ones or not.

pkg/channel/consolidated/reconciler/controller/controller.go
```diff
@@ -154,26 +154,22 @@ func NewController(
...
 	})
 
 	return impl
 }
+
+func getPodInformerEventHandler(ctx context.Context, logger *zap.SugaredLogger, statusProber *status.Prober, impl *controller.Impl, kafkaChannelInformer kafkavmessaging1beta1.KafkaChannelInformer, handlerType string) func(obj interface{}) {
+	return func(obj interface{}) {
+		pod, ok := obj.(*corev1.Pod)
+		if ok && pod != nil {
+			logger.Debugw("%s rods. Refreshing pod probing.", handlerType,
```
lberk, line 169: :) small typo

pkg/channel/consolidated/reconciler/controller/controller.go
```diff
@@ -19,16 +19,15 @@ package controller
...
-
 	"github.com/kelseyhightower/envconfig"
 	"go.uber.org/zap"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/cache"
 
 	"knative.dev/eventing-kafka/pkg/apis/messaging/v1beta1"
 	"knative.dev/eventing-kafka/pkg/channel/consolidated/status"
+	kafkavmessaging1beta1 "knative.dev/eventing-kafka/pkg/client/informers/externalversions/messaging/v1beta1"
```
lberk, line 30: kafkavmessaging1beta1 or kafkamessagingv1beta1

devguyio: woops, typo. Good catch

pkg/channel/consolidated/status/status_test.go
```diff
@@ -39,110 +39,169 @@ import (
...
+
+	// Initially, lister points to d1
+	lister := fakeProbeTargetLister{
+		target: getTargetLister(t, serverD1.URL),
+	}
+
+	prober, ready := getProber(t, &lister)
+
 	done := make(chan struct{})
 	cancelled := prober.Start(done)
```
lberk, line 182: hm, just wondering out loud here, should we be `defer close(probeRequestsDX` elsewhere as well?

lberk:
/lgtm

Thanks for adding the tests @devguyio

pkg/channel/consolidated/status/status_test.go
devguyio: It wouldn't hurt, but it's not needed IMHO. The reason those channels are closed is to shutdown the prober and free the polling goroutines in the prober and not leak those routines. While closing the probeRequests won't be necessary since we're closing the whole server anyway and that'll clean up their goroutines IIUC. And as you know, you don't need to close all the channels since they'll be garbage collected if unreachable.

If you feel I should do that, sure I can. Let me know.

devguyio:
@lberk sure, np. Are you gonna wait for someone else to approve the PR?

lberk (approved):

lionelvillard (approved):

matzew (approved):
/lgtm
/approve

PR #489, opened 2021-03-29
Add "cloudevents" to the eventing-kafka config

eric-sap:
## Proposed Changes

The consolidated channel allows configuration of the cloudevents args (max idle connections, max idle connections per host), but the distributed channel did not have this capability.  This PR will allow configuration of these fields via the config-kafka configmap.  For example:
```
  eventing-kafka: |
    cloudevents:
      maxIdleConns: 50
      maxIdleConnsPerHost: 5
```

If no "cloudevents" field is present in the eventing-kafka YAML string, the original behavior (1000 maxIdleConns, 100 maxIdleConnsPerHost) is maintained, as is true for the consolidated channel equivalents.


aliok:
/lgtm

BTW, this cloud connection arguments thing is one of the reasons I was suggesting we should disable on-the-fly reloading of kafka config since this function can be set once per process.

pkg/channel/distributed/common/constants/constants.go
```diff
@@ -0,0 +1,24 @@
...
+*/
+
+package constants
+
+// Global Constants
+const (
+	// Values Used When Initializing Cloud Events
+	DefaultMaxIdleConns        = 1000
+	DefaultMaxIdleConnsPerHost = 100
+)
```
matzew, line 24: We have similar already here:

https://github.com/knative-sandbox/eventing-kafka/blob/main/pkg/channel/distributed/dispatcher/constants/constants.go#L22-L23

Folks, would it be possible to move all these constants to just one common file, to have em in a single place?

matzew: the above is just one example - there are more occurrences in our codebase

eric-sap: Certainly; I've moved them to pkg/common/constants/constants.go.  It should at least be easier to move them elsewhere if desired from there.

matzew:
@eric-sap mind rebasing? looks like there is a merge conflict

Otherwise looks good - thanks for moving the files around!

eric-sap:
Of course; merge testing should be complete shortly

matzew (approved):
/lgtm
/approve

PR #493, opened 2021-03-31
Move metrics shutdown channels out of dispatcher config

eric-sap:
## Proposed Changes

-🐛 The introduction of the Sarama metrics to the dispatcher (https://github.com/knative-sandbox/eventing-kafka/pull/464) included two channels (MetricsStopChan and MetricsStoppedChan) that were inappropriately added to the DispatcherConfig instead of the DispatcherImpl.  Since the config is re-used between creation of dispatchers, these closed channels were copied to the new DispatcherImpl's config and caused a panic when Shutdown is called again.  Moving these channels to the DispatcherImpl, the fields of which are recreated each time, solves this issue.


travis-minke-sap (approved):
lgtm

PR #499, opened 2021-04-02
[WIP] Consolidated ch dispatcher restart when config changes

aliok:
To be rebased: builds on top of https://github.com/knative-sandbox/eventing-kafka/pull/486

## Proposed Changes

-
-
-

Similar to https://github.com/knative-sandbox/eventing-kafka/pull/486, tested with https://github.com/aliok/kafka-ch-reload-tests

Things in that repo starts a receiver and a sender then records the messages in the receiver. Later, I change `config-kafka` and the dispatcher kills itself. A new one is created shortly after.

Here are the results:
```
sender:
-----------
tried to send 11748 messages
Success:11653
Errors:95

receiver:
----------------
11925 received
manual investigation: 272 received duplicates
manual investigation: unique received: 11653
```

So, there's some downtime and the sender cannot send the messages during that time.

How should we handle this?

I think I would see some downtime in distributed channel too which is doing on-the-fly reloading. Though, the duration would be shorter.

Alternatively, can we achieve this zero downtime with a proper rolling deployment? Like, trigger a rolling deployment of the dispatcher in the channel controller. Kube should handle the rest, right?

### DIFF
--- a/pkg/channel/consolidated/reconciler/controller/controller.go
+++ b/pkg/channel/consolidated/reconciler/controller/controller.go
@@ -107,8 +107,21 @@ func NewController(
 	)
 	r.statusManager = statusProber
 	statusProber.Start(ctx.Done())
+
+	// Call GlobalResync on kafkachannels.
+	grCh := func(obj interface{}) {
+		logger.Info("Changes detected, doing global resync")
+		impl.GlobalResync(kafkaChannelInformer.Informer())
+	}
+
+	handleKafkaConfigMapChange := func(ctx context.Context, configMap *corev1.ConfigMap) {
+		logger.Info("Configmap is updated or, it is being read for the first time")
+		r.updateKafkaConfig(ctx, configMap)
+		grCh(configMap)
+	}
+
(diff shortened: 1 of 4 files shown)


slinkydeveloper (commented):
I like it

aliok:
What do you think about the rolling deployment approach ideally with 0 downtime?

slinkydeveloper:
> What do you think about the rolling deployment approach ideally with 0 downtime?

It's never going to be 0 downtime, but yeah maybe it's more "delicate" if you get what I mean...

slinkydeveloper:
@aliok I was more thinking at producer side, when there are 2 instances up, a message could be accepted but never pushed... But in theory, it shouldn't happen because we wait the ack from kafka for every message...

So yeah maybe we don't have any race conditions, just wanted to make sure you double checked it :smile:

aliok:
Ok, I am trying the approach above here: https://github.com/knative-sandbox/eventing-kafka/pull/514

Still a WIP

aliok:
/close 
Gonna go with https://github.com/knative-sandbox/eventing-kafka/pull/514

PR #510, opened 2021-04-07
Dependency update and some hack/test

matzew:

Fixes #

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**


test/e2e-common.sh
```diff
@@ -564,3 +564,10 @@ function parse_flags() {
   esac
   return 0
 }
+
+## hack
+function get_latest_knative_yaml_source() {
+  echo "https://storage.googleapis.com/knative-releases/eventing/latest/eventing.yaml"
+}
+
+readonly KNATIVE_EVENTING_RELEASE="$(get_latest_knative_yaml_source)"
```
devguyio, line 573: I think you missed removing the condition on line 174 to always install eventing from that version?

matzew: donks!

I indeed forgot that

devguyio: turns out that URL resolves to nightly releases :disappointed: 
```
>> Install Knative Eventing from https://storage.googleapis.com/knative-nightly/eventing/latest/eventing.yaml
```

devguyio: one quick fix to unblock main maybe is to git checkout latest release and use this funcion to get the latest release? https://gist.github.com/lukechilds/a83e1d7127b78fef38c2914c4ececc3c 
so the code in else (lines 178..) can continue but on the checked out release?

just thinking out loud

devguyio (changes requested):
It still installs from nightly, left a comment with an idea

test/e2e-common.sh
matzew: let's see what the tmp URL with explicit pin-point says

matzew:
@devguyio   if I directly apply 0.22  ... or check it out.... there is IMO no difference .

devguyio:
@matzew yeah sure, I was mainly saying to "detect" the latest release, but I'm fine having this go in for now and fixing it properly after we're unblocked

devguyio (approved):
lgtm

PR #524, opened 2021-04-08
Run KafkaSource smokes in upgrade test

### LINKED ISSUE #67: Upgrade tests for KafkaSource
We need to implement upgrade tests for KafkaSource and KafkaBinding so that in the future we can change the storage version to v1beta1 from v1alpha1.

Best idea is to keep eventing at latest release and only changing the eventing-kafka version. We can in the future run the tests again with eventing head too.

Part of umbrella issue, #65

cardil:
Partially resolves #67 

## Proposed Changes

- Run pre, post smoke tests for `KafkaSource` on upgrade tests

Continual tests will follow to fully resolve #67


test/e2e/kafka_source_test.go
```diff
@@ -1,5 +1,5 @@
-//+build e2e mtsource
-//+build source
+// +build e2e
+// +build source mtsource
```
cardil, line 2: I think this is more correct. 

It means: `e2e AND (source OR mtsource)`. 

Previous have meant: `(e2e or mtsource) AND source`, which is weird.

test/e2e-common.sh
```diff
@@ -315,12 +315,40 @@ function install_consolidated_channel_crds {
   wait_until_pods_running "${SYSTEM_NAMESPACE}"
 }
 
+function install_released_consolidated_source {
```
lionelvillard, line 318: I would remove `consolidated`. Just  `install_released_source`. 

I suppose the upgrade tests are black-box tests, in which case they can also run against the multi-tenant source variant.

cardil: The function name contained `consolidated` word before, that's why I didn't remove it.

That's right those tests should be a black box, but I believe future conditional logic should be in golang in installation package rather than bash scripts.

mgencur: My two cents: After reading `install_released_consolidated_source` my first question is - what "source" ? It could be "source code". Maybe we could say `install_released_event_source` instead.

test/e2e/kafka_source.go
```diff
@@ -0,0 +1,439 @@
...
+
+	kafkaSASLSecret = "strimzi-sasl-secret"
+	kafkaTLSSecret  = "strimzi-tls-secret"
+)
+
+var (
+	testMtSource = os.Getenv("TEST_MT_SOURCE")
+)
+
+// SourceTestScope a func that will return true if we should proceed with given
```
mgencur, line 61:
```suggestion
// SourceTestScope returns true if we should proceed with given
```

test/upgrade/continual.go
```diff
@@ -67,6 +67,22 @@ func ChannelContinualTest() pkgupgrade.BackgroundOperation {
 		"ChannelContinualTest", setup, verify)
 }
 
+// SourceContinualTest tests source operation in continual manner during the
+// whole upgrade and downgrade process asserting that all event are propagated
+// well.
+func SourceContinualTest() pkgupgrade.BackgroundOperation {
+	setup := func(c pkgupgrade.Context) {
+		// TODO: not yet implemented
```
mgencur, line 75: I suppose this comment is redundant? Same for the comment a few lines below

test/upgrade/postdowngrade.go
```diff
@@ -28,3 +28,11 @@ func ChannelPostDowngradeTest() pkgupgrade.Operation {
 			runChannelSmokeTest(c.T)
 		})
 }
+
+// SourcePostDowngradeTest tests source operations after downgrade.
+func SourcePostDowngradeTest() pkgupgrade.Operation {
```
mgencur, line 33: I don't know but does it make sense to rename all these functions to have the `Event` prefix? 
E.g. EventSourceDowngradeTest

mgencur (commented):
Looks very good! Just a few comments here and there

test/upgrade/postdowngrade.go
cardil: This repo contains Kafka things. KafkaSource and KafkaChannel. I'm shorting Kafka, but not only I. That naming was already in this repo before. Let's not create new names like enentsource.

cardil: This repo contains Kafka things. KafkaSource and KafkaChannel. I'm shorting Kafka, but not only I. That naming was already in this repo before. Let's not create new names like enentsource.

aliok:
Looks very good to me :)
Thanks!

Leaving the merge to @mgencur

test/upgrade/continual.go
cardil: I'd leave it be. The `// TODO:` is reported by IDE and other dev tools. Warn log is useful on logs. That's why it's repeated.

test/upgrade/postdowngrade.go
mgencur: It looks like these two (Source and KafkaSource) are totally mixed in function names in this repo. E.g. runSourceSmokeTest vs. AssureKafkaSourceIsOperational vs. testKafkaSource vs. SourcePostDowngradeTest
That's a cosmetic thing, having just "Source" is a bit more misleading to me. Nevermind

aliok (approved):
/lgtm
/approve

PR #525, opened 2021-04-08
Fix memory and CPU leaks with the distributed secret watcher

eric-sap:
## Proposed Changes

- 🐛 Replaced the secret watcher with a secret informer; otherwise the watcher times out after 30-60 minutes, resulting in secrets not being watched _and_ the select loop spinning the CPU out of control.
-🐛  Replaced calls to time.After() with a single re-used timer in infinite channel select loops (secret watcher, metrics loops) as these can cause some severe memory leaks in cases where the select loop is run frequently (noticed due to the other bug)
-🐛 Setting the Sarama SASL type in the kafka-cluster secret to an empty string was resulting in any secret update causing a reconfigure to happen on dispatchers and receivers (because "" isn't literally the default SASL type of "PLAIN" during the comparison)

### DIFF
--- a/cmd/channel/distributed/dispatcher/main.go
+++ b/cmd/channel/distributed/dispatcher/main.go
@@ -164,7 +164,7 @@ func main() {
 	}
 
 	// Watch The Secret For Changes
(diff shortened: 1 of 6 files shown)


travis-minke-sap (approved):
lgtm!

PR #514, opened 2021-04-07
[consolidated] Rolling deployment dispatcher on configmap change

aliok:
Instead of doing more complex things (making the pods commit suicide, on-the-fly reloading, shaving horde of yaks etc) within the dispatcher to reload the configuration, just watch the configmap in the controller and do a rolling update on the dispatcher deployment.

Hashes the configmap, stores it on the dispatcher deployment as an annotation. When configmap is changed (watching it in the controller), update it on the deployment annotation. That will make Kubernetes to do a rolling deployment.

I tested my changes with https://github.com/aliok/kafka-ch-reload-tests. There are 2 pods:
- sender sends messages for 2 mins
- receiver receives messages for 3 mins

With a message frequency of 10 ms, there were only 4 times the sender was unable to send messages to the channel during the rolling deployment.

## Proposed Changes

- Rolling deployment dispatcher on configmap change
- Alternative approach to https://github.com/knative-sandbox/eventing-kafka/pull/499, as discussed in that ticket

**Release Note**

```release-note
- 🧽  Consolidated Kafka Channel dispatcher is now reconfigured when `config-kafka` configmap is updated

```

**Docs**

Issue #455, opened 2021-03-11
Remove on-the-fly reloading of `config-kafka` in distributed channel


aliok:
**Problem**
I started with on-the-fly reloading of `config-kafka` in consolidated channel: https://github.com/knative-sandbox/eventing-kafka/issues/318

Distributed channel already does this, but it is missing in consolidated channel.

I tried this for a while but in the end the code was getting too complex and also there were some technical issues (e.g. some stuff read from `config-kafka` can only be set once in the underlying libraries, otherwise memory leaks can happen). In the end, I decided to cancel this.

Since we're trying to align both channels, I think it is a good idea to also remove this reloading in distributed channel (their implementation also suffers from the problems above). @travis-minke-sap  agreed that we should remove this in there as well: https://groups.google.com/g/knative-dev/c/u1-P03-VZh0

**UPDATE:**
- Consolidated channel now reloads the configuration. However, dispatcher doesn't do on-the-fly reloading. Instead it restarts.
- https://github.com/knative-sandbox/eventing-kafka/pull/423: Watch `config-kafka` in controller
- https://github.com/knative-sandbox/eventing-kafka/pull/486: Do global resync of everyting if `config-kafka` is updated. This triggers reconciliation of the CRs and dispatcher deployments.
- https://github.com/knative-sandbox/eventing-kafka/pull/514: Trigger a rolling deployment of dispatcher in case the configmap is updated

We should do the same in distributed channel too.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)**
Developer

**Exit Criteria**
No more on-the-fly reloading of `config-kafka` in distributed channel

**Time Estimate (optional):**
How many developer-days do you think this may take to resolve?

**Additional context (optional)**
Ticket for adding this in consolidated channel: https://github.com/knative-sandbox/eventing-kafka/issues/318
ML discussion: https://groups.google.com/g/knative-dev/c/u1-P03-VZh0

PR #529, opened 2021-04-09
Distributed channel - watch config-kafka like consolidated channel

aliok:
Addresses controller part of #455.

I tried to do things in small steps in separate PRs.

## Proposed Changes

- Watch the configmap in the controller (already existing)
- When the configmap is changed, do a global resync of the KafkaChannel CRs so that they're reconciled

- Bonus: copy over some good stuff from distributed channel to consolidated channel (some nil checks etc. I found missing in consolidated channel)

I tested my changes with https://github.com/aliok/kafka-ch-reload-tests/tree/distr-channel (beware the `distr-channel` branch). There are 2 pods:
- sender sends messages for 2 mins
- receiver receives messages for 3 mins

I haven't seen any message lost (sent to Kafka but not received by the receiver) or non-posted (couldn't sent by the sender to Kafka).

FYI, this the plan I have in my mind:
- Do a global resync of CRs when configmap changes
- Remove configmap watching in dispatcher and receiver and mount the configmap to these
- Controller to trigger a rolling update of dispatcher and receiver deployments when configmap changes (similar to https://github.com/knative-sandbox/eventing-kafka/pull/514)

With these done, the behaviour and code will look very very similar on both channel implementations when it comes to handling the changes in `config-kafka`.

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/channel/consolidated/reconciler/controller/kafkachannel.go
+++ b/pkg/channel/consolidated/reconciler/controller/kafkachannel.go
@@ -568,10 +568,22 @@ func (r *Reconciler) deleteTopic(ctx context.Context, channel *v1beta1.KafkaChan
 
 func (r *Reconciler) updateKafkaConfig(ctx context.Context, configMap *corev1.ConfigMap) {
 	logger := logging.FromContext(ctx)
+
+	if r == nil {
+		// This typically happens during startup and can be ignored
+		return
+	}
+
+	if configMap == nil {
+		logger.Warn("Nil ConfigMap passed to configMapObserver; ignoring")
+		return
+	}
+
 	logger.Info("Reloading Kafka configuration")
 	kafkaConfig, err := utils.GetKafkaConfig(configMap.Data)
 	if err != nil {
 		logger.Errorw("Error reading Kafka configuration", zap.Error(err))
+		return
 	}
 
 	if kafkaConfig.AuthSecretName != "" {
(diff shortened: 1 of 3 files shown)


aliok:
cc @eric-sap mind having a look?

eric-sap:
From what I can tell, this should do "what it did before" in the distributed channel when a the configmap changes, but also do a resync (re-reconcile).  Right now that won't do anything but could eventually use the hash differences to do rollout-restart-style updates of the deployments.   If that's an accurate summary then it looks good; thanks!

aliok:
> Right now that won't do anything but could eventually use the hash differences to do rollout-restart-style updates of the deployments.

Yes, exactly.

Thanks for the review.

Let me ping @devguyio @slinkydeveloper for lgtm/approve.

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
```diff
@@ -568,10 +568,22 @@ func (r *Reconciler) deleteTopic(ctx context.Context, channel *v1beta1.KafkaChan
...
+	if configMap == nil {
+		logger.Warn("Nil ConfigMap passed to configMapObserver; ignoring")
+		return
+	}
+
 	logger.Info("Reloading Kafka configuration")
 	kafkaConfig, err := utils.GetKafkaConfig(configMap.Data)
 	if err != nil {
 		logger.Errorw("Error reading Kafka configuration", zap.Error(err))
+		return
```
slinkydeveloper, line 586: Should you return an error here?

aliok: I will give it a try

aliok: This function is not called within the reconciliation loop. It is called when config-kafka is changed.
A global resync is done later.

slinkydeveloper: ah ok gotcha!

slinkydeveloper (approved):
/lgtm

PR #537, opened 2021-04-14
Distributed channel - config-kafka reloading

### LINKED ISSUE #455: Remove on-the-fly reloading of `config-kafka` in distributed channel

aliok:
Fixes #455

## Proposed Changes

1. Mount config-kafka to controller, dispatcher and receiver
2. Don't watch config-kafka in dispatcher and receiver
3. Make changes in the controller so that it does a redeployment of dispatcher and receiver when config-kafka changes

**Release Note**

```release-note
- 🧽  Distributed Kafka Channel dispatcher and receiver are now reconfigured when `config-kafka` configmap is updated

```

**Docs**


pkg/channel/distributed/common/kafka/sarama/sarama.go
```diff
@@ -45,28 +41,19 @@ func EnableSaramaLogging(enable bool) {
...
 // The Provided Context Must Have A Kubernetes Client Associated With It
-func LoadSettings(ctx context.Context, clientId string, kafkaAuthConfig *client.KafkaAuthConfig) (*sarama.Config, *commonconfig.EventingKafkaConfig, error) {
-	if ctx == nil {
-		return nil, nil, fmt.Errorf("attempted to load settings from a nil context")
-	}
-
-	configMap, err := kubeclient.Get(ctx).CoreV1().ConfigMaps(system.Namespace()).Get(ctx, constants.SettingsConfigMapName, metav1.GetOptions{})
-	if err != nil {
-		return nil, nil, err
+func LoadSettings(ctx context.Context, clientId string, configMap map[string]string, kafkaAuthConfig *client.KafkaAuthConfig) (*sarama.Config, *commonconfig.EventingKafkaConfig, error) {
```
eric-sap, line 44: If we're going to have this configmap "loader" inside the context anyway, would it save some coding complexity (i.e. having to call configmaploader.FromContext(ctx) and the configmapLoader() call itself, including all of the error handling) to just have this LoadSettings function extract the configmaploader function from the context (which is already a parameter here) and call it, instead of calling that in each place that's calling LoadSettings() ?  It would involve fiddling with the tests a bit and I am not 100% sure the result would be "better" but thought I would mention it.

aliok: Sounds good. 
I have some changes stashed, but I will get back at this before merging the PR.

aliok: @eric-sap this is done now, thanks for the suggestion.

eric-sap: I like it; thanks!

aliok: I had to revert this later because I needed to load the config outside this function as well.
Specifically, when I needed to compute the configmap hash on controller start.

aliok:
@eric-sap 
Pushed new commits that address the 3rd part in the PR description.
Sorry the PR got big now :(

I also did 1-2 things on the consolidated channel code in this PR, since I found some trivial problems when I compared the code with distributed channel.

aliok:
/hold
I haven't done the manual tests yet

aliok:
/unhold

I tested my changes with https://github.com/aliok/kafka-ch-reload-tests. There are 2 pods:

- sender sends messages for 2 mins
- receiver receives messages for 3 mins with a message frequency of 10 ms

There were no messages lost and no messages undelivered. Also no duplicate messages. I don't get how that's possible actually.

@eric-sap would that be expected? The channel would be unavailable for a few milliseconds in the consolidated channel when `config-kafka` is changed and thus sender can't send a few messages.
This doesn't happen here in distributed channel. 
Perhaps, we're reporting the consolidated dispatcher ready before it is actually ready.

@eric-sap I would use some help in manual testing if receiver deployment is updated when config-kafka is changed. Any instructions would be nice to make the reconcilers create a receiver deployment.

eric-sap:
> @eric-sap would that be expected? The channel would be unavailable for a few milliseconds in the consolidated channel when `config-kafka` is changed and thus sender can't send a few messages.
> This doesn't happen here in distributed channel.

I haven't looked at all of the code here yet, but ideally when a deployment is changed, Kubernetes starts the new pod and does not switch the target or terminate the old pod until the readiness check says that the new pod is okay.  In the distributed channel's case, that means calling health.SetDispatcherReady() (in the dispatcher's main.go) and the SetChannelReady/SetProducerReady calls made after the receiver's components are created.  Theoretically this means that the instant the pod is "ready" it really can begin processing messages (and local strimzi testing has borne this out).

That said, I have seen duplicate messages happen during rollout restarts when using an external kafka broker and not local strimzi.  Our guarantee is only that no messages will be lost (that is, if a client receives a 202 from the receiver, that means the message will be delivered "at least one" (or retried until final failure and possible redelivery to a dead-letter-queue if configured).

pkg/channel/distributed/common/kafka/sarama/sarama.go
eric-sap: Doh; thanks for the effort!  Sorry for the extra time!

aliok:
/hold
Receiver is not restarted. Having a look now

aliok:
/unhold
Thanks @eric-sap.
I pushed the code change that fixes global resync in kafkaSecret controller

pkg/channel/distributed/controller/kafkachannel/dispatcher.go
```diff
@@ -300,6 +301,37 @@ func (r *Reconciler) reconcileDispatcherDeployment(ctx context.Context, logger *
 
 		// Log Deletion Timestamp & Finalizer State
 		if deployment.DeletionTimestamp.IsZero() {
+			deploymentCopy := deployment.DeepCopy()
+			needsUpdate := false
+
+			if deploymentCopy.Spec.Template.Annotations == nil {
+				logging.FromContext(ctx).Infof("Configmap hash is not set. Updating the dispatcher deployment.")
```
eric-sap, line 308: This is probably because of grabbing code from the consolidated side, but we already have the 'logger' variable available in this function, so re-extracting it from the context here(and a couple of other places; won't comment on them all) isn't necessary.

pkg/channel/distributed/controller/kafkasecret/controller.go
```diff
@@ -79,11 +93,30 @@ func NewController(ctx context.Context, _ configmap.Watcher) *controller.Impl {
 		kafkachannelLister: kafkachannelInformer.Lister(),
 		deploymentLister:   deploymentInformer.Lister(),
 		serviceLister:      serviceInformer.Lister(),
+		kafkaConfigMapHash: commonconfig.ConfigmapDataCheckSum(configMap),
 	}
 
 	// Create A New KafkaSecret Controller Impl With The Reconciler
 	controllerImpl := kafkasecretinjection.NewImpl(ctx, r)
 
+	// Call GlobalResync on kafkachannels.
```
eric-sap, line 102: This comment doesn't match the code any more

pkg/channel/distributed/controller/kafkasecret/receiver.go
```diff
@@ -226,6 +226,31 @@ func (r *Reconciler) reconcileReceiverDeployment(ctx context.Context, logger *za
...
+			if deploymentCopy.Spec.Template.Annotations == nil {
+				logging.FromContext(ctx).Infof("Configmap hash is not set. Updating the receiver deployment.")
+				deploymentCopy.Spec.Template.Annotations = map[string]string{
+					commonconstants.ConfigMapHashAnnotationKey: r.kafkaConfigMapHash,
+				}
+				needsUpdate = true
+			}
+
+			if deploymentCopy.Spec.Template.Annotations[commonconstants.ConfigMapHashAnnotationKey] != r.kafkaConfigMapHash {
+				logging.FromContext(ctx).Infof("Configmap hash is changed. Updating the receiver deployment.")
```
eric-sap, line 241: Here also the logger variable is already available to use

pkg/channel/distributed/controller/kafkachannel/dispatcher.go
aliok: Yeah, copy-paster error. Pushed a new commit now.

eric-sap:
I've tested this locally and it seems to do what it's supposed to.  There are still some linter errors but other than that it looks good to me.

aliok:
Linter error is about a deprecated dependency in test images. We need to fix it, but it is irrelevant to this PR.

@travis-minke-sap can you have a look and lgtm it?

travis-minke-sap (approved):
Thanks Ali !

PR #548, opened 2021-04-16
Restart distributed deployments when changed

### LINKED ISSUE #502: Distributed channel dispatcher/receiver deployments not updated after upgrade
**Describe the bug**
The distributed channel controller has no mechanism to identify and update dispatcher or receiver image changes during reconciliation.  This leaves pods with their original image until the deployments are manually deleted or modified (via "rollout restart" or similar)

 **Expected behavior**
When a KafkaChannel is reconciled, any dispatcher deployments that don't specify the current image should be updated (allowing Kubernetes to do a restart of any affected pods).  Secrets reconciliation and receiver deployments should behave similarly.

 **To Reproduce**
Install an old version of eventing-kafka and upgrade to a newer version.  Notice that when a KafkaChannel is reconciled, the associated dispatcher pod remains running with the old image.

 **Knative release version**
0.21.3

 **Additional context**
Restarting the deployments when the image changes is the bare minimum, but ideally we would like to, during reconciliation, check all of our settings against the existing deployments (not just the image) and reconfigure them as necessary.  In addition, the services would be good to verify/update during this process.

eric-sap:
Fixes #502

## Proposed Changes

- 🎁 Modified the distributed channel's receiver and dispatcher reconcilers to trigger new deployments and/or services when relevant pieces of the the desired YAML for those resources changes.
- 🧽 Moved some of the logic from the fix for #455 (https://github.com/knative-sandbox/eventing-kafka/pull/537) into the new CheckDeploymentChanged() function in pkg/channel/distributed/controller/util/compare.go (triggers on any change to the required annotations, which includes the configmap hash).

No user-actions should be necessary, but this code does alter the behavior of installing a new version of the eventing-kafka distributed channel in that previously a manual deletion of the receiver and dispatchers' deployments was necessary to have their pod code updated.  After this change those rollouts will happen automatically as the controller realizes that the image fields are different.

**Release Note**

Changes to dispatcher/receiver values in the distributed controller (e.g. Image) now updates those deployments automatically.


eric-sap:
@aliok I moved the configmap hash annotation checking logic from the reconciler functions directly into the CheckDeploymentChanged() function.  It lost some of the granularity of the explicit log messages for "Configmap hash is not set" and "Configmap hash is changed" because now it's just checking "all annotations" -- if those are useful to have as explicit messages I would be happy to put those particular checks back; let me know.

pkg/channel/distributed/controller/kafkachannel/dispatcher.go
```diff
@@ -144,20 +148,35 @@ func (r *Reconciler) reconcileDispatcherService(ctx context.Context, logger *zap
 		}
 	} else {
 
+		// Determine whether the existing service is different in a way that demands a patch
+		// such as missing required labels or spec differences
+		patch, needsUpdate := util.CheckServiceChanged(logger, existingService, newService)
+
```
travis-minke-sap, line 154: Can this be moved down a bit (after deletion timestamp check just before `if needsUpdate`?  Just in the spirit of declaring things close to where used.

pkg/channel/distributed/controller/kafkachannel/dispatcher.go
```diff
@@ -144,20 +148,35 @@ func (r *Reconciler) reconcileDispatcherService(ctx context.Context, logger *zap
...
-			if util.HasFinalizer(r.finalizerName(), &service.ObjectMeta) {
+			if util.HasFinalizer(r.finalizerName(), &newService.ObjectMeta) {
 				logger.Info("Blocking Pending Deletion Of Dispatcher Service (Finalizer Detected)")
 			} else {
 				logger.Warn("Unable To Block Pending Deletion Of Dispatcher Service (Finalizer Missing)")
 			}
 		}
 
-		// Return Success
-		return nil
```
travis-minke-sap, line 159: I think the finalizer log statements above were relying on this common exit `return nil` and might not make sense without it?

pkg/channel/distributed/controller/kafkachannel/dispatcher.go
```diff
@@ -144,20 +148,35 @@ func (r *Reconciler) reconcileDispatcherService(ctx context.Context, logger *zap
...
+		// Patch the service in Kubernetes if necessary
+		if needsUpdate {
+			logger.Info("Dispatcher Service Changed - Patching")
+			_, err = r.kubeClientset.CoreV1().Services(existingService.Namespace).Patch(ctx, existingService.Name, types.JSONPatchType, patch, metav1.PatchOptions{})
+			if err == nil {
+				controller.GetEventRecorder(ctx).Event(channel, corev1.EventTypeNormal, event.DispatcherServicePatched.String(), "Dispatcher Service Patched")
+			} else {
+				controller.GetEventRecorder(ctx).Event(channel, corev1.EventTypeWarning, event.DispatcherServicePatchFailed.String(), "Dispatcher Service Patch Failed")
+				channel.Status.MarkDispatcherFailed(event.DispatcherServicePatchFailed.String(), "Failed To Patch Dispatcher Service: %v", err)
+				return err
```
travis-minke-sap, line 175: everywhere else we create events we also log error/success - can we do the same here?

pkg/channel/distributed/controller/kafkachannel/dispatcher.go
eric-sap: But of course

pkg/channel/distributed/controller/kafkachannel/dispatcher.go
eric-sap: It's still returning nil, it's just after the "if needsUpdate" block

pkg/channel/distributed/controller/kafkachannel/dispatcher.go
eric-sap: Yes; that was a miss on my part

pkg/channel/distributed/controller/util/compare.go
```diff
@@ -0,0 +1,192 @@
...
+	"go.uber.org/zap"
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"knative.dev/pkg/apis/duck"
+)
+
+// CheckDeploymentChanged Modifies An Existing Deployment With New Fields (If Necessary)
+// Returns True If Any Modifications Were Made
+func CheckDeploymentChanged(logger *zap.Logger, existingDeployment, newDeployment *appsv1.Deployment) (*appsv1.Deployment, bool) {
+
```
travis-minke-sap, line 31: minor nit/suggestion - i would rename "existingDeployment" to "oldDeployment" in this context - the utility function doesn't know if it's "existing in k8s" or what (or shouldn't anyways ; )

eric-sap: I agree; I replaced other "existing" with "old" in this file where appropriate

pkg/channel/distributed/controller/util/compare.go
```diff
@@ -0,0 +1,192 @@
...
+	// Verify everything in the container spec aside from some particular exceptions (see "ignoreFields" below)
+	existingContainerCount := len(existingDeployment.Spec.Template.Spec.Containers)
+	if existingContainerCount == 0 {
+		// This is unlikely but if it happens, replace the entire existing deployment with a proper one
+		logger.Error("Deployment Has No Containers")
+		return newDeployment, true
+	} else if existingContainerCount > 1 {
+		logger.Warn("Deployment Has Multiple Containers; Comparing First Only")
+	}
```
travis-minke-sap, line 74: Unless there's some K8S standard guaranteeing this we need to find the right container.  Istio Injected sidecars, etc...

eric-sap: I've added a guard (findContainer) to ensure that the old container name is the same as the new once.  Istio (and, I presume, other) sidecar injectors don't modify the deployment, though; they only add containers to the pods themselves.
Other things do modify the deployment, like HorizonalPodAutoscalers (directly changes the Replicas), though, and something unknown might add more containers to it, so I think this was a good call regardless.

pkg/channel/distributed/controller/util/compare.go
```diff
@@ -0,0 +1,205 @@
...
+		if container.Name == name {
+			return &container
+		}
+	}
+	return nil
+}
+
+// CheckServiceChanged Modifies A Service With New Fields (If Necessary)
+// Returns True If Any Modifications Were Made
+func CheckServiceChanged(logger *zap.Logger, oldService, newService *corev1.Service) ([]byte, bool) {
```
aliok, line 135: Any specific reason that this function returns a patch and then the patch is applied but in deployment the updated deployment is returned and it is updated with that?

aliok: I am approving the PR. This was a question, not a problem with the PR.

aliok (approved):
/lgtm
/approve

pkg/channel/distributed/controller/util/compare.go
eric-sap: It was originally more similar to the CheckDeploymentChanged function, but the generation of the patch (and related error handling) was a collection of duplicated code for the two services that call CheckServiceChanged(), so I put the patch code in compare.go as well (I then separated it out into createJsonPatch() later for readability and testing purposes, so I suppose I could have made that a public function instead but it didn't occur to me at the time).

eric-sap: I just realized that the question might be more "why am I using a patch and not an update," which is because for some reason Kubernetes rejects service updates that contain immutable fields (such as the "ClusterIP") _even if_ the value in the field is the same as the one in the current resource.  The patches don't contain those fields, so they cause less trouble.

aliok: > even if the value in the field is the same as the one in the current resource. The patches don't contain those fields, so they cause less trouble.

Ok, got it.

Thanks for the detailed reply.

PR #561, opened 2021-04-21
[distributed kafkachannel] Distributed Dispatcher with common/consumer

travis-minke-sap:
In order to better understand our options for attempting to further consolidate some of the ConsumerGroup logic between the _consolidated_ and _distributed_ Dispatcher implementations, I have created this POC to switch the _distributed_ implementation to use the **common/consumer** logic. 

## Proposed Changes

- _distributed_ Dispatcher uses the common/consumer KafkaConsumerGroupFactory instead of it's own Sarama ConsumerGroup lifecycle management.
- _distributed_ Dispatcher's Handler now implements the common/consumer KafkaConsumerHandler interface instead of the Sarama interface.
- Moved / enhanced some _consolidated_ mocks to common/consumer/testing package.
- Fixed the _distributed_ unit tests to align with new implementation

## TODOs

- Minor nit but the common/consumer KafkaConsumerHandler method `GetConsumerGroup()` function should be renamed `GetConsumerGroupId()` to better express what it does.  Also, that function exists only for a single log message in the factory so not sure how useful it is.
- The common/consumer logic (e.g. `SetReady()`) expects to be used with the current "Prober" implementation for tracking ConsumerGroup status.  The _distributed_ Dispatcher does not make use of this so there are some loose ends.  Not sure it would be worthwhile to implement support for the "Prober" since it is expected to be replaced with a new control-protocol based implementation in the near future?
- Importantly, the error returned from the `Handle()` function is used in the Prober implementation to set ConsumerGroup status.  While this isn't really an issue in this POC it is something we need to sort out if we move forward with this idea.  The _distributed_ implementation WILL mark the offset of failed messages in order to move on and not block the queue and we wouldn't want an error in sending an event to result in marking the ConsumerGroup status as failed/error.  We could always "swallow" the error inside the `Handle()` implementation but that doesn't seem correct.  Just a difference to iron out - shouldn't be a huge deal.
- I've only performed simple vanilla testing and would need to more closely confirm no message loss on shutdown/restarts etc.


travis-minke-sap:
This is currently just a POC for discussion / review...

/hold

aliok:
@travis-minke-sap  Sorry I missed this PR.

I think this is still a valid improvement. I would be happy to see this merged.  Thanks for the work.

I like the direction of the PR, not much to add at the moment.

eric-sap (approved):
Looks good!  Thanks!

travis-minke-sap:
Switching over!

/unhold

PR #568, opened 2021-04-26
Drop --strict flag on ko

markusthoemmes:
As per title. Since https://github.com/google/ko/commit/6586a72f8a458b6b2139fd8b97d4412afab91f03 this has been the default and we recently updated `ko` in CI.

/assign @matzew @slinkydeveloper


slinkydeveloper (approved):
/lgtm

PR #572, opened 2021-04-26
[Consolidated Channel] Revert "Dispatcher readiness e2e test (#476)"

devguyio:
This reverts commit dcbe68cc1b0625fa16572097cc421ba3267e248f.

We see a lot of flakiness in consolidated channel e2e tests and this test might be the reason cause the tests aren't ready for scaling the deployment during the execution. 

We still need to investigate more though and reintroduce this test after fixing the root cause.


devguyio:
Gonna unhold when it's clear this fixes the flakes
/hold

lberk:
instead of outright reverting the change -- could we just `t.Skip()` it for now?

devguyio:
@lberk why do you think we should skip instead of reverting? are there specific reasons?

devguyio:
@lberk I had a change of heart trying to think which approach to weigh in for. I'm leaning more towards skipping to keep the test in mind and make it handy  to enable it locally whenever someone starts working on fixing the real root cause of the flakiness.

I'll close this and introduce another PR.

PR #573, opened 2021-04-26
[Consolidated Channel] Skipping subscription readiness test when scaling dispatcher

devguyio:
We see a lot of flakiness in consolidated channel e2e tests and this test might be the reason cause the tests aren't ready for scaling the deployment during the execution.

We still need to investigate more though and enable this test after fixing the root cause.

### DIFF
--- a/test/e2e/channel_subscription_ready_test.go
+++ b/test/e2e/channel_subscription_ready_test.go
@@ -27,5 +27,6 @@ import (
 )
 
 func TestChannelSubscriptionScaleReadyV1(t *testing.T) {
+	t.Skipf("Skipping test due to flakiness.")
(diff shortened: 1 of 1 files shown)


lberk:
Thanks for putting this together!

At the risk of falling into the meme of '100 line patch -- LGTM, 10 line patch -- 10 comments',
We could achieve the same result in a one line diff.  Just stuff a `t.Skip(...)` at the top of the function (first line) and you can leave the rest in there without having to mess with imports. It'll make re-enabling this test easier once we find the root cause.

Mind doing that? Thanks!

devguyio:
@lberk the linter will complain from unused imports

devguyio:
@lberk done

lberk (approved):
Lgtm

PR #583, opened 2021-04-28
Shared common config files

lionelvillard:
Fixes #

## Proposed Changes

- Share KafkaSource config files when appropriate.
-
-

**Release Note**

```release-note

```

**Docs**


config/source/common/serviceaccount.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 The Knative Authors
+# Copyright 2019 The Knative Authors
```
matzew, line 1: hrm.... ?

matzew:
Can you change the mixed up / incorrect header dates ? 

perhaps bump all to 2021 ?

config/source/common/serviceaccount.yaml
antoineco: Looks like a Git glitch. It was probably another file that got wrongly considered as "renamed".

antoineco (approved):
/lgtm

lionelvillard:
These are not new files, so the copyright should not be changed.

@antoineco thanks for approving! `/lgtm` has been deprecated in favor of GitHub approval process.

antoineco:
Good to know!

Regarding the copyright headers, in theory the date should reflect the year the code was "published", which on GitHub should correspond to a "release". In practice, I've never seen anyone apply that correctly :)

matzew (approved):
/lgtm
/approve

PR #586, opened 2021-04-29
Remove istio sidecar annotation

### LINKED ISSUE #294: make istio annotation optional
**Problem**

The kafka receiver adapter always has the istio proxy side car, currently the annotation `"sidecar.istio.io/inject"` is always set to true. This feature would make it opt in by using the annotation mentioned in the kafkasource

Example
```
apiVersion: sources.knative.dev/v1beta1
kind: KafkaSource
metadata:
  annotations:
    sidecar.istio.io/inject: "false"
  name: kafka-source-test
spec:
  bootstrapServers:
  - someserver
  consumerGroup: testgrp
  sink:
    ref:
      apiVersion: serving.knative.dev/v1
      kind: Service
      name: testsvc
  topics:
  - test
```

matzew:

Fixes #294

A bit related to an oder PR in this direction, where same has done to all components that have existed back than:
https://github.com/knative/eventing/pull/1119

At the time of that PR, there was no KafkaSource

## Proposed Changes

- Remove Istio sidecar from the kafka source
-

**Release Note**

```release-note
KafkaSource will not have Istio side-cars, it does not need Istio to function, however it will continue work if Istio is present.
```

**Docs**

### DIFF
--- a/pkg/source/reconciler/source/resources/receive_adapter.go
+++ b/pkg/source/reconciler/source/resources/receive_adapter.go
@@ -93,9 +93,6 @@ func MakeReceiveAdapter(args *ReceiveAdapterArgs) *v1.Deployment {
 			Replicas: args.Source.Spec.Consumers,
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
-					Annotations: map[string]string{
-						"sidecar.istio.io/inject": "true",
-					},
 					Labels: args.Labels,
 				},
 				Spec: corev1.PodSpec{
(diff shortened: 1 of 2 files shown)


matzew:
Alternative we could just render it as `false`, instead of `true`... Open for feedback

nak3:
> Alternative we could just render it as false, instead of true... Open for feedback

eventing Kafka does not work with istio sidecar injection at all? If it does not work, it might be worth to explicitly set `false` by default.
But LGTM for this PR as well.

matzew:
@nak3 I have never used Eventing-Kafka with Istio. I was just wondering if instead of removing we keep it at `false`, to basically opt out.

That said, I think it's legacy (see the referenced PR)

aliok (approved):
/lgtm
/approve

Looks a lot like https://github.com/knative/eventing/pull/1119

PR #587, opened 2021-04-29
[scheduler, mtsource] Initial HA support for MT Kafka source

### LINKED ISSUE #463: [scheduler,mtsource] support for HA
**Problem**
By default, the scheduler tries to allocate vpod replicas onto the same pod. When this pod becomes unavailable, the vpod also become unavailable. Not good.

Tasks:
- [ ] [design - scheduling strategies](https://docs.google.com/document/d/1vR5CVLIVvvVdnlLuS7HtQbE69sYsI9k833pk_aohtmo/edit#heading=h.u20336n3f3mt)
- [ ] implementation

aavarghese:
Fixes #463

Depends on https://github.com/knative/eventing/pull/5360

## Proposed Changes

- Controller has new env var to control "scheduler policy type",
- Stateful set adapter has new anti-affinity rules for pods to be placed on new nodes in new zone topology,
- "scheduler policy type" is an enum with two values currently MaxFillup and EvenSpread. For EvenSpread, scheduler places vreplicas for HA by spreading replicas equally across zones. Within a zone, vreplicas are filled up in each pod upto capacity. If MaxFillup strategy is chosen, scheduler places vreplicas in existing pods and fills them up before creating new pods (zones aren't considered)
- Placement type in KafkaSource status has a new type for ZoneName when EvenSpread scheduling
- HA controls are also shared with autoscaler, and state accessor
- State accessor computes the number of zones in cluster and stores a map of node name to zone info

**Release Note**

```release-note

```


config/source/multi/deployments/adapter.yaml
```diff
@@ -20,7 +20,7 @@ metadata:
   labels:
     kafka.eventing.knative.dev/release: devel
 spec:
-  replicas: 1
+  replicas: 3
```
matzew, line 23: not sure this is correct, to set that on default.
should be more done via an operator thing, deploying the bits

config/source/multi/deployments/controller.yaml
```diff
@@ -47,12 +47,20 @@ spec:
...
         - name: AUTOSCALER_REFRESH_PERIOD
-          value: '10'
+          value: '100'
 
         # The number of virtual replicas this pod can handle.
         - name: POD_CAPACITY
           value: '100'
 
+        # The number of zones in a multi-zone cluster, for spreading vreplica placement with HA
+        - name: SCHEDULE_SPREAD_ZONES
```
lionelvillard, line 57: What about `SCHEDULER_SPREAD_ZONES_COUNT`? At least `SCHEDULER` to match `AUTOSCALER` above

config/source/multi/deployments/controller.yaml
```diff
@@ -47,12 +47,20 @@ spec:
...
+          value: '3'
+
+        # The scheduling policy type for placing vreplicas on pods (a boolean value temporarily set to true for even spread across zones)
+        - name: SCHEDULE_POLICY_TYPE
+          value: 'true'
```
lionelvillard, line 62: should this be the scheduler strategy name (eg. 'even-spread`, `max??`)?

pkg/common/scheduler/statefulset/state.go
```diff
@@ -38,6 +38,8 @@ type state struct {
 
 	// Pod capacity.
 	capacity int32
+
+	schedulePolicy bool
```
lionelvillard, line 42: should be a string

pkg/common/scheduler/statefulset/scheduler.go
```diff
@@ -128,10 +159,16 @@ func (s *StatefulSetScheduler) Schedule(vpod scheduler.VPod) ([]duckv1alpha1.Pla
 		return placements, nil
 	}
 
+	//TODO: change policy type from bool to enum and send appropriate values
```
lionelvillard, line 162: ok seeing this comment now. Sounds good!

pkg/common/scheduler/statefulset/scheduler.go
```diff
@@ -128,10 +159,16 @@ func (s *StatefulSetScheduler) Schedule(vpod scheduler.VPod) ([]duckv1alpha1.Pla
...
+	if s.schedulePolicy {
+		spreadVal = int32(math.Ceil(float64(vpod.GetVReplicas()) / float64(s.numZones)))
```
lionelvillard, line 164: Can you add a comment defining what is spreadVal?

pkg/common/scheduler/statefulset/scheduler.go
```diff
@@ -160,73 +197,237 @@ func (s *StatefulSetScheduler) Schedule(vpod scheduler.VPod) ([]duckv1alpha1.Pla
 	return placements, nil
 }
 
-func (s *StatefulSetScheduler) removeReplicas(diff int32, placements []duckv1alpha1.Placement) []duckv1alpha1.Placement {
+func (s *StatefulSetScheduler) removeReplicas(diff int32, placements []duckv1alpha1.Placement, evenSpread int32) []duckv1alpha1.Placement {
```
lionelvillard, line 200: For clarity I think we should move the two schedule strategies in two different files.

pkg/common/scheduler/statefulset/scheduler.go
```diff
@@ -160,73 +197,237 @@ func (s *StatefulSetScheduler) Schedule(vpod scheduler.VPod) ([]duckv1alpha1.Pla
...
-		if diff >= placements[i].VReplicas {
-			// remove the entire placement
-			diff -= placements[i].VReplicas
-		} else {
-			newPlacements = append(newPlacements, duckv1alpha1.Placement{
-				PodName:   placements[i].PodName,
-				VReplicas: placements[i].VReplicas - diff,
-			})
-			diff = 0
+	logger := s.logger.With("removeReplicas()", zap.Int32("schedule with even spread:", evenSpread))
```
lionelvillard, line 202: I think you want `s.logger.Named("remove replicas").With(zap.Int32("evenSpread", evenSpread))`

pkg/common/scheduler/statefulset/scheduler.go
```diff
@@ -160,73 +197,237 @@ func (s *StatefulSetScheduler) Schedule(vpod scheduler.VPod) ([]duckv1alpha1.Pla
...
+
+	if evenSpread > 0 {
+		placementsByZone := getPlacementsByZoneKey(placements)
+		logger.Infow("removeReplicas()", zap.Int32("diff:", diff), zap.Any("placementsByZone:", placementsByZone))
```
lionelvillard, line 206: no need to repeat `"removeReplicas()` here. And I would remove `:` to be consistent.

pkg/common/scheduler/statefulset/scheduler.go
```diff
@@ -160,73 +197,237 @@ func (s *StatefulSetScheduler) Schedule(vpod scheduler.VPod) ([]duckv1alpha1.Pla
...
+					if deallocation < placement.VReplicas {
+						newPlacements = append(newPlacements, duckv1alpha1.Placement{
+							PodName:   placement.PodName,
+							ZoneName:  placement.ZoneName,
+							VReplicas: placement.VReplicas - deallocation,
+						})
+					}
+					diff -= deallocation
+					totalInZone -= deallocation
+					logger.Infow("removeReplicas()", zap.Int32("diff", diff), zap.Int32("totalInZone", totalInZone))
```
lionelvillard, line 228: same here

pkg/common/scheduler/statefulset/scheduler.go
```diff
@@ -160,73 +197,237 @@ func (s *StatefulSetScheduler) Schedule(vpod scheduler.VPod) ([]duckv1alpha1.Pla
...
-			diff -= allocation
-			state.SetFree(ordinal, f-allocation)
-		} else {
-			newPlacements = append(newPlacements, placements[i])
+		if diff > 0 {
+			for ordinal := int32(0); ordinal < s.replicas; ordinal++ {
+				f := state.Free(ordinal)
+				logger.Infow("Trying to schedule on new pods", zap.Int32("ordinal", ordinal), zap.Int32("free", f))
+				if f > 0 { //here it is possible to hit pods that are in existing placements
+					podName, zoneName, err := s.getZoneNameFromPodOrdinal(ordinal)
```
lionelvillard, line 300: Is it possible to have this info as part of the state? I'm worried about race condition here.

lionelvillard: also it's not clear what to do when there is an error. It shouldn't be `continue`

pkg/source/reconciler/mtsource/controller.go
```diff
@@ -40,6 +40,8 @@ import (
 type envConfig struct {
 	SchedulerRefreshPeriod int64 `envconfig:"AUTOSCALER_REFRESH_PERIOD" required:"true"`
 	PodCapacity            int32 `envconfig:"POD_CAPACITY" required:"true"`
+	NumZones               int32 `envconfig:"SCHEDULE_SPREAD_ZONES" required:"true"`
```
lionelvillard, line 43: Is this strictly needed? I assume you can compute the number of zones by getting all nodes, right?

pkg/common/scheduler/statefulset/scheduler.go
aavarghese: Ok. Sounds good. Will add. Most of these log lines may be cleaned out after testing concludes.

pkg/common/scheduler/statefulset/scheduler.go
aavarghese: yeah, I see ```s.podClient.Get(context.Background(), podName, metav1.GetOptions{})``` erroring here and there with this error ```"error":"pods \"kafkasource-mt-adapter-4\" not found",``` and then most times during the next reconcile, it passes. Finding pod fails which means we dont get node info and then the zone info.

aavarghese: Storing this in state is a good idea. I will try that too.

pkg/common/scheduler/statefulset/scheduler.go
aavarghese: Separated into different methods at the moment..

aavarghese:
Testing continues on my end....

Few more things remaining to solve in this PR, (or in separate future ones):
-  Implementing HA across pods within a zone (# of pods is not static unlike # of zones, so spreading equally may also mean rebalancing) [TBD]
- When multiple kafkasources with large # of consumers are installed, a race condition creates pods to have more vreplicas than capacity (non-HA problem too?) Error msg: ```{"level":"info","ts":"2021-05-03T20:53:31.852Z","logger":"kafka-controller","caller":"statefulset/state.go:126","msg":"pod is overcommitted","podName":"kafkasource-mt-adapter-5"}```
- Using kubeclient to get pod info doesn't always work. Some times error is returned. Trying again may cause it to succeed. https://github.com/knative-sandbox/eventing-kafka/pull/587#discussion_r623980525 Not sure why this happens from researching on kubernetes issues opened.
- Currently, number of zones and node to zone information is computed and stored in State accessor when state() is called. We need a watch for zone/node changes to recompute. 
cc: @lionelvillard

config/source/multi/deployments/adapter.yaml
```diff
@@ -80,3 +80,18 @@ spec:
           containerPort: 8008
 
       terminationGracePeriodSeconds: 10
+      affinity:
```
matzew, line 83: do we set this anywhere in Knative already?

Is that something that we might have the user/admin have control on? Or allow tweaking?

lionelvillard: [Yes (single zone only)](https://github.com/knative/eventing/blob/main/config/core/deployments/controller.yaml#L33) and yes, though it really depends on the available scheduling strategies.

config/source/multi/deployments/controller.yaml
```diff
@@ -47,12 +47,16 @@ spec:
...
-          value: '10'
+          value: '100'
 
         # The number of virtual replicas this pod can handle.
         - name: POD_CAPACITY
           value: '100'
 
+        # The scheduling policy type for placing vreplicas on pods (see type SchedulerPolicyType for enum list)
+        - name: SCHEDULER_POLICY_TYPE
+          value: '1'
```
lionelvillard, line 58: I'd like the value to be more descriptive (eg. `maxfillup`). We can do this in another PR.

pkg/common/scheduler/statefulset/scheduler.go
```diff
@@ -37,14 +42,28 @@ import (
 	"knative.dev/eventing-kafka/pkg/common/scheduler"
 )
 
+type SchedulerPolicyType int
+
+const (
+	// MaxFillup policy type adds replicas to existing pods to fill them up before adding to new pods
+	MaxFillup SchedulerPolicyType = iota
+	// EvenSpread policy type spreads replicas uniformly across zones in different regions for HA, and within each zone, fills up replicas in existing and new pods
```
lionelvillard, line 50: I would generalize this statement by saying vreplicas are spread across [failure-domains such as regions, zones, nodes](https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/).

pkg/common/scheduler/statefulset/state.go
```diff
@@ -112,7 +130,31 @@ func (s *stateBuilder) State() (*state, error) {
 			}
 		}
 	}
-	return &state{free: free, lastOrdinal: last, capacity: s.capacity}, nil
+
+	if s.schedulerPolicy == EvenSpread {
+		//TODO: need a node watch to see if # nodes/ # zones have gone up or down
+		nodes, err := kubeclient.Get(s.ctx).CoreV1().Nodes().List(s.ctx, metav1.ListOptions{})
```
lionelvillard, line 136: We should use a NodeLister here to hit the cache.

aavarghese:
Test files in this PR depend on https://github.com/knative/eventing/pull/5360

pkg/common/scheduler/statefulset/state.go
aavarghese: Scheduler and state now using `node lister` and `pod lister` now to get zone info from all nodes, and also to get nodename from pod to compute its zone. 

Only piece remaining as part of this refactoring are the three test files (state_test, autoscaler_test and scheduler_test) that do create dummy nodes, pods etc and need to pass a node lister instance from the test context to the state accessor. Unsure of how to use the new method [nodelister()](https://github.com/knative-sandbox/eventing-kafka/pull/587/files#diff-376bd4cf279c82c42a8d2d1e7b070dd370c89f8148948b230822e8ddd94b8dcfR216) in all three of my testcase files  without a [Listers](https://github.com/knative/eventing/blob/main/pkg/reconciler/testing/v1/listers.go#L71) object that may need some setup with reconciler testing methods instead of what we have now.... @lionelvillard will need to check with you on this.

aavarghese: Done. Changes in latest test files

lionelvillard:
Let's get this in so we can move forward with the issues raised during the review.

Thanks @aavarghese this is amazing work!

lionelvillard (approved):

Issue #588, opened 2021-04-29
ResetOffset Command


travis-minke-sap:
**Problem**
Users of Kafka resources need the ability to reposition the Offsets of a particular ConsumerGroup in order to recover from unexpected downtime. The associated Feature Track [document](https://docs.google.com/document/d/1sbvxEWTB7l6eADk4WyddHLEtH1q9cv7rlSbiBivHGMs/edit?usp=sharing&resourcekey=0-Ngyt3ddYC1ltSP1g59eYXQ) contains a detailed description of the feature. This Issue is the top-level coordination point for the various PRs expected to be required to implement this feature.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)**
This new feature is targeted at the Cluster Admin and/or Event Subscriber personas.

**Exit Criteria**
Users are provided a mechanism by which they can reposition the Offsets of a particular ConsumerGroup based on a specified timestamp.

**Time Estimate (optional):**
Many... This is a new feature comprising a new CRD, Controller, etc which is being implemented in a re-usable fashion so that multiple Kafka Channel/Broker implementations can make use of it.

**Tasks**
- [X] Create ResetOffset CRD and generate informers / listers / etc. (PR #589)
- [X] Create ResetOffset Controller with initial reconciliation framework. (partial implementation) (PR #626)
- [X] Define stop/start ConsumerGroup Messages (PRs #724, #725)
- [X] Map ResetOffset Ref to dataplane and send control-protocol messags (PR #725)
- [X] Common Consumer Stop/Start ConsumerGroups in response to control-protocol messaging. (PR #724)
- [X] Add Sync/Locking to Consumer Manager & Controller (PR #741 and #742 and #746)
- [X] Add ResetOffset Admission/Validation support to KafkaChannel Webhook (#753)
- [X] Verify / handle Subscription Status while consumergroup stopped. (#755)
- [ ] ~~Deal with duplicate reconciliation race condition (K8S Status update)~~ NO LONGER OCCURRING
- [X] Create ResetOffset Controller common RBAC (#750)
- [X] Add ResetOffset Controller to KafkaChannel Controllers (#761)
- [X] New Documentation on ResetOffset usage in KafkaChannels (#745, and [docs 4012](https://github.com/knative/docs/pull/4012))
- [X] Create e2e tests (#781)

ssandanshi:
[...]

PR #589, opened 2021-04-29
ResetOffset API & CRD

travis-minke-sap:
Sub-Task of Issue #588 to create the ResetOffset API, CRD, and generated source

## Proposed Changes
- Defined the CRD
- Created the API
- Generated DeepCopy, Client, Informers, Listers, etc

**Release Note**

**Docs**

**Notes**
- This is the first step in the ResetOffset implementation.  The new API / CRD are not used anywhere and should not impact any existing implementations.
- This implementation treats the Status slightly different from the FeatureTrack document.  The same approach is still in place, but the names have been altered to fit with the Knative `NewBatchConditionSet()` and `Succeeded` state.  I can update the document to match if necessary once this PR is approved.

A sample ResetOffset CRD can be created as...
```yaml
apiVersion: kafka.eventing.knative.dev/v1alpha1
kind: ResetOffset
metadata:
  name: sample-reset-offset-1
  namespace: tenant1
spec:
  offset: earliest
  ref:
    apiVersion: messaging.knative.dev/v1beta1
    kind: Subscription
    namespace: tenant1
    name: sample-kafka-subscription-1
```
...and an example status could be...
```bash
> kubectl get resetoffset -n tenant1 sample-reset-offset-1
NAME                    SUCCEEDED   REASON                      AGE
sample-reset-offset-1   False       FailedToStopConsumerGroup   15s

> kubectl get resetoffset -n tenant1 sample-reset-offset-1 -o wide
NAME                    SUCCEEDED   REASON                      REF NAMESPACE   REF NAME                      AGE
sample-reset-offset-1   False       FailedToStopConsumerGroup   tenant1         sample-kafka-subscription-1   18
```
...with detailed status of...
```bash
> kubectl get resetoffset -n tenant1 sample-reset-offset-1 -o yaml
apiVersion: kafka.eventing.knative.dev/v1alpha1
kind: ResetOffset
metadata:
  creationTimestamp: "2021-04-29T18:59:11Z"
  finalizers:
  - resetoffsets.kafka.eventing.knative.dev
  generation: 1
  managedFields:
    ...
  name: sample-reset-offset-1
  namespace: tenant1
  resourceVersion: "50428288"
  selfLink: /apis/kafka.eventing.knative.dev/v1alpha1/namespaces/tenant1/resetoffsets/sample-reset-offset-1
  uid: f2b8c647-c717-4b68-8dec-3baa8cff716e
spec:
  offset: earliest
  ref:
    apiVersion: messaging.knative.dev/v1beta1
    kind: Subscription
    name: sample-kafka-subscription-1
    namespace: tenant1
status:
  annotations:
    group: my-group
    partitions: '[{"partition":0,"oldOffset":1,"newOffset":2},{"partition":1,"oldOffset":2,"newOffset":3},{"partition":2,"oldOffset":3,"newOffset":4},{"partition":3,"oldOffset":4,"newOffset":5}]'
    topic: my-topic
  conditions:
  - lastTransitionTime: "2021-04-29T18:59:11Z"
    status: "True"
    type: 1-ResetInitiated
  - lastTransitionTime: "2021-04-29T18:59:11Z"
    status: "True"
    type: 2-OffsetsCalculated
  - lastTransitionTime: "2021-04-29T18:59:11Z"
    message: failed to stop consumergroup 123
    reason: FailedToStopConsumerGroup
    status: "False"
    type: 3-ConsumerGroupsStopped
  - lastTransitionTime: "2021-04-29T18:59:11Z"
    status: Unknown
    type: 4-OffsetsUpdated
  - lastTransitionTime: "2021-04-29T18:59:11Z"
    status: Unknown
    type: 5-ConsumerGroupsStarted
  - lastTransitionTime: "2021-04-29T18:59:11Z"
    message: failed to stop consumergroup 123
    reason: FailedToStopConsumerGroup
    status: "False"
    type: Succeeded
  observedGeneration: 1
```

/hold


pkg/apis/kafka/v1alpha1/reset_offset_types_test.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/stretchr/testify/assert"
+	duckv1 "knative.dev/pkg/apis/duck/v1"
+)
+
+func TestResetOffset_GetGroupVersionKind(t *testing.T) {
+	resetOffset := ResetOffset{}
+	gvk := resetOffset.GetGroupVersionKind()
+	if gvk.Kind != "ResetOffset" {
+		t.Errorf("Should be 'ResetOffset'.")
+	}
```
eric-sap, line 33: This also seems odd vs. using assert.Equal

eric-sap (commented):
Some trivial stuff

pkg/apis/kafka/v1alpha1/reset_offset_annotations_test.go
```diff
@@ -0,0 +1,118 @@
...
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			ros := &ResetOffsetStatus{}
+			if len(test.topic) > 0 {
+				ros.SetTopicAnnotation(test.topic)
+			}
+			got := ros.GetTopicAnnotation()
+			if diff := cmp.Diff(test.topic, got); diff != "" {
+				t.Errorf("unexpected conditions (-want, +got) = %v", diff)
+			}
```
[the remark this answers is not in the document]
travis-minke-sap: Yeah... I was just following the patterns in the existing api tests for consistency.  There is only one other use of testify in the apis/ that was subsequently added after the initial implementation.  The testify dependency shouldn't ever be a problem for anyone coding against the apis/ since they wouldn't pull in the _test.go files.

@matzew or @slinkydeveloper - Any opinions on using/not-using the testify assertions in the apis/ packages?  Concerns about consistency one way or another or just not a big deal either way?

pkg/apis/kafka/v1alpha1/reset_offset_types_test.go
travis-minke-sap: Same response as above - will either leave or refactor everything out depending on feedback - closing this discussion as duplicate...

travis-minke-sap:
CLA broke because I used the github "commit suggestions" feature - anyone know how to recover from that ?

slinkydeveloper:
> CLA broke because I used the github "commit suggestions" feature - anyone know how to recover from that ?

Force push changing the commit message author and adding the sign off string is the easiest way :)

pkg/apis/kafka/v1alpha1/reset_offset_lifecycle.go
```diff
@@ -0,0 +1,135 @@
...
+	ResetOffsetConditionResetInitiated apis.ConditionType = "1-ResetInitiated"
+
+	// ResetOffsetConditionOffsetsCalculated has status True when the individual offsets of each
+	// Partition in the Topic have been calculated from the "offset" indicator in the Spec and
+	// have been verified as valid relative to the persistence window.
+	ResetOffsetConditionOffsetsCalculated apis.ConditionType = "2-OffsetsCalculated"
+
+	// ResetOffsetConditionConsumerGroupsStopped has status True when all of the ConsumerGroups
+	// associated with the referenced object (Subscription, Trigger, etc.) have been stopped.
+	ResetOffsetConditionConsumerGroupsStopped apis.ConditionType = "3-ConsumerGroupsStopped"
```
slinkydeveloper, line 50: Having the numbers in the condition string doesn't seem very kubernet-ish

pkg/apis/kafka/v1alpha1/reset_offset_annotations.go
```diff
@@ -0,0 +1,87 @@
...
+	ResetOffsetAnnotationGroup      = "group"
+	ResetOffsetAnnotationPartitions = "partitions"
+)
+
+// OffsetMapping is a simple struct representing a single Kafka
+// Topic/Partition Offset value before and after repositioning
+type OffsetMapping struct {
+	Partition int32 `json:"partition"`
+	OldOffset int64 `json:"oldOffset"`
+	NewOffset int64 `json:"newOffset"`
```
slinkydeveloper, line 34: I'm sorry, I don't understand the reason of this. Is it for some sort of "signaling" the result? If that's the case, why it's not in the `ResetOffsetStatus`?

pkg/apis/kafka/v1alpha1/reset_offset_types.go
```diff
@@ -0,0 +1,133 @@
...
+// ResetOffsetSpec defines the specification for a ResetOffset.
+type ResetOffsetSpec struct {
+
+	// Offset is a String representing the desired offset position to which all partitions
+	// will be reset.  Supported values include "earliest", "latest", or a valid date / time
+	// string in the time.RFC3339 format. The "earliest" and "latest" values indicate the
+	// beginning and end, respectively, of the persistence window of the Topic.  There is no
+	// default value, and invalid values will result in the ResetOffset operation being
+	// rejected as failed.
+	Offset string `json:"offset"`
```
slinkydeveloper, line 78: I think the name here is straightforward, and fits the use case, so I like it.

However, for the sake of future-proof, I cannot think how to name an eventual future "numeric" offset change... How would you do that @travis-minke-sap?

Did you thought to something like:

```
offset:
  time: [earliest|latest|time string]
```

```
offset:
  value: 10
```

slinkydeveloper (commented):
I like it, left just a comment about the annotations and a consideration about the offset field name

pkg/apis/kafka/v1alpha1/reset_offset_lifecycle.go
travis-minke-sap: Yeah I agree it is a bit odd, the reason for that was trying to convey the "order" of the algorithm to the user.  I can remove the numbering but then the status fields are just alphabetical.  For a normal aggregated status case that's not so bad but for this `NewBatchConditionSet` usage it felt wrong.   Still not a big deal - happy to remove the numbers if you still prefer it that way?

pkg/apis/kafka/v1alpha1/reset_offset_types.go
travis-minke-sap: I was just thinking we would attempt to parse the string value as an int64 and see if it was successful (assuming the other checks failed to parse).   I'm fine with making it more explicit though (via the field typing/separation as you suggest).  This will just mean some additional "overhead" of another json struct, etc. which is fine.  I'll make the change.

pkg/apis/kafka/v1alpha1/reset_offset_annotations.go
travis-minke-sap: The struct is used to track the old/new offsets and is written to the duck Status Annotations which seems to be their intended use right...

```
        // Annotations is additional Status fields for the Resource to save some
	// additional State as well as convey more information to the user. This is
	// roughly akin to Annotations on any k8s resource, just the reconciler conveying
	// richer information outwards.
	Annotations map[string]string `json:"annotations,omitempty"`
```

I think this makes sense vs trying to inline this data in the Status block itself.  I can rework it if I've misunderstood the intent though?

travis-minke-sap: As to the "why"... I think knowing the old/new offsets is really useful historical information for the end user to save off and track when they do this kind of thing (recovering from downtime, etc).   At a minimum it doesn't hurt anything to expose it ; )

pkg/apis/kafka/v1alpha1/reset_offset_lifecycle.go
slinkydeveloper: How kube jobs deal with it?

pkg/apis/kafka/v1alpha1/reset_offset_annotations.go
slinkydeveloper: Yeah I think this definitely goes in the CRD status, not inside the conditions but just as a field of the status. `Annotations` are a bit hacky for that use case IMO

pkg/apis/kafka/v1alpha1/reset_offset_types.go
travis-minke-sap: Also - in your example it wouldn't be a simple "value" but would instead have to be an array/map of values for each Partition, which is an even better reason to do as you suggested ; )

slinkydeveloper: > This will just mean some additional "overhead" of another json struct, etc. which is fine. I'll make the change.

For the sake of forward compatibility with future additions to the API, I think that's a reasonable overhead. Now just add the `time` field (or just choose another name you prefer), then we'll reiterate on it when we want to further expand the PR.

pkg/apis/kafka/v1alpha1/reset_offset_types.go
```diff
@@ -87,19 +87,26 @@ type ResetOffsetSpec struct {
 	Ref duckv1.KReference `json:"ref"`
 }
 
+// OffsetIndicator defines the intended values to move the offsets to.
+// Note: This simple wrapper might seem unnecessary, but is provided to allow future extension
+//       in order to support specifying explicit offset (int64) values for each Partition.
+type OffsetIndicator struct {
```
slinkydeveloper, line 93: A better name is `OffsetSpec` IMO

pkg/apis/kafka/v1alpha1/reset_offset_annotations_test.go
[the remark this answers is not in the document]
slinkydeveloper: I'm +1000 on testify, manual assertions are just evil

pkg/apis/kafka/v1alpha1/reset_offset_annotations.go
travis-minke-sap: It IS inside the Status section as Annotations (not inside Conditions)...
```yaml
  status:
    annotations:
      group: my-group
      partitions: '[{"partition":0,"oldOffset":1,"newOffset":2},{"partition":1,"oldOffset":2,"newOffset":3},{"partition":2,"oldOffset":3,"newOffset":4},{"partition":3,"oldOffset":4,"newOffset":5}]'
      topic: my-topic
    conditions:
    - lastTransitionTime: "2021-05-03T16:55:21Z"
      status: Unknown
      type: ConsumerGroupsStarted
```
...which appears to be the intended use of the duck/v1 Status type?   I think if we are not going to use that type in this use case I would argue it should be removed from the duck type altogether because what else would it possibly be intended for?

pkg/apis/kafka/v1alpha1/reset_offset_types.go
travis-minke-sap: ok - i can rename it

pkg/apis/kafka/v1alpha1/reset_offset_lifecycle.go
travis-minke-sap: Looks like they have an array of conditions that I suppose could be considered "ordered".  I've removed the numbering and I think we can just start with that.  If it is a problem later (during initial development/testing/usage of ResetOffset) we can look at enhancements (new knative ConditionSet type or something) before final release.

pkg/apis/kafka/v1alpha1/reset_offset_annotations_test.go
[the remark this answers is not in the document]
travis-minke-sap: ok - good to hear - I agree - just a lot of the knative legacy tests are using -got +want and cmp.Diff, etc. so I wasn't sure.

pkg/apis/kafka/v1alpha1/reset_offset_annotations.go
slinkydeveloper: > which appears to be the intended use of the duck/v1 Status type?

TBH, I don't know what's the best solution here... Maybe should we mimic what job does?

```
type JobStatus struct {
	Conditions []JobCondition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,1,rep,name=conditions"`

	StartTime *metav1.Time `json:"startTime,omitempty" protobuf:"bytes,2,opt,name=startTime"`

	CompletionTime *metav1.Time `json:"completionTime,omitempty" protobuf:"bytes,3,opt,name=completionTime"`

	Active int32 `json:"active,omitempty" protobuf:"varint,4,opt,name=active"`

	Succeeded int32 `json:"succeeded,omitempty" protobuf:"varint,5,opt,name=succeeded"`

	Failed int32 `json:"failed,omitempty" protobuf:"varint,6,opt,name=failed"`
}
```

`Conditions` is the same as our conditions stuff inside `duckv1.Status`, while the rest of the fields, to me, sounds like what you're pushing inside annotations. So I fail to see the reason why we should have that information inside `annotations`.

We can't remove `duckv1.Status` though, because it contains `ObservedGeneration`. That field is necessary to make our reconcilers work. It also contains `Conditions`, which is something we need too :)

matzew: The question is... is that `optional` ... or more something "real":
https://github.com/knative/pkg/blob/main/apis/duck/v1/status_types.go#L50-L54

I could live with both... pumping/stashing these on the `Annotations`, or more explicit :man_shrugging:

travis-minke-sap: From reading the duckv1 Status.Annotations documentation I think we are using that field exactly as intended and don't see a problem.  I can move the fields to be explicit top level members of `ResetOffsetStatus` if there is consensus but... I would then argue that the duckv1 Status.Annotations field is worthless.   What kind of information WOULD you put there if not this?

travis-minke-sap: I guess the argument is that the Status.Annotations field is not "official" enough or something?

travis-minke-sap: I will make the change (moving to top level status fields) - I was just trying to use the duck types as I interpreted the intent from the docs.

slinkydeveloper: > I would then argue that the duckv1 Status.Annotations field is worthless. What kind of information WOULD you put there if not this?

TBH I have no strong argument about that and I frankly never used `Status.Annotation` for something, nor I'm aware of any API eventing using it. I won't argue its usefulness or not, because I don't have enough context, but keeping things in the `ResetOffsetStatus` struct IMO is better and you can have nested informations without serializing to json (like you do for partitions). If you look at how kafka source status and kafka channel status show some of their status metadata, they don't use annotations.

travis-minke-sap: sure, fair enough - most of the discussion was just confusion on my part as to what was being requested and what the concern was.  thx

travis-minke-sap: OK, so I refactored and it now looks like this...
```yaml
status:
  conditions:
  - lastTransitionTime: "2021-05-05T20:37:09Z"
    message: failed to start consumergroup 123
    reason: FailedToStartConsumerGroup
    status: "False"
    type: ConsumerGroupsStarted
  - lastTransitionTime: "2021-05-05T20:37:09Z"
    status: "True"
    type: ConsumerGroupsStopped
  - lastTransitionTime: "2021-05-05T20:37:09Z"
    status: "True"
    type: OffsetsUpdated
  - lastTransitionTime: "2021-05-05T20:37:09Z"
    status: "True"
    type: ResetInitiated
  - lastTransitionTime: "2021-05-05T20:37:09Z"
    message: failed to start consumergroup 123
    reason: FailedToStartConsumerGroup
    status: "False"
    type: Succeeded
  group: my-group
  observedGeneration: 1
  partitions:
  - newOffset: 2
    oldOffset: 1
    partition: 0
  - newOffset: 3
    oldOffset: 2
    partition: 1
  topic: my-topic
```

matzew: thanks for the refactors, @travis-minke-sap 

/lgtm

slinkydeveloper (approved):

travis-minke-sap:
/unhold

Thanks for all the reviews @slinkydeveloper and @matzew and for bearing with me on the annotations discussion/confusion ; )

PR #591, opened 2021-04-29
Copy volumes as well as container when creating deployment updates

### LINKED ISSUE #590: Distributed deployments aren't being updated when the volumes change
**Describe the bug**
The combination of code for issues #455 and #502 have resulted in an upgrade problem where the deployments for the distributed channel aren't updated properly from old versions because the volumeMounts, which are part of the container, are being changed but the volumes, which are outside of the container list, are not.  Volume changes should be included in the set of "allowable updates" along with the containers.

 **Expected behavior**
Updates to the receiver and dispatcher pods should work even if the volumes are changed.

 **To Reproduce**
Create a receiver deployment without any volumeMounts or volumes, and then start the current controller.  It will attempt to reconcile the deployments by upgrading them but not include the volumes, resulting in errors in the log:
```
Deployment.apps \"kafka-cluster-32603413-receiver\" is invalid: spec.template.spec.containers[0].volumeMounts[0].name: Not found: \"config-kafka\"
```

 **Knative release version**
0.22.1
 
 **Additional context**
I believe this will be a small fix to pkg/channel/distributed/controller/util/compare.go to include the Volumes from the spec as well as the Container.

eric-sap:
Fixes #590

## Proposed Changes

- 🐛 Volumes were not included in the list of fields that deployment updates should be able to change in a rollout-restart fashion; added them to the CheckDeploymentChanged function (but only if something in the container also changed -- in the case of the bug that prompted this fix, it was the addition of the "config-kafka" volumeMount in the container that demanded the change)

### DIFF
--- a/pkg/channel/distributed/controller/util/compare.go
+++ b/pkg/channel/distributed/controller/util/compare.go
@@ -116,6 +116,7 @@ func CheckDeploymentChanged(logger *zap.Logger, oldDeployment, newDeployment *ap
 	}
 	if !containersEqual {
 		updatedDeployment.Spec.Template.Spec.Containers[0] = *newContainer
+		updatedDeployment.Spec.Template.Spec.Volumes = newDeployment.Spec.Template.Spec.Volumes
 	}
 	return updatedDeployment, true
 }


matzew (commented):
/ok-to-test
/lgtm

aliok (approved):
/approve
/lgtm

PR #603, opened 2021-05-06
[distributed kafkachannel] Refactor remove secret watcher

### LINKED ISSUE #602: Distributed channel no longer needs to watch secrets
**Describe the bug**
Originally, the distributed channel had a requirement to support multiple Kafka auth secrets, which were dynamically watched and resulted in the creation and deletion of Receiver services and deployments based on those credentials.  The use case for this has been deprecated but the kafkasecret controller remains in the controller code.

 **Expected behavior**

- Secret watching should be removed from the controller - modifying the (single) secret should adjust the behavior of existing and future dispatcher and receiver deployments (as it does now), but deletion and creation of secrets should not have any immediate effect on resources.
- When a kafkachannel is reconciled, a missing secret should continue to cause the single reconciler service and deployment to be removed, but this should now also return a reconciliation error (as the single secret is mandatory).  Conversely, if the secret is present but the Receiver resources are missing, they should be created.
- The default secret should no longer have the "eventing-kafka.knative.dev/kafka-secret" label as there is no need to locate them in this manner.
- The secret should be located via new entries in the config-kafka configmap under the eventing-kafka.kafka field, named 'authSecretName' and 'authSecretNamespace' (to be consistent with similar consolidated channel behavior), with defaults of 'kafka-cluster' and 'knative-eventing' respectively.
- The Receiver service and deployment should have a constant name ("kafka-cluster-receiver") instead of the current hash-based name

 **To Reproduce**
Create multiple Kafka secrets and observe that multiple Receiver deployments and services are created.

 **Knative release version**
0.22.1

 **Additional context**
If customers are using the original multi-Kafka ability, this cleanup will break that, so a hefty release note would be a good plan

eric-sap:
Fixes #602 

## Proposed Changes

- 🗑️  Secrets no longer watched by eventing-kafka-channel-controller
- 🧽  Receiver deployment and service reconciliation logic moved to kafkachannel reconciler
- 🗑️ "eventing-kafka.knative.dev/kafka-secret" label removed from default Kafka secret and readmes
- 🎁  Added 'authSecretName' and 'authSecretNamespace' to EKKafkaConfig with defaults (kafka-cluster and knative-eventing)
- 🧽  Extensive rework of pkg/channel/distributed/controller/kafkachannel/reconciler_test.go in an attempt to keep the mountain of TableRow structs in TestReconcile() from being mentally overwhelming

Note:  The ability to specify separate brokers per secret had already been removed some time ago (#370)

I have created issue #608 to address the need for migration of existing KafkaChannel services to use a new externalName if the authSecretName is changed (and a new Receiver created because of it).

**Release Note**

```release-note
- action required: This update WILL BREAK any distributed-channel installations that are using an adminType of 'azure' and the old ability to specify multiple Kafka auth configurations in separate secrets.  This ability is no longer supported.
- If you are using the distributed channel and NOT using the default secret name of 'kafka-cluster' you MUST put your secret name into the config-kafka configmap (eventing-kafka.kafka.authSecretName).  If you do this before the upgrade, there will be no issues.  If you forget and do it after the upgrade, you will need to manually clean up (delete) the kafka-cluster-32603413-receiver service and deployment that will be created due to the default.
- Changing the authSecretName in the configmap of a live system is not recommended, as there is not yet a mechanism for migrating existing KafkaChannel services to point to the new Receiver
```


matzew:
@eric-sap I think the release notes should reflect that's on the distributed impl of the channel (speaking about the ANY :smile_cat: )

docs/kind_kafka_cluster.md
```diff
@@ -68,7 +68,7 @@ data:
 
 ```
 Specify Strimzi Kafka Version (See... https://github.com/strimzi/strimzi-kafka-operator/releases)
-export STRIMZI_VERSION=0.18.0
+export STRIMZI_VERSION=0.22.1
```
matzew, line 71: you might run into the following with that version:
```
The CustomResourceDefinition "kafkas.kafka.strimzi.io" is invalid: metadata.annotations: Too long: must have at most 262144 bytes
```

see here https://github.com/matzew/kn-box/commit/33d81c2d0964360f4cd5e730eadad8673bb32a1e#diff-2c73b452095ee223ceb2ca2dedbb5a6d59e2e2e32140d5bc71fdd762c96d31f7

pkg/channel/distributed/controller/kafkachannel/controller.go
```diff
@@ -20,22 +20,21 @@ import (
 	"context"
 	"sync"
 
+	"knative.dev/eventing-kafka/pkg/channel/distributed/common/config"
+
```
aliok, line 24: nit, can you move this down?

pkg/channel/distributed/controller/kafkachannel/controller.go
```diff
@@ -81,36 +80,25 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
...
-				SaslType: kafkaSaslType,
-			},
-		}
+	// Extract The Relevant Data From The Kafka Secret And Create A Kafka Auth Config From The
+	// Current Credentials (Secret Data Takes Precedence Over ConfigMap)
+	kafkaAuthCfg, err := config.GetAuthConfigFromKubernetes(ctx, configuration.Kafka.AuthSecretName, configuration.Kafka.AuthSecretNamespace)
+	if err != nil {
+		logger.Fatal("Failed To Load Kafka Auth From Secret", zap.Error(err),
+			zap.String("AuthSecretName", configuration.Kafka.AuthSecretName),
+			zap.String("AuthSecretNamespace", configuration.Kafka.AuthSecretNamespace))
```
matzew, line 94: the namespace and name "flexibility" works for you now?

pkg/channel/distributed/controller/kafkachannel/reconciler.go
```diff
@@ -206,17 +214,37 @@ func (r *Reconciler) reconcile(ctx context.Context, channel *kafkav1beta1.KafkaC
...
+
+		// Not having an auth secret for the Receiver is a problem
+		channel.Status.MarkConfigFailed(event.KafkaSecretReconciled.String(), "No Kafka Secret For KafkaChannel")
+		return fmt.Errorf(constants.ReconciliationFailedError)
+	} else if err != nil {
+		logging.FromContext(ctx).Error("Error reading receiver secret", zap.Error(err))
+		return fmt.Errorf(constants.ReconciliationFailedError)
+	}
+
+	receiverError := r.reconcileReceiver(ctx, channel, secret, true)
```
aliok, line 241: `r.reconcileReceiver` is called 13 lines above with `exists=false` and I see this call is with `exists=true`. But do we need to call it again here?

aliok:
Thanks for this PR 😍 
I love that channel implementations are getting closer and closer :)

I added 2 comments (one is a nit) and other than those it looks great to me.

docs/kind_kafka_cluster.md
eric-sap: I was naïvely pulling sample versions along as I found them in the readme; would 0.21.0 be a better example to leave here for now?

pkg/channel/distributed/controller/kafkachannel/controller.go
eric-sap: Sure, as long as the go formatting bot doesn't mind :)

pkg/channel/distributed/controller/kafkachannel/controller.go
eric-sap: We talked about it and decided that if there's only one secret and only one receiver, there's zero reason to name the receiver after the secret (which was the only reason this option was going to cause trouble).

Although, an upgrade will probably leave an old-named receiver orphaned the way it is; that probably needs its own release note.

eric-sap: Or the reconciler could go through and look for secrets that have the old label and delete any deployments/services that match the old hash-name .. that might be a better idea than dumping that responsibility on the admin.  Thoughts?

pkg/channel/distributed/controller/kafkachannel/reconciler.go
eric-sap: It's not really being called again here; if the "exists=false" path is taken, an early error-return from the function is guaranteed.
The purpose of the earlier call is to allow the reconciler to remove the receiver if the secret is missing - This is the original behavior of the multi-secret reconciler, though, and it actually may not even be important to delete the single receiver deployment/service if the secret is missing.  That would simplify this code quite a bit but would leave open the possibility of running receiver deployments that don't have a secret (which would lead to errors if that receiver needs to reconfigure itself).

aliok:
/lgtm

Leaving the approval to Travis/Matthias

pkg/channel/distributed/controller/kafkachannel/controller.go
travis-minke-sap: At a minimum we need a release note - it would be bad to have two receiver pods running.

config/channel/distributed/README.md
```diff
@@ -82,17 +82,16 @@ Example value for a multi-broker Kafka (must be base64 encoded):
...
-the required authentication with the Kafka cluster.
+`knative-eventing` namespace which is by default named `kafka-cluster` (the
+kafka.authSecretName and/or kafka.authSecretNamespace fields in the
+[eventing-kafka-configmap.yaml](300-eventing-kafka-configmap.yaml) may be
+changed if another secret location is desired). The
+[kakfa-secret.yaml](300-kafka-secret.yaml) is included in the config directory,
+but must be modified to hold real values. The values from this file will
+override the username/password in the
+[eventing-kafka-configmap.yaml](300-eventing-kafka-configmap.yaml).
+It is also expected that the `Config.Net.SASL` and `Config.Net.TLS` are enabled
```
travis-minke-sap, line 93: can you be more specific and say where in the configmap the username/password are (the sarama yaml string with mention of object path).  My first thought was "where do have that in the configmap" and then remembered it was in the sarama yaml ; )

pkg/channel/distributed/common/kafka/README.md
```diff
@@ -46,13 +47,11 @@ layer. Instead, users are required to either use their
...
 go-client is easier to use but only provides limited EventHub related
-functionality (e.g - No "namespace" management, etc...) This implementation is
+functionality (e.g. - No "namespace" management, etc...) This implementation is
 currently based on the go-client and expects Azure Namespaces to be manually
 allocated and pre-existing. The implementation does, however, abstract away
 these Azure Namespaces so that a user's Azure Subscription looks like one big
-allocation of possible EventHubs. The creation of new Topics / EventHubs will be
-load-balanced across the available Azure EventHub Namespaces as identified by
-their K8S Secret (instead of dynamic lookup via the Azure REST API).
+allocation of possible EventHubs.
```
travis-minke-sap, line 55: Not your fault (thanks for catching this) but... I don't think this is true any longer.  When I removed the ability to support multiple Kafka Secrets for Azure we lost the ability to span multiple Azure EventHub Namespaces and are now limited to one.   How about something like ...

```
While Azure EventHubs support the standard Kafka interfaces for Producer /
Consumer based event transmission, it does not yet provide a similar management
layer. Instead, users are required to either use their
[REST API](https://docs.microsoft.com/en-us/rest/api/eventhub/) or the
[go-client](https://github.com/Azure/azure-event-hubs-go/tree/master). The
go-client is easier to use but only provides limited EventHub related
functionality (e.g. - No "namespace" management, etc...) This implementation is
currently based on the go-client and expects Azure Namespaces to be manually
allocated and pre-existing.  Since Azure requires unique authentication for each
EventHub Namespace, we are currently limited to supporting a single instance
with its inherent limitations as to the number of Topics that can be created.
```

pkg/channel/distributed/common/kafka/sarama/sarama.go
```diff
@@ -23,14 +23,18 @@ import (
...
 	"github.com/Shopify/sarama"
 	"github.com/ghodss/yaml"
 	"knative.dev/eventing-kafka/pkg/common/client"
 	commonconfig "knative.dev/eventing-kafka/pkg/common/config"
 	"knative.dev/eventing-kafka/pkg/common/constants"
 )
 
-// Utility Function For Enabling Sarama Logging (Debugging)
+const DefaultAuthSecretName = "kafka-cluster"
+
```
travis-minke-sap, line 36: This constant is currently in `pkg/common/constants/constants.go` and is also being added below in `distributed/controller/constants/constants.go`.   Is it possible to provide a single place for this?

pkg/channel/distributed/controller/README.md
```diff
@@ -5,12 +5,11 @@ reconciler, etc.) based on the latest knative-eventing SharedMain reconciler
...
-watching "Kafka" Secrets (those in knative-eventing labelled
-`eventing-kafka.knative.dev/kafka-secret: "true"`) which creates the Receiver /
-Producer Deployment & Service, and another which is watching `KafkaChannel`
-resources which provisions the Kafka Topic and creates the Dispatcher / Consumer
-Deployment & Service.
+instances in the K8S Cluster, watching `KafkaChannel` resources and performing
+the following functions:
+ - Provision the Kafka Topic
+ - Create the Dispatcher / Consumer Deployment And Service
+ - Create the Receiver / Producer Deployment And Service
```
travis-minke-sap, line 13: nit-picking, but maybe a bit more detail... ?

```
 - Provision the Kafka Topic
 - Create the single Receiver (Kafka Producer) Deployment And Service
 - Create a unique  Dispatcher (Kafka Consumer) Deployment And Service
```

pkg/channel/distributed/controller/kafkachannel/reconciler.go
```diff
@@ -206,17 +214,37 @@ func (r *Reconciler) reconcile(ctx context.Context, channel *kafkav1beta1.KafkaC
...
+		}
+
+		// Not having an auth secret for the Receiver is a problem
+		channel.Status.MarkConfigFailed(event.KafkaSecretReconciled.String(), "No Kafka Secret For KafkaChannel")
+		return fmt.Errorf(constants.ReconciliationFailedError)
+	} else if err != nil {
+		logging.FromContext(ctx).Error("Error reading receiver secret", zap.Error(err))
+		return fmt.Errorf(constants.ReconciliationFailedError)
+	}
+
```
travis-minke-sap, line 240: I would vote (unless theres a good reason not to?) that the above logic be refactored INTO the `reconcileReceiver()` function such that this top-level "reconciler" function can just ALWAYS call `reconcileReceiver()` which will check for secrets and such.  would make this function more "consistent" and keep the "receiver" specific logic in the receiver.go file, etc..

pkg/channel/distributed/receiver/README.md
```diff
@@ -4,12 +4,10 @@ The Receiver implementation is a Kafka Producer which is responsible for
...
-A unique Deployment / Service is created for every Kafka Secret (in the
-knative-eventing namespace and labelled with
-`eventing-kafka.knative.dev/kafka-secret: "true"`). The single Deployment is
-horizontally scalable as necessary (via controller environment variables.) This
-allows for an efficient use of cluster resources while still supporting high
-volume and multi-tenant use cases.
+A unique Deployment / Service is created for every installation.
+The single Deployment is horizontally scalable as necessary (via controller
+environment variables). This allows for an efficient use of cluster resources
+while still supporting high volume and multi-tenant use cases.
```
travis-minke-sap, line 11: maybe add something about users setting up their own HPA instead of the Controller setting replicas?

pkg/channel/distributed/common/kafka/README.md
eric-sap: Modified as suggested

config/channel/distributed/README.md
eric-sap: Sure; added references to Net.SASL.User and Net.SASL.Password explicitly

pkg/channel/distributed/common/kafka/sarama/sarama.go
eric-sap: The other constant (ReceiverPrefix) has been removed due to the reversion to secret-name-and-hash based receiver naming

pkg/channel/distributed/controller/README.md
eric-sap: Modified this a bit

pkg/channel/distributed/controller/kafkachannel/reconciler.go
eric-sap: That's a very good idea; it saved a bit of trouble as well as being better organized

pkg/channel/distributed/receiver/README.md
eric-sap: Added a note to that effect

pkg/channel/distributed/controller/kafkachannel/controller.go
eric-sap: After some more discussion we've decided to leave the deployment and service name along (reverted the code back to "use the secret name and hash for the deployment/service names"  Changing that for the sake of appearance would involve writing a new version-upgrade task to move the endpoints of all of the existing KafkaChannel services to point to the newly-renamed Receiver service.  Not impossible by any means but not really worth the bother for what is effectively "just a name change" for this issue.

docs/kind_kafka_cluster.md
eric-sap: I've changed the example to use 0.21.1

eric-sap:
/unhold
This was a false alarm; my upgrade configmap was incorrect (TLS and SASL were enabled when they weren't beforehand).  I have re-tested the distributed upgrade scenario and it does work.  Changing the secret name afterward (to verify that the new code is in place) also does create a receiver deployment based on the new secret name (as expected; issue #608 to resolve).  I think this is ready to go.

travis-minke-sap (approved):
/lgtm

eric-sap:
/retest
Looks like a timeout issue; will investigate further if it happens again

PR #604, opened 2021-05-06
[source] fix issue with offset not being initialized before session is closed

lionelvillard:
Make sure the KafkaSource is Ready (ie. guarantee to not lose events) before starting the consumer group. 

Without this fix, events sent to partitions with an uninitialized consumer group offset could be lost (as per def above) if these events are sent while the consumer group session is closed or is being rebalanced.

## Proposed Changes

- Make sure all consumer group offsets are initialized before starting consuming events.
- 
-

**Release Note**

```release-note

```

**Docs**

 

/cc @devguyio @slinkydeveloper

### DIFF
--- a/pkg/common/consumer/consumer_handler.go
+++ b/pkg/common/consumer/consumer_handler.go
@@ -24,8 +24,6 @@ import (
 	"go.uber.org/zap"
 )
 
-var _ sarama.ConsumerGroupHandler = (*SaramaConsumerHandler)(nil)
-
 type KafkaConsumerHandler interface {
 	// When this function returns true, the consumer group offset is marked as consumed.
 	// The returned error is enqueued in errors channel.
@@ -54,16 +52,6 @@ func WithSaramaConsumerLifecycleListener(listener SaramaConsumerLifecycleListene
 	}
 }
 
-func noopOffsetInitializer(sarama.ConsumerGroupSession) error {
-	return nil
-}
-
-func WithOffsetInitializer(initializer func(sarama.ConsumerGroupSession) error) SaramaConsumerHandlerOption {
-	return func(handler *SaramaConsumerHandler) {
-		handler.offsetInitializer = initializer
-	}
-}
-
 // ConsumerHandler implements sarama.ConsumerGroupHandler and provides some glue code to simplify message handling
 // You must implement KafkaConsumerHandler and create a new SaramaConsumerHandler with it
 type SaramaConsumerHandler struct {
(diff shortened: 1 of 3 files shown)


pkg/common/consumer/consumer_handler.go
```diff
@@ -52,13 +54,24 @@ func WithSaramaConsumerLifecycleListener(listener SaramaConsumerLifecycleListene
 	}
 }
 
+func noopOffsetInitializer(sarama.ConsumerGroupSession) error {
+	return nil
+}
+
+func WithOffsetInitializer(initializer func(sarama.ConsumerGroupSession) error) SaramaConsumerHandlerOption {
```
slinkydeveloper, line 61: Can you add a comment explaining where this is invoked? And why can't you just use the lifecycle listener instead?

pkg/source/adapter/adapter.go
```diff
@@ -226,3 +232,71 @@ func (a *Adapter) Cleanup(sess sarama.ConsumerGroupSession) {
...
+		// consumed from ALL partitions.
+		// If not, an event sent to a partition with an uninitialized offset
+		// will not be forwarded when the session is closed (or a rebalancing is in progress).
+		kafkaClient, err := sarama.NewClient(addrs, config)
+		if err != nil {
+			return fmt.Errorf("failed to create a Kafka client: %w", err)
+		}
+		defer kafkaClient.Close()
+
+		kafkaAdminClient, err := sarama.NewClusterAdminFromClient(kafkaClient)
```
slinkydeveloper, line 250: The problem with this approach is that it makes the non-so-trivial assumption that the kafka config provided to access the kafka cluster has admin rights. IMO this assumption might not work in a lot of production use cases... Any ideas how to go around it? Or at least, make this behaviour optional?

pkg/common/consumer/consumer_handler.go
lionelvillard: I can't recall exactly why. Maybe because the lifecyclelistener does return an error. Anyway I changed it as suggested.

slinkydeveloper: if you need to, change the signature to return an error

pkg/source/adapter/adapter.go
lionelvillard: Now if the admin client cannot be created a warning is printed in the log. It's not an error anymore.

slinkydeveloper (approved):
/lgtm

PR #618, opened 2021-05-11
[mtsource] wait for stadapter to stop before starting the new one

lionelvillard:
## Proposed Changes

- Make sure the old adapter really stop before starting the new one.
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/source/mtadapter/adapter.go
+++ b/pkg/source/mtadapter/adapter.go
@@ -55,6 +55,11 @@ func NewEnvConfig() adapter.EnvConfigAccessor {
 	return new(AdapterConfig)
 }
 
+type cancelContext struct {
+	fn      context.CancelFunc
+	stopped chan bool
+}
+
 type Adapter struct {
 	config      *AdapterConfig
 	logger      *zap.SugaredLogger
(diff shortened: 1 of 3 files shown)


pkg/source/mtadapter/adapter.go
```diff
@@ -241,14 +249,21 @@ func (a *Adapter) Update(ctx context.Context, obj *v1beta1.KafkaSource) {
 	}
 
 	ctx, cancelFn := context.WithCancel(ctx)
+
+	cancel = cancelContext{
+		fn:      cancelFn,
+		stopped: make(chan bool),
```
steven0711dong, line 255: Why do we want to name this `stopped`, since we put true into this channel after adapter starts, I think we should rename it to `inProgress` indicating adapter is currently running.

lionelvillard: adapter.Start() is blocking. We put true when the adapter stopped.

lionelvillard:
/hold

the single-tenant adapter does not wait...

lionelvillard:
/unhold

should be ok now.

/cc @steven0711dong

lionelvillard:
/hold

seems like it's not quite right yet

pkg/source/mtadapter/adapter.go
steven0711dong: I see. Ok, it should be named stopped.

steven0711dong (approved):
LGTM

Issue #608, opened 2021-05-07
Distributed channel needs to handle changing the authSecretName/authSecretNamespace


eric-sap:
**Describe the bug**
After the update for issue #602 (PR #603), the name of the single secret is given in the config-kafka configmap via the authSecretName and authSecretNamespace fields in the eventing-kafka.kafka section.  As long as this secret name is the same as whatever the cluster admin has been using prior to this update, there will be no trouble.  However, if the admin changes this authSecretName (either implicitly by using a different secret name than the default of 'kafka-cluster' before the upgrade, or explicitly by changing the authSecretName in the configmap of a live system), a new Receiver deployment and service will be created, which will be used for any new KafkaChannel reconciliations, but existing KafkaChannel services will NOT be adjusted to use the new Receiver Service.  The old Receiver deployment and service will remain and will be functional as long as the old secret is in place, but are orphaned and the migration and cleanup must be done manually.

 **Expected behavior**
Changing the authSecretName in the config-kafka configmap should result in either:
- A new Receiver Service and Deployment are created and an automatic migration of existing KafkaChannel Services to use the new externalName, OR
- The existing Receiver Service and Deployment should continue to be the only ones, and no migration is necessary

The difference between those two approaches is that for the former, the name of the Receiver is still based on the name of the Secret.  This allows old installations to continue to work without changes but requires the migration code whenever the secret name is changed.  For the latter, there should be one constant name for the Receiver Service and Deployment (e.g. "kafka-cluster-receiver") to which all KafkaChannel Services always point.

The second method is the idea solution but still involves a necessary migration step to move existing KafkaChannel Services to the new one.  Theoretically, we should be able to create the new, constant-name service/deployment and run the migration without any downtime, and then delete the old service/deployment after the KafkaChannels are all migrated.

 **To Reproduce**
Set up an existing distributed-channel system (with KafkaChannels) using a secret named 'eventing-kafka-secret', then upgrade to a version that includes the code from PR #603.  Note that a new Receiver Service/Deployment is created named 'kafka-cluster-32603413-receiver' and that new KafkaChannel Services have 'externalName: kafka-cluster-32603413-receiver.knative-eventing.svc.cluster.local' in their spec.  Old KafkaChannels will be pointing to the one named 'eventing-kafka-secret-ed340a19-receiver.knative-eventing.svc.cluster.local' and are never updated unless destroyed and recreated. 

 **Knative release version**
0.22.1

 **Additional context**
[...]

PR #622, opened 2021-05-15
Consolidate configmap

### LINKED ISSUE #405: Leverage eventing-kafka field of config-kafka in consolidated channel and in source
**Problem**
There's an `eventing-kafka` [field](https://github.com/knative-sandbox/eventing-kafka/blob/master/config/channel/distributed/300-eventing-kafka-configmap.yaml#L39) in `config-kafka` of the distributed channel:

```yaml
eventing-kafka: |
    receiver:
      cpuRequest: 100m
      memoryRequest: 50Mi
      replicas: 1
    dispatcher:
      cpuRequest: 100m
      memoryRequest: 50Mi
      replicas: 1
    kafka:
      brokers: REPLACE_WITH_CLUSTER_URL
      enableSaramaLogging: false
      topic:
        defaultNumPartitions: 4
        defaultReplicationFactor: 1 # Cannot exceed the number of Kafka Brokers!
        defaultRetentionMillis: 604800000  # 1 week
      adminType: kafka # One of "kafka", "azure", "custom"
```

To be consistent (see the umbrella ticket, https://github.com/knative-sandbox/eventing-kafka/issues/404), we can also leverage these settings in consolidated channel and also in Kafka source.

To talk about fields specifically:
- `receiver`: could only be used in distributed channel
- `dispatcher`: could be used in distributed and consolidated channel
- `kafka.brokers`: Consolidated channel has a separete field in `config-kafka` called `bootstrapServers`. We should leverage that in distributed channel too. `eventing-kafka.kafka.brokers` then would be used as a default OR if possible, we can get rid of that.
- `kafka.enableSaramaLogging`: could be used in all 3 components (2 channels and the source)
- `kafka.topic`: could be also used in consolidated channel
- `kafka.adminType`: could be also used in consolidated channel

**Exit Criteria**

**Time Estimate (optional):**

eric-sap:
Fixes #405 

## Proposed Changes

- 🧽  Changed sarama.LoadSettings() to return only one "settings" struct (the EventingKafkaConfig) which includes the Sarama config (with auth embedded) as a separate entry, and refactored the related calling functions
- 🧽  Moved a fair bit of sarama-related code into common for use by consolidated and distributed channels
- 🧽  Reorganized config-kafka configmap for better use by multiple components (consolidated and distributed) that don't have the same dispatcher/receiver structure
- 🧽  Made "replicas" an optional entry (default 1) for the dispatcher and receiver config fields with the expectation that HorizontalPodAutoscaler resources will be the preferred mechanism for adjusting these
- 🧽 Moved the "enableSaramaLogging" field from eventing-kafka to the sarama field "enableLogging" (and moved the sarama config string one level down to the "config" field

**Release Note**

```release-note
The current config-kafka configmap format has changed (no immediate action required; the old format will continue to work)
```


eric-sap:
/hold
I would like to hold this for review for at least a few days as this is a cross-channel-type change and it will be far easier to address any desired adjustments to the configmap now rather than after it's in use.

pkg/channel/distributed/controller/testing/data.go
```diff
@@ -165,6 +164,7 @@ Producer:
   Idempotent: true
   RequiredAcks: -1
 `
+	ConfigMapVersion = "1.0.0"
 )
```
travis-minke-sap, line 168: Is there another real (non-test) constant for this elsewhere that should be used ?

test/e2e-common.sh
```diff
@@ -370,7 +370,14 @@ function run_postinstall_jobs() {
...
   kubectl delete secret -n "${SYSTEM_NAMESPACE}" kafka-cluster
-  sleep 10 # Give Controller Time To React To Kafka Secret Deletion ; )
+
+  # The distributed channel controller no longer actively monitors the kafka-cluster secret,
+  # so the receiver will only be torn down if there is a change to the kafkachannels or the
+  # config-kafka configmap.  However, if the secret is missing, the global resync does not happen
+  # (the controller presumes this missing secret is an error and reacts accordingly), so this
+  # manual deletion of the receiver deployment is the easiest solution here.
+  [[ $1 == "distributed" ]] && kubectl delete deployment -n "${SYSTEM_NAMESPACE}" kafka-cluster-32603413-receiver
+
```
travis-minke-sap, line 380: do we need to delete the receiver Service as well ?

travis-minke-sap (changes requested):
looking good - thanks Eric!

pkg/common/config/eventingkafkaconfig.go
```diff
@@ -58,17 +60,43 @@ type EKCloudEventConfig struct {
...
+}
+
+// EKSourceConfig is reserved for configuration fields needed by the Kafka Source component
+type EKSourceConfig struct {
+}
+
+// EKChannelConfig contains items relevant to the eventing-kafka channels (distributed and consolidated)
+type EKChannelConfig struct {
+	Dispatcher   EKDispatcherConfig   `json:"dispatcher,omitempty"`
+	Distributed  EKDistributedConfig  `json:"distributed,omitempty"`
```
travis-minke-sap, line 85: just wondering if the ConfigMap would "read" better (more consistent) if the `Dispatcher` was under Distributed AND Consolidated.  I realize it is "common" to both Channels but a third channel could come along with a different architecture?  It also aligns nicely next to the "Receiver" in distributed case.  Users would only ever have one type of ConfigMap (consolidated / distributed) so it wouldn't appear in there twice.   Anyway - not a big deal, just a minor organziation preference - would be good to hear from others as well ; )

pkg/channel/distributed/controller/testing/data.go
eric-sap: Originally I made this its own constant in the test code specifically so that if we change the "real" version (CurrentConfigVersion in common/config/eventingkafkaconfig.go) without paying attention to the test, then the test will fail.  However ... the upgrade code became less complicated and the tests that used it aren't there, so now it's only in TestNewController(), which should be using the common constant.  Will fix.

pkg/common/kafka/sarama/sarama.go
```diff
@@ -0,0 +1,263 @@
...
+	if enable {
+		sarama.Logger = log.New(os.Stdout, "[sarama] ", log.LstdFlags)
+	} else {
+		sarama.Logger = log.New(ioutil.Discard, "[Sarama] ", log.LstdFlags)
+	}
+}
+
+// GetAuth Is The Function Type Used To Delay Loading Auth Config Until The Secret Name/Namespace Are Known
+type GetAuth func(context.Context, string, string) (*client.KafkaAuthConfig, error)
+
```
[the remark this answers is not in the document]
eric-sap: Yep; this used to be a test-only function and when I moved it out to real-code use I neglected to pay attention to that; will fix.

test/e2e-common.sh
eric-sap: That would be a good idea; I forgot about that.  Distributed is the last test before the cluster gets torn down anyway but it's good to be clean.

config/channel/distributed/300-eventing-kafka-configmap.yaml
```diff
@@ -14,45 +14,52 @@
...
+      authSecretName: kafka-cluster
+      authSecretNamespace: knative-eventing
       brokers: REPLACE_WITH_CLUSTER_URL
-      enableSaramaLogging: false
       topic:
         defaultNumPartitions: 4
         defaultReplicationFactor: 1 # Cannot exceed the number of Kafka Brokers!
         defaultRetentionMillis: 604800000  # 1 week
-      adminType: kafka # One of "kafka", "azure", "custom"
+    channel:
```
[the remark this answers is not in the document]
eric-sap: This shouldn't be much trouble if desired; a change to consolidated/reconciler/controller/kafkachannel.go where it calls kafkaClusterAdmin.CreateTopic()  (that function now has access to r.kafkaConfig.EventingKafka.Kafka.Topic so it seems easy enough)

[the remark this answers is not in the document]
eric-sap: I added this (and moved the NumPartitions/etc. functions to common/config/util.go)

pkg/common/config/eventingkafkaconfig.go
eric-sap: I tried it out and it does look less wonky that way.  New configmap format:
```
  eventing-kafka: |
    ...
    channel:
      distributed:
        adminType: kafka
        dispatcher:
          cpuRequest: 100m
          memoryRequest: 50Mi
        receiver:
          cpuRequest: 100m
          memoryRequest: 50Mi
```
or
```
  eventing-kafka: |
    ...
    channel:
      consolidated:
        dispatcher:
          cpuRequest: 100m
          memoryRequest: 50Mi
```
And as an extra bonus, the EKConsolidatedConfig struct is no longer "reserved for future use"

pkg/channel/consolidated/utils/util.go
```diff
@@ -115,7 +114,7 @@ func GetKafkaAuthData(ctx context.Context, secretname string, secretNS string) *
 	parseTls(secret, kafkaAuthConfig)
 	parseSasl(secret, kafkaAuthConfig)
 
-	return kafkaAuthConfig
+	return kafkaAuthConfig, nil
 }
```
travis-minke-sap, line 118: This seems odd - we're adding an `error` return value but then ALWAYS returning nil ?  What value does it provide?

eric-sap: It's allowing the GetKafkaAuthData function to satisfy the sarama.GetAuth prototype so it can be passed in directly to the LoadSettings call in GetKafkaConfig().  Otherwise it would need a one-line wrapper to return the nil error (which I would be happy to do if that would look better).

pkg/channel/distributed/controller/config/config_test.go
```diff
@@ -43,6 +44,8 @@ const (
 	receiverCpuRquest     = "10m"
 	receiverMemoryLimit   = "20Mi"
 	receiverCpuLimit      = "100m"
+
+	consolidatedDispatcherReplicas = 1
 )
```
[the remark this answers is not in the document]
eric-sap: You're right; I was originally going to call VerifyConfiguration in consolidated as well, but decided that was not the scope here and forgot to remove this kind of stuff.  I'll clean it up.

pkg/common/config/util.go
```diff
@@ -54,3 +57,38 @@ func GetAuthConfigFromSecret(secret *corev1.Secret) *client.KafkaAuthConfig {
...
+	}
+	return value
+}
+
+// RetentionMillis Gets The RetentionMillis - First From Channel Spec And Then From ConfigMap-Provided Settings
+func RetentionMillis(channel *kafkav1beta1.KafkaChannel, configuration *EventingKafkaConfig, logger *zap.SugaredLogger) int64 {
+	//
+	// TODO - The eventing-contrib KafkaChannel CRD does not include RetentionMillis so we're
+	//        currently just using the default value specified in Controller Environment Variables.
+	//
```
travis-minke-sap, line 86: The `eventing-contrib` comment is out-of-date and should be reworded to reflect the history.  Essentially the current KafkaChannel spec doesn't have this and maybe we should look at adding it (was in old versions of distributed channel CRD.)

pkg/common/config/util_test.go
```diff
@@ -120,3 +130,64 @@ func getSaramaTestSecret(t *testing.T, name string,
...
+	// Test Data
+	configuration := &EventingKafkaConfig{Kafka: EKKafkaConfig{Topic: EKKafkaTopicConfig{DefaultRetentionMillis: defaultRetentionMillis}}}
+
+	// Test The Default Failover Use Case
+	channel := &kafkav1beta1.KafkaChannel{}
+	actualRetentionMillis := RetentionMillis(channel, configuration, logger)
+	assert.Equal(t, defaultRetentionMillis, actualRetentionMillis)
+
+	// TODO - No RetentionMillis In eventing-contrib KafkaChannel
+	//// Test The Valid RetentionMillis Use Case
```
travis-minke-sap, line 189: same - fix comment

travis-minke-sap:
I added a couple more trivial comments but otherwise LGTM - will wait for a "consolidated" expert to do final approval.
Lots of great work here - thanks Eric!

pkg/common/config/util.go
eric-sap: I removed the RetentionMillis call and am now just using the Kafka.Topic.DefaultRetentionMillis value from the configmap directly where it is needed (consolidated reconcileTopic and distributed reconcileKafkaTopic).  I put the updated TODO comment in those locations so it will not be lost.

pkg/common/config/util_test.go
eric-sap: Functionality/Comment has been moved (see previous review comment)

eric-sap:
/retest
The linter appears to have .. crashed?

eric-sap:
The test are passing and I think this is ready but I'd like to leave it for just a bit more in a final state in case someone from the consolidated side wants to weigh in.

eric-sap:
This is showing up in the kafka-go-coverage test, and I'm fairly certain it's the extra newline in the Bearer field (which is why the length is 41), but I don't know what's creating that value.
```
2021/05/21 01:30:22 Running workflow: presubmit
Warning: token len = 41, which is different from 40
...
2021/05/21 01:30:22 error running c.PullRequests.ListFiles(Ctx, repoOwner=knative-sandbox, RepoName=eventing-kafka, pullNum=622, listOptions): Get "https://api.github.com/repos/knative-sandbox/eventing-kafka/pulls/622/files?page=1": net/http: invalid header field value "Bearer ghp_tYww[...]7wKf\n" for key Authorization
```

eric-sap:
@chizhg  much appreciated!

config/channel/distributed/300-eventing-kafka-configmap.yaml
```diff
@@ -14,45 +14,52 @@
...
+      distributed:
```
aliok, line 55: I am not 100% sure but I think it might be a better idea to not have the separation of distributed/consolidated in the configmap. In the long run, if we want to have a single channel, users would do another migration of the configmap.

There's pros/cons in both. I would go for not having `consolidated` or `distributed` nested fields and let the implementations ignore the fields they don't use. Just a thought but I am also fine with the approach in the PR.

aliok: In the end, users can't use both channels at the same time anyway.

pkg/common/client/admin_client.go
```diff
@@ -1,21 +1,14 @@
...
-func MakeAdminClient(ctx context.Context, clientID string, kafkaAuthCfg *KafkaAuthConfig, saramaSettingsYamlString string, brokers []string) (sarama.ClusterAdmin, error) {
-	saramaConf, err := NewConfigBuilder().
-		WithDefaults().
-		WithAuth(kafkaAuthCfg).
-		WithClientId(clientID).
-		FromYaml(saramaSettingsYamlString).
-		Build(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("error creating admin client Sarama config: %w", err)
+func MakeAdminClient(saramaConf *sarama.Config, brokers []string) (sarama.ClusterAdmin, error) {
```
aliok, line 9: Maybe we can get rid of this function and the file?

test/config/config-kafka-sasl.yaml
```diff
@@ -3,9 +3,14 @@ kind: ConfigMap
 metadata:
   name: config-kafka
   namespace: ${SYSTEM_NAMESPACE}
+  # eventing-kafka.kafka.broker: Replace this with the URLs for your kafka cluster,
```
aliok, line 6:
```suggestion
  # eventing-kafka.kafka.brokers: Replace this with the URLs for your kafka cluster,
```

test/config/config-kafka-tls.yaml
```diff
@@ -3,9 +3,14 @@ kind: ConfigMap
 metadata:
   name: config-kafka
   namespace: ${SYSTEM_NAMESPACE}
+  # eventing-kafka.kafka.broker: Replace this with the URLs for your kafka cluster,
```
aliok, line 6:
```suggestion
  # eventing-kafka.kafka.brokers: Replace this with the URLs for your kafka cluster,
```

test/e2e-common.sh
```diff
@@ -370,7 +370,17 @@ function run_postinstall_jobs() {
...
-  sleep 10 # Give Controller Time To React To Kafka Secret Deletion ; )
+
+  # The distributed channel controller no longer actively monitors the kafka-cluster secret,
+  # so the receiver will only be torn down if there is a change to the kafkachannels or the
+  # config-kafka configmap.  However, if the secret is missing, the global resync does not happen
+  # (the controller presumes this missing secret is an error and reacts accordingly), so this
+  # manual deletion of the receiver deployment is the easiest solution here.
+  if [[ $1 == "distributed" ]]; then
+    kubectl delete deployment -n "${SYSTEM_NAMESPACE}" kafka-cluster-32603413-receiver
+    kubectl delete service -n "${SYSTEM_NAMESPACE}" kafka-cluster-32603413-receiver
```
aliok, line 381: `32603413` is there a deterministic process to generate this number or is this just forgotten?

aliok:
@eric-sap I added a few comments. Let's resolve those and I can have a 2nd review.

Thanks a lot for the PR! I truly appreciate your ~30 commits and patience with this consolidation PR. Awesome work!

test/e2e-common.sh
eric-sap: Yes, this comes from GenerateHash() in pkg/channel/distributed/controller/util/hash.go
For a secret named "kafka-cluster" the hash will always be 32603413
We could replicate the GenerateHash function in shell script here if that's more understandable; it would look something like this but with a parameter
```
echo -n "kafka-cluster" | md5 | sed 's/^\(........\).*/\1/g'
```

test/config/config-kafka-sasl.yaml
eric-sap: got it, thanks

test/config/config-kafka-tls.yaml
eric-sap: got it, thanks

pkg/common/client/admin_client.go
eric-sap: Good idea; will remove

test/e2e-common.sh
eric-sap: We are planning, however (with issue #608), to use a non-hash-decorated single-name for the receiver, so I don't think adding this hash function (simple though it is) to the bash script will really be a long-term gain in usefulness.  (It does need to stay in the go code; the same function is used for the dispatcher names -- those pods are removed automatically when the kafkachannel is deleted during the test so a similar kubectl command isn't needed here)

eric-sap: This ended up being such a small change that I opted for the visibility of adding the hash generation and a comment about the reason for the value, even if it ends up being temporary.

config/channel/consolidated/configmaps/kafka-config.yaml
```diff
@@ -13,13 +13,20 @@
 # limitations under the License.
 
 apiVersion: v1
+data:
+  version: 1.0.0
+  # eventing-kafka.kafka.brokers: Replace this with the URLs for your kafka cluster,
+  #   which is in the format of my-cluster-kafka-bootstrap.my-kafka-namespace:9092.
+  # eventing-kafka.kafka.authSecretName: name-of-your-secret-for-kafka-auth
+  # eventing-kafka.kafka.authSecretNamespace: namespace-of-your-secret-for-kafka-auth
```
matzew, line 21: @eric-sap can we have some doc update for these changes on the docs repo:
https://github.com/knative/docs/tree/mkdocs/docs/eventing/samples/kafka/channel#authentication-against-an-apache-kafka

pkg/common/kafka/sarama/sarama.go
```diff
@@ -0,0 +1,266 @@
...
+// StringifyHeaderPtrs Is A Pointer-version of the StringifyHeaders function
+func StringifyHeaderPtrs(headers []*sarama.RecordHeader) map[string][]string {
+	stringHeaders := make(map[string][]string)
+	for _, header := range headers {
+		key := string(header.Key)
+		stringHeaders[key] = append(stringHeaders[key], string(header.Value))
+	}
+	return stringHeaders
+}
```
matzew, line 266: can both call into the same "helper" ?

matzew:
LGTM

test/e2e-common.sh
aliok: Thanks. I am happy with both as long as there's determinism :)

aliok (approved):
/lgtm
/approve

Thanks for addressing the comments.

aliok:
@eric-sap I didn't want to unhold your `/hold` ;)

pkg/common/kafka/sarama/sarama.go
eric-sap: I have tried to combine these functions and the result is not any better than these separate ones because of the amount of overhead necessary ("slice of pointers" is different than "slice of RecordHeaders", despite go papering over that and allowing you to write "header.Value" for either).  A Go master might know something better (happy to accept something!) but I threw in the towel.

config/channel/consolidated/configmaps/kafka-config.yaml
eric-sap: Absolutely; I will get on that.  Thank you for the pointer.

PR #626, opened 2021-05-17
Resetoffset controller

travis-minke-sap:
Subtask of #588 to create the shared ResetOffset Controller framework.

## Proposed Changes

- Create a reusable Controller for the ResetOffset CRD which is flexible enough to support different ResetOffset.Spec.Ref types, as well as custom topic/group naming conventions within a specific Ref type.
- Implementation is incomplete in that it is not yet stopping / starting ConsumerGroups.  To manually test this with the distributed KafkaChannel implementation you can simple scale the Dispatcher replicas between 1 and 0 as needed.
- Enhanced the lifecycle states to include concepts of "initiated" and "completed" as well tracking the status of the "Ref mapping" (turning the Spec.Ref KReference into a Kafka TopicName / ConsumerGroup ID).

**Release Note**

**Docs**

**Dev Note**
To add this new Controller to an existing deployment would look like the following (taken from dev testing against the distributed KafkaChannel)...
```
// Create A Subscription RefMapper Factory With Custom Topic/Group Naming
subscriptionRefMapperFactory := refmappers.NewSubscriptionRefMapperFactory(controllerutil.TopicNameMapper, controllerutil.GroupIdMapper)

// Create A ResetOffset ControllerConstructor Factory With Custom Subscription Ref Mapping
resetOffsetControllerConstructor := resetoffset.NewControllerFactory(subscriptionRefMapperFactory)

// Create The SharedMain Instance With The Various Controllers
sharedmain.MainWithContext(ctx, constants.ControllerComponentName, kafkachannel.NewController, resetOffsetControllerConstructor)
```
...where the above is creating a "Subscription" mapper with custom TopicName / GroupId mappers.  Both layers of abstraction are need to support the desired use cases.  If someone (kafka-broker) wanted to support Triggers instead of Subscriptions then they would implement a new "Trigger" mapper.  The "distributed" and "consolidated" KafkaChannel implementations both want to use the Subscription mapper but with different TopicName / GroupId mappers.

/hold


eric-sap (commented):
This looks like good stuff so far.  Some minor comments

pkg/common/commands/resetoffset/controller/offsets.go
```diff
@@ -0,0 +1,210 @@
...
+	// Return Results
+	return offsetMapping, err
+}
+
+// formatOffsetMetaData returns a "metadata" string, suitable for use with MarkOffset/ResetOffset, for the specified time.
+func formatOffsetMetaData(time int64) string {
+	return fmt.Sprintf("resetoffset.%d", time)
+}
+
+// safeCloseSaramaClient will attempt to close the specified Sarama Client and will reset the Reconciler reference
```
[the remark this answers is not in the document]
travis-minke-sap: yeah - copy/paste comment from kafkachannel controller ; ).   will fix - thanks!

pkg/common/commands/resetoffset/refmappers/subscription.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	// Attempt To Get The Specified Subscription
+	subscription, err := m.subscriptionLister.Subscriptions(refNamespace).Get(ref.Name)
+	if err != nil {
+		logger.Error("Failed to get Subscription referenced by ResetOffset", zap.Error(err))
+		return "", "", fmt.Errorf("failed to get Subscription referenced by ResetOffset.Spec.Ref '%v': %v", ref, err)
+	}
+	if subscription == nil {
+		logger.Info("No Subscription found for ResetOffset reference")
```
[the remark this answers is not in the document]
travis-minke-sap: I think `Info` is appropriate here but good question.  The reason is that it's not a failure/error as nothing unexpected has happened - only that the user has created an invalid ResetOffset CRD (pointing to non-existent Subscription) and we're just logging that fact.  If it was something the code was creating and it should always be there then a Warning might make more sense.   sound ok?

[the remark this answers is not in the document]
eric-sap: It does, especially since the code that calls MapRef will log its own error when this function returns an error.  Originally I was thinking "the admin will be looking for an error in the logs as to why the ResetOffset CR didn't work," but there will still be a "Failed to map ResetOffset.Spec.Ref to Kafka Topic name and ConsumerGroup ID" error in that case.  So I agree.

pkg/common/commands/resetoffset/refmappers/subscription.go
```diff
@@ -0,0 +1,155 @@
...
+
+	// Get The ResetOffset Ref From Spec & Enhance Logger
+	ref := resetOffset.Spec.Ref
+	logger := m.logger.With(zap.Any("Ref", ref))
+
+	// Validate The Reference
+	if !strings.HasPrefix(ref.APIVersion, messaging.GroupName) ||
+		ref.Kind != "Subscription" ||
+		len(ref.Name) <= 0 {
+		m.logger.Warn("Received ResetOffset with invalid Subscription reference")
```
slinkydeveloper, line 118: What about changing this to "ResetOffset supports only subscriptions"

pkg/common/commands/resetoffset/refmappers/subscription.go
```diff
@@ -0,0 +1,155 @@
...
+		return "", "", fmt.Errorf("received ResetOffset with invalid Subscription reference: %v", ref)
+	}
+
+	// Default Optional Ref.Namespace If Not Provided
+	refNamespace := ref.Namespace
+	if len(refNamespace) <= 0 {
```
slinkydeveloper, line 124: `refNamespace == ""` IMO is nicer to read, but you don't need to change it if you don't like it :)

travis-minke-sap: Yeah, not sure how I got into the habit of that - some misplaced blog early in my go development.  You are right the "" check is easier to read and they are essentially equivalent - I will change it  (and my habitual use of len() as well ; ). thx!

pkg/common/commands/resetoffset/controller/reconciler.go
```diff
@@ -0,0 +1,214 @@
...
+		logger.Debug("Skipping reconciliation of previously executed ResetOffset instance")
+		return reconciler.NewEvent(corev1.EventTypeNormal, ResetOffsetSkipped.String(), "Skipped previously executed ResetOffset")
+	}
+
+	// Set Defaults & Verify The ResetOffset Is Valid
+	resetOffset.SetDefaults(ctx)
+	if err := resetOffset.Validate(ctx); err != nil {
+		logger.Error("Received invalid ResetOffset", zap.Error(err))
+		eventRecorder.Event(resetOffset, corev1.EventTypeWarning, ResetOffsetSkipped.String(), "Skipping invalid ResetOffset")
+		return err
```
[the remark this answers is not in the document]
travis-minke-sap: I agree it is redundant once we have a Webhook stood up for this, but... I was following the example in both the consolidated and distributed KafkaChannel `ReconcileKind()` implementations.  Is there a reason for re-validating in those cases?  Is it because we used to have an "open" KafkaChannel CRD that didn't explicitly specify it's payload?  Does it hurt anything to keep this here? (I'm happy to remove it - just double checking on the inconsistency ; )

pkg/common/commands/resetoffset/refmappers/subscription.go
travis-minke-sap: I was using the "invalid Subscription" language to also cover the case where it IS a subscription but the Name was missing.  I will break it up though into two different checks though with better/detailed logging/errors in each case - because as you indicate it would be good to make it clear that it's only expecting Subscriptions.

pkg/common/commands/resetoffset/controller/reconciler.go
```diff
@@ -0,0 +1,214 @@
...
+	eventRecorder.Event(resetOffset, corev1.EventTypeNormal, ResetOffsetMappedRef.String(), "Successfully mapped 'ref' to Kafka Topic and Group")
+
+	// Parse The Sarama Offset Time From ResetOffset Spec
+	offsetTime, err := resetOffset.Spec.ParseSaramaOffsetTime()
+	if err != nil {
+		logger.Error("Failed to parse Sarama Offset Time from ResetOffset Spec", zap.Error(err))
+		return err // Should never happen assuming Validation is in place
+	}
+	logger.Info("Successfully parsed Sarama Offset time from ResetOffset Spec", zap.Int64("Time (millis)", offsetTime))
+	eventRecorder.Event(resetOffset, corev1.EventTypeNormal, ResetOffsetParsedTime.String(), "Successfully parsed Sarama offset time from Spec")
```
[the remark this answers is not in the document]
travis-minke-sap: ah, ok - sorry - wasn't sure of the best practice and err'd on the side of over communicating - I will scale it back - thanks!

pkg/common/commands/resetoffset/controller/reconciler.go
```diff
@@ -0,0 +1,214 @@
...
+	// TODO - Send "Stop" Message to Dispatcher Replicas to stop ConsumerGroups
+	resetOffset.Status.MarkConsumerGroupsStoppedTrue()
+
+	// Update The Sarama Offsets & Update ResetOffset CRD With OffsetMappings
+	// TODO offsetMappings, err := UpdateOffsetsFn(logger, r.kafkaBrokers, r.saramaConfig, topic, group, offsetTime)
+	offsetMappings, err := r.reconcileOffsets(ctx, topic, group, offsetTime)
+	if err != nil {
+		logger.Error("Failed to update Offsets of one or more partitions", zap.Error(err))
+		resetOffset.Status.MarkOffsetsUpdatedFailed("FailedToUpdateOffsets", "Failed to update offsets of one or more partitions: %v", err)
+		eventRecorder.Event(resetOffset, corev1.EventTypeWarning, ResetOffsetUpdatedOffsets.String(), "Failed to update offsets of one or more partitions")
```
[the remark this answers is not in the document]
travis-minke-sap: 👍  Will fix

pkg/common/commands/resetoffset/controller/offsets.go
```diff
@@ -0,0 +1,210 @@
...
+	} else if newOffset < currentOffset {
+		partitionOffsetManager.ResetOffset(newOffset, offsetMetaData)
+	}
+
+	// Check For Any Errors From The PartitionOffsetManager
+	select {
+	case err = <-partitionOffsetManager.Errors():
+	case <-time.After(1 * time.Second):
+	}
+
```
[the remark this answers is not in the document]
travis-minke-sap: Yeah, I agree fully and am not very happy with this current implementation.  Obviously anytime you're putting in a "sleep" or "wait" duration it's not good.  

This IS how the Sarama PartitionOffsetManager API is setup to work.  The `MarkOffset()` and `ResetOffset()` functions do NOT return anything, and you are intended to instead monitor the `Errors()` Channel.  The parent OffsetManager does not expose any error handling (say for `Commit()`)

```
	// Errors returns a read channel of errors that occur during offset management, if
	// enabled. By default, errors are logged and not returned over this channel. If
	// you want to implement any custom error handling, set your config's
	// Consumer.Return.Errors setting to true, and read from this channel.
	Errors() <-chan *ConsumerError
```

I think the correct way to improve this is to re-arrange the lifecycle handling of PartitionOffsetManager so that we wait for the `Errors()` channel to close AFTER closing the PartitionOffsetManager.  This is possible but complicates the implementation since we can't just "defer" the closing, and instead have to handle it in all exit cases.

I think what I have above was a temporary/lazy solution while I was busy working on the rest of the controller.  I can/will look into fixing that as described above but would prefer to do it outside (immediately after) this PR if that's acceptable.  The Controller is incomplete and not actively used anywhere so developing it incrementally should hopefully be acceptable?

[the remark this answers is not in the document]
travis-minke-sap: I had some time this afternoon to refactor that logic so it should be better now.  There is still a `time.After()` usage but it is no longer a vague guess at how long it might take the PartitionOffsetManager to write a ConsumerError to the Error channel.  Instead it is a large-enough window for the PartitionOffsetManager to close the Error channel once it is closing itself.  We're no longer `defer()`'ing the PartitionOffsetManager `close()` after we try to process the channel, but are instead closing first which should close the error channel, and are then draining the error channel.  This better aligns with the Sarama documentation on `AsyncClose()` (called by `Close()`)...
```
	// AsyncClose initiates a shutdown of the PartitionOffsetManager. This method will
	// return immediately, after which you should wait until the 'errors' channel has
	// been drained and closed. It is required to call this function, or Close before
	// a consumer object passes out of scope, as it will otherwise leak memory. You
	// must call this before calling Close on the underlying client.
	AsyncClose()
```

pkg/common/commands/resetoffset/controller/reconciler.go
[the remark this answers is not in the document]
slinkydeveloper: > Is there a reason for re-validating in those cases?

TBH I don't know, for sure revalidating doesn't make sense... Maybe setting defaults is important? But I don't think so too, because the webhook does also the defaulting part

pkg/common/commands/resetoffset/controller/offsets.go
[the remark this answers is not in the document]
slinkydeveloper: Just checked right now. More than having the `drainPartitionOffsetManagerErrors`, I would love to have an internal API that hides completely that weirdness, something like:

```
MarkOffset(partitionOffsetManager, newOffset, offsetMetaData) error
```

So when I read this code, I don't need to reason about the bad Sarama API. WDYT?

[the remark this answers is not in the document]
travis-minke-sap: I don't think it's that easy.   The function would have to be "MarkOffsetAndCloseParttitionOffsetManager()" because it's important to close the PartitionOffsetManager prior to draining.  Then we're combining the "end" of the process (marking offset, closing partition manager, and draining error channel) all into one big step and we have to repeat that for ResetOffset.  So it doesn't end up feeling very clean to me.

I think this is just how the Sarama library works and we're dealing with it in a good way.  FWIW... This is the same way the Sarama ConsumerGroup Error handling works (separate async channel) and we haven't tried to do anything there to "hide" the nature of the library.

I would like to keep it this way for now - I don't think it's too confusing (not any worse than some buried/wrapped solution that tries to hide what the Sarama library is).   If it's a problem for some reason going forward we can look at changing it then?

pkg/common/commands/resetoffset/controller/reconciler.go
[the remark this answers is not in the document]
travis-minke-sap: I'll add a TODO comment to that logic indicating it should be re-evaluated and likely removed once we've stood up a webhook.  Until then... if I remove it we won't have the same functionality.  I'll update the parent issue with a note on this as well.

pkg/common/commands/resetoffset/controller/offsets.go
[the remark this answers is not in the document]
travis-minke-sap: **TLDR...**
Well, I take it all back - haha ; ).    Seems I wasn't doing a very good job of handling the Sarama errors.  Don't get me wrong, the Sarama API is still a wonky mess in this area but... my understanding of how it worked was lacking.  I have re-written most of the `offsets.go` file which should probably be reviewed as a new file (easier that way).

**The details...**
My confusion was based on the expectation that the Parent / Child nature of the OffsetManager & PartitionOffsetManager was "logical".  I was creating and Closing() the PartitionOffsetManager in the sub-loop, and was expecting the `MarkOffset()` and `ResetOffset()` calls to possibly generate errors on the `Errors()` channel.  After re-reading the docs for those functions and a fair amount of their source it seems that it's not that straightforward.

The OffsetManager and the PartitionOffsetManager are tightly intertwined and there are expectations about their usage.  The OffsetManager must be closed after the PartitionOffsetManagers, but the PartitionOffsetManager `Close()` will block on returning errors until the OffsetManager is closed (when autocommit is disabled).  Further the expectation is that you drain the PartitionOffsetManager `Errors()` channel AFTER closing the OffsetManager, etc, etc..

The core logic is mosty the same, just the order of things has changed.

Thanks in advance for re-reviewing!

pkg/common/commands/resetoffset/controller/offsets.go
```diff
@@ -88,38 +81,65 @@ func (r *Reconciler) reconcileOffsets(ctx context.Context, topicName string, gro
...
-
-		// Update The Partition's Offset (Process all partitions even on error in order to populate OffsetMappings for CRD)
-		offsetMapping, updateErr := updateOffset(partitionLogger, saramaClient, offsetManager, topicName, partition, offsetTime)
+		partitionOffsetManager, offsetMapping, updateErr := updateOffset(logger, saramaClient, offsetManager, topicName, partition, offsetTime)
+		if updateErr != nil {
+			updatesFailed = true
+		}
+		if partitionOffsetManager != nil {
+			partitionOffsetManagers[index] = partitionOffsetManager
+		}
```
eric-sap, line 104: Since the zero values of the partitionOffsetManagers are nil anyway, this "don't set it if it's nil" logic doesn't really accomplish anything.  Normally I wouldn't mention that here because it doesn't hurt anything, but since you are explicitly setting the value even if there was an error, I might argue that "actually" setting partitionOffsetManagers[index] to nil is the right (philosophically) thing to do here anyway.
```suggestion
		partitionOffsetManagers[index] = partitionOffsetManager
```

pkg/common/commands/resetoffset/controller/offsets.go
```diff
@@ -129,27 +149,31 @@ func updateOffset(logger *zap.Logger,
...
 	// Create A PartitionOffsetManager For The Specified Partition
 	partitionOffsetManager, err := offsetManager.ManagePartition(topic, partition)
-	if err != nil {
+	if partitionOffsetManager == nil || err != nil {
 		logger.Error("Failed to create PartitionOffsetManager", zap.Error(err))
-		closePartitionOffsetManager(logger, partitionOffsetManager)
-		return nil, err
+		if partitionOffsetManager != nil {
+			partitionOffsetManager.AsyncClose()
+		}
```
eric-sap, line 170: There's nothing technically wrong with this being here, but it seems odd for the caller to be responsible for closing an interface that was returned if there was an error also returned (which is the only way this AsynClose could be called).  I'm not sure I'd call for removing it as it doesn't hurt, but I am curious if there was there a path taken during testing that made this necessary?

pkg/common/commands/resetoffset/controller/offsets.go
```diff
@@ -192,42 +208,28 @@ func safeCloseSaramaClient(logger *zap.Logger, client sarama.Client) {
...
+					if multiErr == nil {
+						multiErr = consumerErr.Unwrap()
+					} else {
+						multierr.AppendInto(&multiErr, consumerErr.Unwrap())
+					}
+				}
+				if !ok {
+					break // Error Channel Closed - Stop Draining
+				}
+			case <-time.After(5 * time.Second):
```
eric-sap, line 229: This is giving each POM five full seconds before moving to the next one, right ?  I don't know how many partitions are likely to be in play here but that could add up to a fair bit of sleep-time if they're all timing out.  That may not actually matter but I thought I'd mention it.

eric-sap (commented):
Looks good; some trivial nits and questions but overall seems like a good fixup

pkg/common/commands/resetoffset/controller/offsets.go
travis-minke-sap: no, you're right it's not needed - will remove - thanks

pkg/common/commands/resetoffset/controller/offsets.go
travis-minke-sap: Yeah, that is the error/edge case and normal execution should hit the other case: statement.  I could reduce it slightly but seems safer to err on the side of caution here. going to leave as is - but good question.

pkg/common/commands/resetoffset/controller/offsets.go
travis-minke-sap: i can remove the if/nil check - as you noted they are identical.  The code went back and forth a few times between appending to a list vs default sized, etc.. - was just an artifact - good catch ; )

pkg/common/commands/resetoffset/controller/reconciler.go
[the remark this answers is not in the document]
matzew: > I'll add a TODO comment to that logic indicating it should be re-evaluated and likely removed once we've stood up a webhook. Until then... if I remove it we won't have the same functionality. I'll update the parent issue with a note on this as well.

I am fine to have this in, temporary. Let's capture it in a GH issue

pkg/common/commands/resetoffset/controller/offsets.go
```diff
@@ -0,0 +1,230 @@
...
+		return nil, err
+	}
+
+	// Get The Partitions Of The Specified Kafka Topic
+	partitions, err := saramaClient.Partitions(topicName)
+	if err != nil {
+		logger.Error("Failed to determine Partitions for Topic", zap.Error(err))
+		return nil, err
+	}
+	logger.Info("Found Topic Partitions", zap.Any("Partitions", partitions))
```
matzew, line 82: debug?

pkg/common/commands/resetoffset/controller/offsets.go
```diff
@@ -0,0 +1,230 @@
...
+	updatesFailed := false
+	for index, partition := range partitions {
+		partitionOffsetManager, offsetMapping, updateErr := updateOffset(logger, saramaClient, offsetManager, topicName, partition, offsetTime)
+		partitionOffsetManagers[index] = partitionOffsetManager
+		if offsetMapping != nil {
+			offsetMappings[index] = *offsetMapping
+		}
+		if updateErr != nil {
+			updatesFailed = true
+		}
```
matzew, line 105: isn't the `offsetMapping` only `nil`, where there was an error in the update?

matzew:
I find the `reconcileOffsets` a bit hard to read... but generally OK w/ this PR.

Perhaps any comments from @aliok ?

aliok:
I finally have time to review this PR. Give me some time please.

aliok:
I think the PR is good. I didn't get the `drainPartitionOffsetManagerErrors` part but I saw the discussion above and I don't have something to help.

LGTM

pkg/common/commands/resetoffset/controller/reconciler.go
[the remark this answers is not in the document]
travis-minke-sap: sure - created Issue #657

pkg/common/commands/resetoffset/controller/offsets.go
travis-minke-sap: Sure I can change it.  I tend towards noisy logs in Controllers as they are Control-Plane.  This one in particular will be used infrequently (downtime recovery and other edge cases).

pkg/common/commands/resetoffset/controller/offsets.go
travis-minke-sap: Yeah, this handling I think is part of the overall complexity problem. the new rewrite should be simpler.

pkg/common/commands/resetoffset/controller/offsets.go
```diff
@@ -88,104 +88,117 @@ func (r *Reconciler) reconcileOffsets(ctx context.Context, topicName string, gro
...
+	if err != nil {
+		logger.Error("Failed to update Offsets for Topic Partitions", zap.Error(err))
+		_ = closeManagersAndDrainErrors(logger, offsetManager, partitionOffsetManagers)
+		return nil, err
+	}
+
+	// Close The Sarama Managers And Get Any Accumulated Errors
+	err = closeManagersAndDrainErrors(logger, offsetManager, partitionOffsetManagers)
+	if err != nil {
+		logger.Error("")
```
eric-sap, line 110: I assume this is not intentionally left blank

pkg/common/commands/resetoffset/controller/offsets.go
```diff
@@ -88,104 +88,117 @@ func (r *Reconciler) reconcileOffsets(ctx context.Context, topicName string, gro
...
-		resultErr = updateOffsetsError
-	}
-	return offsetMappings, resultErr
+	// Return Success!
+	return offsetMappings, nil
 }
 
-// updateOffset performs the update of a single Partition's Offset and
-// returns an OffsetMapping representing the old/new state.
+// updateOffset calculated and performs an update of a single Partition's Offset
```
eric-sap, line 165: calculates

travis-minke-sap: 👍

pkg/common/commands/resetoffset/controller/offsets.go
travis-minke-sap: doh!

travis-minke-sap:
> I find the `reconcileOffsets` a bit hard to read... but generally OK w/ this PR.

No one liked the implementation in `offsets.go` so I made another attempt at it.  I think it is better (decomposed a bit) but do keep in mind that "some" of the prior and current complexity is inherently due to the Sarama library itself.  Specifically we need to do things in the following order which just complicates the implementation...

1. Create Sarama Client / OffsetManager / PartitionOffsetManagers
2. Calculate Old/New Offsets
3. MarkOffset() or ResetOffset()
4. Commit
5. AsyncClose PartitionOffsetManagers
6. Close OffsetManager
7. Drain PartitionOffsetManagers.Errors() channel
8. Close the Sarama Client

appreciate any reviews!    You can (should) focus on just the `offsets.go` file and it's probably easier to just open and review that file by itself instead of looking at the Github Diff ; )

pkg/common/commands/resetoffset/controller/reconciler.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	// TODO - This functionality should be handled by the Webhook (Admissions / Defaulting) - Verify & Remove once that is in place.
+	//      - Note: The "consolidated" and "distributed" KafkaChannel implementations have similar logic - possibly investigate why?
+	//      - Note: If we remove here and rely on Webhook then we are effectively requiring all users of the Controller to setup a Webhook?
+	// Set Defaults & Verify The ResetOffset Is Valid
+	resetOffset.SetDefaults(ctx)
+	if err := resetOffset.Validate(ctx); err != nil {
+		logger.Error("Received invalid ResetOffset", zap.Error(err))
+		return fmt.Errorf("received invalid ResetOffset: %v", err)
```
slinkydeveloper, line 73: FYI you don't need SetDefaults() too, I've saw it this morning while reading some of the code generated code and pkg/reconciler runtime code that `SetDefaults()` is invoked before `ReconcileKind` is invoked

travis-minke-sap: OK good to know - I tested it without this logic and it works ok assuming you're reconciling a valid CRD.  I will remove it for now.  I had also created an Issue to address this in the other controllers as Matthias suggested (Issue #657).

PR #635, opened 2021-05-19
Debug log consume message

eric-sap:
## Proposed Changes

- 🧽  Log (in debug mode) the response from DispatchMessageWithRetries() in the consumeMessage() call

Seeing the response from a dispatched message can be extremely useful, especially if the target URL doesn't have easily accessible logs.  Sample log output from this PR's code (omitting the surrounding JSON):

Ordinary 200 response, no response body:
```
Dispatching event
Response is a non event, discarding it
Dispatcher Response: 200 (20.6957ms)
```
500 response (target pod down), descriptive response body:
```
Dispatching event
Dispatcher Response: 500 (15.0306917s): dispatch error: Post "http://my-service.my-namespace.svc.cluster.local/": dial tcp 10.108.46.168:80: connect: connection refused
```
Ordinary 400 response, no response body:
```
Dispatching event
Dispatcher Response: 400 (8.6064ms)
```

### DIFF
--- a/pkg/channel/distributed/dispatcher/dispatcher/handler.go
+++ b/pkg/channel/distributed/dispatcher/dispatcher/handler.go
@@ -19,10 +19,9 @@ package dispatcher
 import (
 	"context"
 	"errors"
+	"fmt"
 	"net/url"
 	"strings"
-
-	kafkasarama "knative.dev/eventing-kafka/pkg/channel/distributed/common/kafka/sarama"
 
 	"github.com/Shopify/sarama"
 	kafkasaramaprotocol "github.com/cloudevents/sdk-go/protocol/kafka_sarama/v2"
@@ -32,6 +31,8 @@ import (
 	eventingduck "knative.dev/eventing/pkg/apis/duck/v1"
 	"knative.dev/eventing/pkg/channel"
 	"knative.dev/eventing/pkg/kncloudevents"
+
+	kafkasarama "knative.dev/eventing-kafka/pkg/channel/distributed/common/kafka/sarama"
 )
 
 // Verify The Handler Implements The Sarama ConsumerGroupHandler
@@ -130,8 +131,9 @@ func (h *Handler) ConsumeClaim(session sarama.ConsumerGroupSession, claim sarama
 // Consume A Single Message
 func (h *Handler) consumeMessage(context context.Context, consumerMessage *sarama.ConsumerMessage, destinationURL *url.URL, replyURL *url.URL, deadLetterURL *url.URL, retryConfig *kncloudevents.RetryConfig) error {
 
+	debug := h.Logger.Core().Enabled(zap.DebugLevel)
 	// Debug Log Kafka ConsumerMessage
-	if h.Logger.Core().Enabled(zap.DebugLevel) {
+	if debug {
 		// Checked Logging Level First To Avoid Calling StringifyHeaderPtrs In Production
 		h.Logger.Debug("Consuming Kafka Message",
 			zap.Any("Headers", kafkasarama.StringifyHeaderPtrs(consumerMessage.Headers)), // Log human-readable strings, not base64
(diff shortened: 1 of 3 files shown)


pkg/channel/distributed/dispatcher/dispatcher/handler.go
```diff
@@ -130,8 +131,9 @@ func (h *Handler) ConsumeClaim(session sarama.ConsumerGroupSession, claim sarama
 // Consume A Single Message
 func (h *Handler) consumeMessage(context context.Context, consumerMessage *sarama.ConsumerMessage, destinationURL *url.URL, replyURL *url.URL, deadLetterURL *url.URL, retryConfig *kncloudevents.RetryConfig) error {
 
+	debug := h.Logger.Core().Enabled(zap.DebugLevel)
 	// Debug Log Kafka ConsumerMessage
-	if h.Logger.Core().Enabled(zap.DebugLevel) {
+	if debug {
```
matzew, line 136: do we really need this check ? and instead do direct `logger.debug` ? 
e.g. https://github.com/knative-sandbox/eventing-kafka/blob/main/pkg/channel/consolidated/reconciler/controller/controller.go#L110

eric-sap: It's not usually necessary, but the goal for this particular use of the debug check is to prevent having to call functions like StringifyHeaderPtrs and formatDispatcherExecutionInfo when debug logging isn't enabled.  The zap.() functions create structs that are clever enough to not call the reflection functions if the log level isn't set, but, e.g. StringifyHeaderPtrs isn't; it would be called even when debug is not enabled, before the Logger.Debug call is made, which is a waste of time in consumeMessage, which is data-plane and execution time is (typically) worth the consideration.  I am open to doing this in a different way (although this particular check was here before this PR, I just moved it to a temp variable so I could use it before calling formatDispatcherExecutionInfo) but I don't want to call these debugging functions in ordinary (production) cases.

eric-sap: (I'd have used the zap.* functions instead but there are no zap structs for "convert these base64-encoded headers into unencoded text" or "truncate the body of the response if over 500 chars"  ...  although I haven't looked into whether it's possible to add custom zap fields - that might be interesting)

eric-sap: It looks like it is possible to add a zapcore.ObjectMarshaler for a type of your own (or a wrapper you make for a third-party object) but I don't know that doing that would add a lot of value here for one function.

pkg/channel/distributed/dispatcher/dispatcher/handler.go
```diff
@@ -153,6 +155,26 @@ func (h *Handler) consumeMessage(context context.Context, consumerMessage *saram
...
+	// The body for Dispatcher responses is usually empty, so don't bother printing it in those cases
+	if len(info.ResponseBody) == 0 {
+		return fmt.Sprintf("%v (%v)", info.ResponseCode, info.Time)
+	}
+	truncate := len(info.ResponseBody)
+	if truncate > 500 {
+		truncate = 500 // Responses aren't typically large, but don't dump a massive response body
+	}
+	return fmt.Sprintf("%v (%v): %v", info.ResponseCode, info.Time, string(info.ResponseBody[:truncate]))
+}
```
travis-minke-sap, line 180: seems fine - only wondering if it's worth trying to log in a more zap/json format for searching? - maybe a function that takes the execution info and makes a clone with truncated body and then use `zap.Any()` on that?  thoughts?

pkg/channel/distributed/dispatcher/dispatcher/handler.go
matzew: thanks for the context

/lgtm

leaving the actual approval for @travis-minke-sap

pkg/channel/distributed/dispatcher/dispatcher/handler.go
eric-sap: I like this idea; I made a little experiment with a custom zap field (which is similar in concept) and it looks a lot better.  I'll clean it up and commit it

travis-minke-sap (approved):
/lgtm - thanks eric!

Issue #411, opened 2021-02-19
Source controllers mark KafkaSource as ready too early


lionelvillard:
**Problem**
Events sent to a source marked as ready might not be forwarded to its sink. For instance, events sent to partitions with an uninitialized consumer group offset could be lost if these events are sent while the source is being scaled up/down.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)**
Which persona is this feature for?

**Exit Criteria**
A measurable (binary) test that would indicate that the problem has been resolved.

**Time Estimate (optional):**
How many developer-days do you think this may take to resolve?
1-2

**Additional context (optional)**
Add any other context about the feature request here.

PR #636, opened 2021-05-19
Some KafkaSource scaling rekt tests.

lionelvillard:
## Proposed Changes

- Add one scaling test sending event for 20s while scaling up the KafkaSource to 200 consumers. Check if there is no events loss.
-
-

**Release Note**

```release-note

```

**Docs**


test/test_images/kafka-proxy/main.go
```diff
@@ -0,0 +1,97 @@
+/*
```
aliok, line 1: Can you add a README file within this folder for this test image? What it does, how to configure it, etc.

Only very briefly as it is not a deliverable of the Knative project

lionelvillard: good idea! Will do.

test/rekt/features/kafkasource/scaling.go
```diff
@@ -0,0 +1,138 @@
...
+	// Setup:
+	// Sender (eventshub) -> KafkaProxy -> KafkaTopic -> KafkaSource -> Sink (eventshub)
+	proxyName := feature.MakeRandomK8sName("proxy")
+	f.Setup("set proxy name state", func(ctx context.Context, t feature.T) {
+		state.SetOrFail(ctx, t, "proxyName", proxyName)
+	})
+
+	f.Setup("install Kafka proxy", kafkaproxy.Install(proxyName,
+		kafkaproxy.WithBootstrapServer(kafkaBootstrapUrlPlain),
+		kafkaproxy.WithTopic(topic)))
```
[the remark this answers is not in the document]
lionelvillard: it's done here: https://github.com/knative-sandbox/eventing-kafka/blob/c20df2dbbcfc35242f50dd2069fea3ca3f5a63cf/test/rekt/resources/kafkaproxy/kafkaproxy.go#L54

However it does not wait for the endpoint to be ready, only the pod, which is wrong. I will fix it. Thanks for pointing me to `k8s.WaitForServiceEndpointsOrFail`!

test/rekt/features/kafkasource/scaling.go
```diff
@@ -0,0 +1,138 @@
...
+	)(ctx, t)
+
+	time.Sleep(2 * time.Second)
+
+	env := environment.FromContext(ctx)
+	source, err := kafkaclient.Get(ctx).SourcesV1beta1().KafkaSources(env.Namespace()).Get(ctx, sourceName, metav1.GetOptions{})
+	if err != nil {
+		t.Fatalf("failed to get the KafkaSource object: %v", err)
+	}
+	source.Spec.Consumers = pointer.Int32Ptr(130)
```
aliok, line 128: Just a random number?

aliok:
@lionelvillard this test is to make sure the problem https://github.com/knative-sandbox/eventing-kafka/issues/411 is fixed? What's the relation between them?

test/rekt/features/kafkasource/scaling.go
lionelvillard: not totally. It has to be over 100 because each pod can have 100 "virtual replicas".

lionelvillard:
> @lionelvillard this test is to make sure the problem #411 is fixed? What's the relation between them?

Yup

lionelvillard:
no more rekt. Will do something else

PR #640, opened 2021-05-20
start to remove vreplicas from biggest ordinal

### LINKED ISSUE #628: [mtsource] Remove vreplicas starting from adapter pods with the biggest ordinal
**Describe the bug**
When KEDA scales vreplicas down, the extra vreplicas were removed from the wrong adapter pod. 

 **Expected behavior**
The extra vreplicas should be removed from adapter pods with the biggest ordinal first 

 **To Reproduce**
Create a Kafka source that spans over 2 pods and then stop sending events to the topic, when KEDA scales the vreplica back to 1, you will see the Kafka source ends up on the adapter pod that has the biggest ordinal number. 

 **Knative release version**

 **Additional context**

steven0711dong:
close https://github.com/knative-sandbox/eventing-kafka/issues/628

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/common/scheduler/statefulset/scheduler.go
+++ b/pkg/common/scheduler/statefulset/scheduler.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"math"
 	"sort"
+	"strconv"
 	"sync"
 	"time"
 
@@ -122,6 +123,19 @@ func NewStatefulSetScheduler(ctx context.Context,
 }
 
 func (s *StatefulSetScheduler) Schedule(vpod scheduler.VPod) ([]duckv1alpha1.Placement, error) {
+	placements, err := s.scheduleVPod(vpod)
+	if placements == nil {
+		return placements, err
+	}
+	sort.SliceStable(placements, func(i int, j int) bool {
+		podIOrdinal, _ := strconv.Atoi(placements[i].PodName[len(placements[i].PodName)-1:])
+		podJOrdinal, _ := strconv.Atoi(placements[j].PodName[len(placements[j].PodName)-1:])
+		return podIOrdinal < podJOrdinal
+	})
+	return placements, err
+}
+
+func (s *StatefulSetScheduler) scheduleVPod(vpod scheduler.VPod) ([]duckv1alpha1.Placement, error) {
 	s.lock.Lock()
 	defer s.lock.Unlock()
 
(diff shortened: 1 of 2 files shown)


pkg/source/reconciler/mtsource/kafkasource.go
```diff
@@ -108,7 +110,11 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, src *v1beta1.KafkaSource
 
 func (r *Reconciler) reconcileMTReceiveAdapter(src *v1beta1.KafkaSource) error {
 	placements, err := r.scheduler.Schedule(src)
-
+	sort.SliceStable(placements, func(i int, j int) bool {
+		podIOrdinal, _ := strconv.Atoi(placements[i].PodName[len(placements[i].PodName)-1:])
+		podJOrdinal, _ := strconv.Atoi(placements[j].PodName[len(placements[j].PodName)-1:])
+		return podIOrdinal < podJOrdinal
+	})
```
lionelvillard, line 117: `Schedule` should return a sorted list on placements as it relies on it.

steven0711dong: The current logic of addVReplica is adding `vreplicas` to existing pod, then allocate additional pods. In case existing pod is not a sequence, it's possible scheduler adds `vreplicas` to a pod that's in between existing pods. In this scenario, Schedule won't return a sorted list.

pkg/common/scheduler/scheduler.go
```diff
@@ -35,6 +35,7 @@ type VPodLister func() ([]VPod, error)
 type Scheduler interface {
 	// Schedule computes the new set of placements for vpod.
 	Schedule(vpod VPod) ([]duckv1alpha1.Placement, error)
+	ScheduleVPod(vpod VPod) ([]duckv1alpha1.Placement, error)
```
lionelvillard, line 38: Let's keep the interface minimal. No need to expose ScheduleVPod

steven0711dong: I tried to change `ScheduleVPod` to `scheduleVPod` but there is compiler error complaining cannot use `StatefulSetScheduler` as interface Scheduler. I think it has something to do with `StatefulSetScheduler` and Scheduler interface being defined in different package so we need to make `scheduleVPod` exported.

lionelvillard: I suggest to remove `ScheduleVPod` altogether from the interface.

steven0711dong: If we remove ScheduleVPod, how do you propose to sort the placements?

lionelvillard (approved):

PR #646, opened 2021-05-24
Set global maxresponsesize

### LINKED ISSUE #645: [mtsource] Revert buffersize calculation
**Problem**
A short explanation of the problem, including relevant restrictions.
The mtsource adapter does some magic to not exceed the memory limit. This magic is not needed (see [KIP-74](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=65864887))

**Exit Criteria**

**Time Estimate (optional):**

lionelvillard:
Fixes #645

## Proposed Changes

- Rely on MaxResponseSize to ensure the client does not run OOM.
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/config/source/multi/deployments/adapter.yaml
+++ b/config/source/multi/deployments/adapter.yaml
@@ -57,7 +57,7 @@ spec:
         # DO NOT MODIFY: The values below are being filled by the kafka source controller
         # See 500-controller.yaml
         - name: K_METRICS_CONFIG
-          value: '{"Domain":"knative.dev/eventing","Component":"kafkasource","ConfigMap":{"profiling.enable": "true"}}'
+          value: '{"Domain":"knative.dev/eventing","Component":"kafkasource","ConfigMap":{"profiling.enable": "false"}}'
         - name: K_LOGGING_CONFIG
           value: '{}'
         - name: K_LEADER_ELECTION_CONFIG
@@ -68,10 +68,10 @@ spec:
         resources:
           requests:
             cpu: 100m
-            memory: 100Mi
+            memory: 250Mi
           limits:
             cpu: 200m
-            memory: 150Mi
+            memory: 300Mi
 
         ports:
         - name: metrics
--- a/pkg/source/mtadapter/adapter.go
+++ b/pkg/source/mtadapter/adapter.go
@@ -65,6 +65,7 @@ type Adapter struct {
 	client      cloudevents.Client
 	adapterCtor adapter.MessageAdapterConstructor
 	kubeClient  kubernetes.Interface
+	memLimit    int32
 
 	sourcesMu sync.RWMutex
 	sources   map[string]cancelContext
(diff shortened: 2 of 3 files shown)


config/source/multi/deployments/adapter.yaml
```diff
@@ -57,7 +57,7 @@ spec:
         # DO NOT MODIFY: The values below are being filled by the kafka source controller
         # See 500-controller.yaml
         - name: K_METRICS_CONFIG
-          value: '{"Domain":"knative.dev/eventing","Component":"kafkasource","ConfigMap":{"profiling.enable": "true"}}'
+          value: '{"Domain":"knative.dev/eventing","Component":"kafkasource","ConfigMap":{"profiling.enable": "false"}}'
```
steven0711dong, line 60: what is profiling.enable for? Is it for sarama logging?

pkg/source/mtadapter/adapter.go
```diff
@@ -81,15 +79,19 @@ func NewAdapter(ctx context.Context, processed adapter.EnvConfigAccessor, ceClie
 func newAdapter(ctx context.Context, processed adapter.EnvConfigAccessor, ceClient cloudevents.Client, adapterCtor adapter.MessageAdapterConstructor) adapter.Adapter {
 	logger := logging.FromContext(ctx)
 	config := processed.(*AdapterConfig)
+
 	ml := resource.MustParse(config.MemoryLimit)
+	if ml.Value() > 0 {
+		logger.Infof("Setting MaxResponseSize to %d bytes", int32(ml.Value()))
+		sarama.MaxResponseSize = int32(ml.Value())
```
steven0711dong, line 86: Reading the doc, I saw this: 
```
The implementation on the server side would grab data from each partition in the fetch request until it hit this limit, then send back just the data for the partitions that fit in the response. The implementation would need to start from a random position in the list of topics included in the fetch request to ensure that in a case of backlog we fairly balance between partitions (to avoid first giving just the first partition until that is exhausted, then the next partition, etc).
```
It seems like the this MaxResponseSize is for all the messages in different partitions the broker returns all at once. Am I understanding this correctly? Currently this value comes from `VREPLICA_LIMITS_MEMORY`, which is set at only 2Mi. Should this number be changed as well? 2Mi seems very small.

config/source/multi/deployments/adapter.yaml
lionelvillard: it's for enabling Golang profiling. Shouldn't be enabled by default.

pkg/source/mtadapter/adapter.go
lionelvillard: Yes that's my understanding as well. MaxResponseSize should be VREPLICA_LIMITS_MEMORY * # of vreplicas. Fixing.. Thanks for catching this!

config/source/multi/deployments/adapter.yaml
```diff
@@ -68,10 +68,10 @@ spec:
...
           requests:
             cpu: 100m
-            memory: 100Mi
+            memory: 250Mi
           limits:
             cpu: 200m
-            memory: 150Mi
+            memory: 300Mi
 
         ports:
```
matzew, line 76: relevant for the "single" too?

lionelvillard: the single does not set any requests/limits.

steven0711dong (approved):
LGTM

PR #649, opened 2021-05-24
Mark KafkaSource as ready when all initial offsets have been committed

### LINKED ISSUE #411: Source controllers mark KafkaSource as ready too early

lionelvillard:
Fixes #411

## Proposed Changes

- Move InitOffsets out of the ST adapter to the source controllers
- Add two new conditions in KafkaSourceStatus: ConnectionEstablished, InitialOffsetsCommitted. 

**Release Note**

```release-note
🎁 KafkaSource is marked as ready when the connection to Kafka has been established and when all consumer group offsets have been initialized.
```

**Docs**


pkg/source/mtadapter/adapter.go
```diff
@@ -183,54 +183,11 @@ func (a *Adapter) Update(ctx context.Context, obj *v1beta1.KafkaSource) error {
...
-		return err
-	}
-
-	tlsCert, err := a.ResolveSecret(ctx, obj.Namespace, obj.Spec.Net.TLS.Cert.SecretKeyRef)
-	if err != nil {
-		return err
-	}
-
-	tlsKey, err := a.ResolveSecret(ctx, obj.Namespace, obj.Spec.Net.TLS.Key.SecretKeyRef)
-	if err != nil {
```
matzew, line 207: isn't there a way to have the adapters mostly similar? Not sure I like they seem to diverge

matzew: eg. have come "common adapter" ?

lionelvillard: the mtadapter is built on top of the stadapter so there is no real need for a common adapter. This code is not needed by the stadapter since secrets are resolved by k8s itself.

pkg/source/mtadapter/adapter.go
```diff
@@ -133,54 +133,11 @@ func (a *Adapter) Update(ctx context.Context, obj *v1beta1.KafkaSource) error {
 		return nil
 	}
 
-	saslUser, err := a.ResolveSecret(ctx, obj.Namespace, obj.Spec.Net.SASL.User.SecretKeyRef)
```
steven0711dong, line 136: should we also remove ResolveSecret defined in adapter.go?

lionelvillard: good catch! Thanks!

steven0711dong (approved):
/lgtm

PR #668, opened 2021-05-28
Adjust MaxResponseSize based on the number of active stadapters

lionelvillard:
## Proposed Changes

- MaxResponseSize is now Container Memory request / # of source instances in container. Cap at 50MB 
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/source/mtadapter/adapter.go
+++ b/pkg/source/mtadapter/adapter.go
@@ -42,6 +42,10 @@ import (
 	"knative.dev/eventing-kafka/pkg/source/client"
 )
 
+const (
+	responseSizeCap = 50 * 1024 * 1024
+)
+
 type AdapterConfig struct {
 	adapter.EnvConfig
 
@@ -82,8 +86,9 @@ func newAdapter(ctx context.Context, processed adapter.EnvConfigAccessor, ceClie
 	logger := logging.FromContext(ctx)
 	config := processed.(*AdapterConfig)
 
-	mr := resource.MustParse(config.MemoryRequest)
+	sarama.MaxResponseSize = responseSizeCap
 
+	mr := resource.MustParse(config.MemoryRequest)
 	return &Adapter{
 		client:        ceClient,
 		config:        config,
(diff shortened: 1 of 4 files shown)


pkg/source/mtadapter/adapter.go
```diff
@@ -279,3 +286,24 @@ func (a *Adapter) ResolveSecret(ctx context.Context, ns string, ref *corev1.Secr
 	a.logger.Errorw("missing secret key or empty secret value", zap.String("secretname", ref.Name), zap.String("secretkey", ref.Key))
 	return "", fmt.Errorf("missing secret key or empty secret value (%s/%s)", ref.Name, ref.Key)
 }
+
+// adjustResponseSize ensures the sum of all Kafka clients memory usage does not exceed the container memory request.
+func (a *Adapter) adjustResponseSize() {
```
steven0711dong, line 291: Is this the best of what we can do? We should prefer to allocate memory resources to Kafka sources based on the number of vreplicas per Kafka source.

lionelvillard: Do you have a suggestion?

pkg/source/mtadapter/adapter.go
```diff
@@ -279,3 +286,24 @@ func (a *Adapter) ResolveSecret(ctx context.Context, ns string, ref *corev1.Secr
...
+		maxResponseSize := int32(float64(a.memoryRequest) / float64(len(a.sources)))
+
+		// cap the response size to 50MB.
+		if maxResponseSize > responseSizeCap {
+			maxResponseSize = 50 * 1024 * 1024
+		}
+		// Check for compliance.
+		if maxResponseSize < 64*1024 {
+			// Not CloudEvent compliant (https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#size-limits)
+			a.logger.Warnw("Kafka response size is lower than 64KB. Increase the pod memory request and/or lower the pod capacity.",
```
matzew, line 302: would we need it on a more common place, for both variants ?

pkg/source/mtadapter/adapter.go
steven0711dong: not really. I thought about setting consumer.fetch.max based to the number of partitions handled by the pod in total but this seems not a good way. Once a partition from any Kafka source joins or leaves, it affects all the Kafka source.

steven0711dong (approved):
LGTM

PR #682, opened 2021-06-02
[mtsource] MaxResponseSize considers the number of connected brokers

lionelvillard:
https://github.com/knative-sandbox/eventing-kafka/pull/668 followup.

This is a quick fix until a proper solution is implemented.

## Proposed Changes

- MaxResponseSize is per broker so lower its value based on the potential number of connected brokers (set to 6)
- 
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/source/mtadapter/adapter.go
+++ b/pkg/source/mtadapter/adapter.go
@@ -43,7 +43,8 @@ import (
 )
 
 const (
-	responseSizeCap = 50 * 1024 * 1024
+	responseSizeCap = int32(50 * 1024 * 1024)
+	maxBrokers      = 6
 )
 
 type AdapterConfig struct {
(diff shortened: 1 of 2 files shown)


pkg/source/mtadapter/adapter.go
```diff
@@ -290,12 +291,21 @@ func (a *Adapter) ResolveSecret(ctx context.Context, ns string, ref *corev1.Secr
...
+		}
 		// cap the response size to 50MB.
 		if maxResponseSize > responseSizeCap {
 			maxResponseSize = 50 * 1024 * 1024
 		}
+
+		// maxResponseSize is per connected brokers.
+		// For now cap the number of connected brokers to maxBrokers until a better solution
+		// comes along
+		maxResponseSize = int32(float32(maxResponseSize) / float32(maxBrokers))
```
steven0711dong, line 307: `maxBrokers` is not fixed at 6. If someone uses a Kafka instance with 12 brokers, then this would be a problem. I guess for now, we can hard code it as 6. But this value is really per Kafka source. It should go onto the CR.

steven0711dong (approved):
LGTM

pkg/source/mtadapter/adapter.go
matzew: +1

PR #659, opened 2021-05-26
Make sure ConsumeClaim exit before rebalance.timeout

lionelvillard:
Fixes #629

## Proposed Changes

-  Cancel dispatching messages when consumer group session is closed 

**Release Note**

```release-note
- 🐛Fix a bug in consolidate channel where deleting a KafkaChannel-backed Broker with events in flight blocks new subscriptions in SubscriptionNotMarkedReadyByChannel . 
- 🐛Fix a bug in KafkaSource potentially sending duplicates when Kafka rebalancing occurs
```

**Docs**

PR #685, opened 2021-06-04
Revert test skip for subscription readiness

matzew:
This reverts commit 15d22560efc769fe272db010a4c5a07139cb2768.

Fixes #

## Proposed Changes

- undoing #573 since this should be better that #659 is in

**Release Note**

```release-note

```

**Docs**


matzew:
@lberk @devguyio even with the latest changes, we are not able to remove the `skip` of this one test :disappointed:

PR #694, opened 2021-06-08
Revert to setting consumer.fetch options instead of MaxReponseSize

### LINKED ISSUE #691: [mtsource] mtadapter crashes due to OOM
**Describe the bug**
When the mtadapter starts it sometimes enters in the crashloop due to OOM. This happens when there are many queued messages in a topic, leading to the input buffers to exceed the container memory limit. 

MaxResponseSize was used to mitigate this issue but its scope is per Fetch Request, which when the consumer group starts, the number of Fetch Requests can somewhere between 1 to number of partitions, until all partitions are assigned to a broker, in which case the number of Fetch requests stabilizes to the number of brokers. 

 **Expected behavior**

No crash

 **To Reproduce**
Create a KafkaSource, Produce event to a topic. Delete the KafkaSource and wait some time for the topic to be filled with messages. Recreate the KafkaSource, using the same consumergroup and observe the crash.

 **Knative release version**

 **Additional context**

lionelvillard:
Fixes #691

## Proposed Changes

- Revert https://github.com/knative-sandbox/eventing-kafka/pull/668/files as MaxResponseSize does not work as expected (see #691 for an explanation)
- Beef up the adapter pod to be able to handle larger messages. Not a definitive number, we can adjust as we gather more data.
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/source/mtadapter/adapter.go
+++ b/pkg/source/mtadapter/adapter.go
@@ -19,9 +19,10 @@ package mtadapter
 import (
 	"context"
 	"fmt"
+	"math"
+	"strconv"
 	"sync"
 
-	"github.com/Shopify/sarama"
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 	"go.uber.org/zap"
 	"golang.org/x/time/rate"
@@ -42,17 +43,12 @@ import (
 	"knative.dev/eventing-kafka/pkg/source/client"
 )
 
-const (
-	responseSizeCap = int32(50 * 1024 * 1024)
-	maxBrokers      = 6
-)
-
 type AdapterConfig struct {
 	adapter.EnvConfig
 
-	PodName       string `envconfig:"POD_NAME" required:"true"`
-	MPSLimit      int    `envconfig:"VREPLICA_LIMITS_MPS" required:"true"`
-	MemoryRequest string `envconfig:"REQUESTS_MEMORY" required:"true"`
+	PodName     string `envconfig:"POD_NAME" required:"true"`
+	MPSLimit    int    `envconfig:"VREPLICA_LIMITS_MPS" required:"true"`
+	MemoryLimit string `envconfig:"VREPLICA_LIMITS_MEMORY" required:"true"`
 }
 
 func NewEnvConfig() adapter.EnvConfigAccessor {
@@ -70,8 +66,7 @@ type Adapter struct {
 	client      cloudevents.Client
 	adapterCtor adapter.MessageAdapterConstructor
 	kubeClient  kubernetes.Interface
-
-	memoryRequest int64
+	memLimit    int32
 
 	sourcesMu sync.RWMutex
 	sources   map[string]cancelContext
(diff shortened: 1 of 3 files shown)


matzew (approved):
/lgtm
/approve

pkg/source/mtadapter/adapter.go
```diff
@@ -288,32 +302,55 @@ func (a *Adapter) ResolveSecret(ctx context.Context, ns string, ref *corev1.Secr
...
+
+	logger.Infow("partition count",
+		zap.Int("total", totalPartitions),
+		zap.Int("averagePerPod", partitionsPerPod),
+		zap.Int("handled", handledPartitions))
+
+	// A partition consumes about 2 * fetch partition size
+	// Once by FetchResponse blocks and a second time when those blocks are converted to messages
+	// see https://github.com/Shopify/sarama/blob/83d633e6e4f71b402df5e9c53ad5c1c334b7065d/consumer.go#L649
+	return int(math.Floor(float64(a.memLimit) / float64(handledPartitions) / 2.0)), nil
```
steven0711dong, line 355: What does a.memLimit look like? Is it going to be '6.0'?

pkg/source/mtadapter/adapter.go
```diff
@@ -288,32 +302,55 @@ func (a *Adapter) ResolveSecret(ctx context.Context, ns string, ref *corev1.Secr
...
-		if maxResponseSize < 64*1024 {
-			// Not CloudEvent compliant (https://github.com/cloudevents/spec/blob/v1.0.1/spec.md#size-limits)
-			a.logger.Warnw("Kafka response size is lower than 64KB. Increase the pod memory request and/or lower the pod capacity.",
-				zap.Int32("responseSize", maxResponseSize))
-		}
+	partitionsPerPod := int(math.Ceil(float64(totalPartitions) / float64(podCount)))
 
-		sarama.MaxResponseSize = maxResponseSize
-		a.logger.Infof("Setting MaxResponseSize to %d bytes", sarama.MaxResponseSize)
+	// Ideally, partitions are evenly spread across Kafka consumers.
```
steven0711dong, line 338: Does Kafka consumer here refer to Kafka client?

pkg/source/mtadapter/adapter.go
lionelvillard: it's an int and the unit is bytes. So more like 6291456

pkg/source/mtadapter/adapter.go
lionelvillard: yup

pkg/source/mtadapter/adapter.go
```diff
@@ -288,32 +302,55 @@ func (a *Adapter) ResolveSecret(ctx context.Context, ns string, ref *corev1.Secr
...
+	// However, due to rebalancing or consumer (un)availability, a consumer
+	// might have to handle more partitions than expected.
+	// For now, account for 1 unavailable consumer.
+	handledPartitions := 2 * partitionsPerPod
```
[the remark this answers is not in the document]
lionelvillard: because they are semantically different.

PR #696, opened 2021-06-08
do not print out  users' apikey

steven0711dong:
Fixes #

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/common/client/config.go
+++ b/pkg/common/client/config.go
@@ -253,7 +253,6 @@ func (b *configBuilder) Build(ctx context.Context) (*sarama.Config, error) {
 				config.Net.SASL.Mechanism = sarama.SASLTypeSCRAMSHA512
 			}
 			config.Net.SASL.User = b.auth.SASL.User
-			config.Net.SASL.Password = b.auth.SASL.Password
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/common/client/config.go
```diff
@@ -265,9 +264,6 @@ func (b *configBuilder) Build(ctx context.Context) (*sarama.Config, error) {
 		config.ClientID = b.clientId
 	}
 
-	logger := logging.FromContext(ctx)
-	logger.Infof("Built Sarama config: %+v", config)
```
aliok, line 269: Can't we have the message without the config?

matzew: yeah, would be handy - and perhaps... in DEBUG ? :smile:

aliok: Ok, then, and info message that the config is built, but the actual config in a debug message?

lionelvillard: even in DEBUG I don't think we want to print the user apikeys. What about printing the config before the apikey is set? Like maybe set it first to `XXX` and then substitute after printing the config? 

A better solution would be the override the String() function (haven't tried but should work)

steven0711dong: I can try override the String function

lionelvillard (approved):

PR #700, opened 2021-06-09
Remove creation of new dispatcher when secret changes

eric-sap:
This isn't directly tied to an issue, but the changes for #537 made a few things in the distributed channel's "reconfigure" function obsolete; in particular it used to be called when the configmap changed as well, which could require a new dispatcher.  Now the only possible changes are from the secret, which is limited to username/password/SASLtype.  Changing those settings does not require a new dispatcher, so this PR will remove that behavior.

I would leave this alone but there is also a bug in the way the new dispatcher is created which leaves the old dispatcher referenced in the dispatcher's controller.  The controller.NewController call uses the original dispatcher and nothing ever changes that, which is something that will cause a problem with some upcoming changes that include a "consumer group manager" that maintains some state information about consumer groups.  This manager can't have "a different manager" in an abandoned dispatcher elsewhere, so it needs to be fixed.

## Proposed Changes

- 🧽  Changing the kafka-cluster secret no longer creates a new distributed dispatcher; instead simply reconfigures the existing one
- 🐛 Removing (vs. changing) authentication by changing the kafka-cluster secret had no effect; now it removes the auth properly (dispatcher and receiver)
- 🧽  merged the "reconfigure" functions from dispatcher and receiver into the SecretChanged functions (since they were only being called once now that there is no longer ConfigMap-watching functionality in these components)

### DIFF
--- a/cmd/channel/distributed/dispatcher/main.go
+++ b/cmd/channel/distributed/dispatcher/main.go
@@ -209,24 +209,19 @@ func flush(logger *zap.Logger) {
 
 // secretObserver is the callback function that handles changes to our Secret
 func secretObserver(ctx context.Context, secret *corev1.Secret) {
-	logger := logging.FromContext(ctx)
+	secretLogger := logging.FromContext(ctx)
 
 	if secret == nil {
-		logger.Warn("Nil Secret passed to secretObserver; ignoring")
+		secretLogger.Warn("Nil Secret passed to secretObserver; ignoring")
 		return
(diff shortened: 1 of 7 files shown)


travis-minke-sap:
Looks good to me - leaving open for others to review.  Changes are primarily in distributed channel with only a small change in "common" code.  I will likely merge Thursday 6/10 morning US time so get any review comments in prior to that ; )

PR #705, opened 2021-06-14
[scheduler] Compact vreplicas

### LINKED ISSUE #699: [scheduler] mtadapters may not scale down
**Describe the bug**
The scheduler does not reassign vreplicas to "lower" pods  (ie: with a lower ordinal), preventing the autoscaler to scale down the mtadapter. 

 **Expected behavior**
The mtadapter should be able to scale down when there are unused pods.

 **To Reproduce**
Create 1 KafkaSource CR and scale it to 150. In knative-eventing, observe 2 adapter replicas. Create a second KafkaSource CR. Scale down the first KafkaSource CR to 1. The total number of vreplicas is now 2, spread across the 2 adapter replicas. In theory only one adapter is needed. 

 **Knative release version**

 **Additional context**

lionelvillard:
Fixes #699

## Proposed Changes

- The statefulset autoscaler now attempts to compact vreplicas (MaxFillUp strategy only)
-
-

**Release Note**

```release-note
🎁 The multi-tenant KafkaSource controller periodically attempts to relocate KafkaSource consumers to Statefulset pods with a lower ordinal
```

**Docs**

### DIFF
--- a/pkg/common/scheduler/scheduler.go
+++ b/pkg/common/scheduler/scheduler.go
@@ -31,6 +31,9 @@ var (
 // VPodLister is the function signature for returning a list of VPods
 type VPodLister func() ([]VPod, error)
 
+// Evictor allows for vreplicas to be evicted
+type Evictor func(vpod VPod, from *duckv1alpha1.Placement) error
+
 // Scheduler is responsible for placing VPods into real Kubernetes pods
 type Scheduler interface {
 	// Schedule computes the new set of placements for vpod.
--- a/pkg/common/scheduler/scheduler.go
+++ b/pkg/common/scheduler/scheduler.go
@@ -31,7 +31,9 @@ var (
 // VPodLister is the function signature for returning a list of VPods
 type VPodLister func() ([]VPod, error)
 
-// Evictor allows for vreplicas to be evicted
+// Evictor allows for vreplicas to be evicted.
+// For instance, the evictor is used by the statefulset scheduler to
+// move vreplicas to pod with a lower ordinal.
 type Evictor func(vpod VPod, from *duckv1alpha1.Placement) error
 
 // Scheduler is responsible for placing VPods into real Kubernetes pods
--- a/pkg/common/scheduler/statefulset/autoscaler.go
+++ b/pkg/common/scheduler/statefulset/autoscaler.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	clientappsv1 "k8s.io/client-go/kubernetes/typed/apps/v1"
 
+	"knative.dev/eventing-kafka/pkg/common/scheduler"
 	kubeclient "knative.dev/pkg/client/injection/kube/client"
 	"knative.dev/pkg/logging"
 )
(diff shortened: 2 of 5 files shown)


aavarghese:
/lgtm /approve
Tried it out - all good!!

aavarghese (approved):

pkg/source/reconciler/mtsource/controller.go
```diff
@@ -73,7 +81,53 @@ func NewController(
...
+		after := before.DeepCopy()
+
+		bp := after.GetPlacements()
+		ap := make([]duckv1alpha1.Placement, 0, len(bp)-1)
+		for _, p := range bp {
+			if p.PodName != from.PodName {
+				ap = append(ap, p)
+			}
+		}
+		after.Status.Placement = ap
```
steven0711dong, line 102: 1. If we have a Kafka source on the last pod with one replica and this Kafka source has only one vreplica. What would happen? 
2. Once we remove the replica from the last pod, what if KEDA scaled the replica again? Almost seems like KEDA and this logic works in the opposite direction.

lionelvillard: 1. if there are enough free capacity before the last pod, the vreplica will be moved
2. Keda is about increasing/decreasing the number of vreplicas. But I see what you mean. When scaling down, the vreplicas closest to the last pod are first removed (after your fix) so from this PoV Keda and this PR are working in the same direction.

pkg/common/scheduler/statefulset/autoscaler.go
```diff
@@ -155,7 +162,60 @@ func (a *autoscaler) doautoscale(ctx context.Context, attemptScaleDown bool, pen
...
+		return err
+	}
+
+	for _, vpod := range vpods {
+		placements := vpod.GetPlacements()
+		for _, placement := range placements {
+			ordinal := ordinalFromPodName(placement.PodName)
+
+			if ordinal == s.lastOrdinal {
+				a.logger.Infow("evicting vreplica(s)", zap.String("podname", placement.PodName), zap.Int("vreplicas", int(placement.VReplicas)))
```
steven0711dong, line 211: please also log the name of the Kafka source being evicted as well.

lionelvillard: done!

pkg/common/scheduler/statefulset/autoscaler.go
```diff
@@ -42,9 +43,11 @@ type Autoscaler interface {
 type autoscaler struct {
 	statefulSetClient clientappsv1.StatefulSetInterface
 	statefulSetName   string
+	vpodLister        scheduler.VPodLister
 	logger            *zap.SugaredLogger
 	stateAccessor     stateAccessor
 	trigger           chan int32
+	evict             scheduler.Evictor
```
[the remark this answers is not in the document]
lionelvillard: done.

pkg/common/scheduler/scheduler.go
```diff
@@ -31,6 +31,9 @@ var (
 // VPodLister is the function signature for returning a list of VPods
 type VPodLister func() ([]VPod, error)
 
+// Evictor allows for vreplicas to be evicted
```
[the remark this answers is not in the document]
lionelvillard: I added an example of how this can be used. Thanks!

aavarghese (approved):
/lgtm

PR #707, opened 2021-06-14
[scheduler, mtsource] Even Spread across Nodes in each zone for HA scheduler

### LINKED ISSUE #593: [scheduler,mtsource] Implementing EvenSpread HA support across nodes within a zone
**Problem**
Implementing HA across pods within a zone (# of pods is not static unlike # of zones, so spreading equally may also mean rebalancing)

More info in https://github.com/knative-sandbox/eventing-kafka/pull/587, https://github.com/knative-sandbox/eventing-kafka/issues/463

**Exit Criteria**
Vreplicas are evenly spread across pods in each zone (check kafkasource status placement field)

**Time Estimate (optional):**

aavarghese:
Fixes #593 
🎁  Implementing HA across nodes within a zone as a new scheduling strategy 
```        
        - name: SCHEDULER_POLICY_TYPE
          value: 'EVENSPREAD_BYNODE'
```
Continuation of https://github.com/knative-sandbox/eventing-kafka/pull/587

## Proposed Changes

- Controller has another new env var to control "scheduler policy type" called  'EVENSPREAD_BYNODE',
- "scheduler policy type" is an enum with three values MaxFillup, EvenSpread and EvenSpreadByNode. For EvenSpreadByNode, scheduler places vreplicas for HA by spreading replicas equally across nodes within a zone. Within a node, vreplicas are filled up in each pod upto capacity.
- Placement type in KafkaSource status has a new type for NodeName when EvenSpreadByNode scheduling
- HA controls are also shared with autoscaler, and state accessor
- State accessor computes the number of nodes in cluster and stores a map of node name to zone info

**Release Note**

```
New SCHEDULER_POLICY_TYPE called 'EVENSPREAD_BYNODE' for a High Availability scheduler that uniformly distributes vreplicas across all nodes in each zone to reduce impact of failure when a node or zone fails.
```

### DIFF
--- a/config/source/multi/deployments/controller.yaml
+++ b/config/source/multi/deployments/controller.yaml
@@ -55,7 +55,7 @@ spec:
 
         # The scheduling policy type for placing vreplicas on pods (see type SchedulerPolicyType for enum list)
         - name: SCHEDULER_POLICY_TYPE
-          value: 'MAXFILLUP'
+          value: 'EVENSPREAD_BYNODE'
 
         resources:
           requests:
--- a/config/source/multi/deployments/controller.yaml
+++ b/config/source/multi/deployments/controller.yaml
@@ -57,6 +57,9 @@ spec:
         - name: SCHEDULER_POLICY_TYPE
           value: 'EVENSPREAD_BYNODE'
 
+        - name: CONFIG_SCHEDULER
+          value: config-scheduler
+
         resources:
           requests:
             cpu: 20m
--- a/pkg/common/scheduler/statefulset/scheduler.go
+++ b/pkg/common/scheduler/statefulset/scheduler.go
@@ -45,10 +45,12 @@ import (
 type SchedulerPolicyType string
 
 const (
-	// MAXFILLUP policy type adds replicas to existing pods to fill them up before adding to new pods
+	// MAXFILLUP policy type adds vreplicas to existing pods to fill them up before adding to new pods
 	MAXFILLUP SchedulerPolicyType = "MAXFILLUP"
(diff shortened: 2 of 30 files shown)


config/source/multi/deployments/controller.yaml
```diff
@@ -55,7 +55,7 @@ spec:
 
         # The scheduling policy type for placing vreplicas on pods (see type SchedulerPolicyType for enum list)
         - name: SCHEDULER_POLICY_TYPE
-          value: 'MAXFILLUP'
+          value: 'EVENSPREAD_BYNODE'
```
aavarghese, line 58: Will be reverted later...

pkg/common/scheduler/statefulset/scheduler.go
```diff
@@ -47,8 +47,10 @@ type SchedulerPolicyType string
 const (
 	// MAXFILLUP policy type adds replicas to existing pods to fill them up before adding to new pods
 	MAXFILLUP SchedulerPolicyType = "MAXFILLUP"
-	// EVENSPREAD policy type spreads replicas uniformly across failure-domains such as regions, zones, nodes, etc
+	// EVENSPREAD policy type spreads replicas uniformly across zones to reduce impact of failure
 	EVENSPREAD = "EVENSPREAD"
```
aavarghese, line 51: Should we differentiate this by adding ZONE to the policy name?

aavarghese: Irrelevant since recent convo with Lionel about one HA scheduler type with all of above.

PR #710, opened 2021-06-15
Add resource requests and limits to kafka-webhook

lberk:
The horizontalPodAutoscaler fails without set resources.

## Proposed Changes
- Add resources (values taken from serving, similar to eventing webhook defaults)
- :bug: Allows for hpa to work for kafka-webhook

**Release Note**
```release-note
The kafka-webhook deployment now sets resource requests and limits to enable hpa
```

### DIFF
--- a/config/channel/webhook/webhook-deployment.yaml
+++ b/config/channel/webhook/webhook-deployment.yaml
@@ -58,13 +58,11 @@ spec:
           containerPort: 8008
         resources:
           requests:
-            # taken from serving.
             cpu: 100m
             memory: 50Mi
           limits:
-            # taken from serving.
             cpu: 200m
-            memory: 200Mi
+            memory: 1024Mi
         readinessProbe: &probe
           periodSeconds: 1
           httpGet:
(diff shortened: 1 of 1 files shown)


lionelvillard:
@lberk FYI: https://github.com/knative-sandbox/eventing-kafka/pull/711

config/channel/webhook/webhook-deployment.yaml
```diff
@@ -56,8 +56,15 @@ spec:
...
-
+        resources:
+          requests:
+            # taken from serving.
+            cpu: 100m
+            memory: 50Mi
+          limits:
+            # taken from serving.
+            cpu: 200m
+            memory: 200Mi
```
matzew, line 67: I know this was taken from serving (and from there to eventing as well).
But something to have in mind that in "real world" we usually want larger value.

for Openshift Serverless, we do override that (in the extensions for serving and eventing) to 1G:
https://github.com/openshift-knative/serverless-operator/blob/main/openshift-knative-operator/pkg/eventing/extension.go#L55-L56

lberk: sure -- I can change the memory limit to 1G, anything else you'd prefer changed value wise?

matzew (approved):
/lgtm
/approve

PR #724, opened 2021-06-18
Create stop/start ConsumerGroup Manager with Control-Protocol

eric-sap:
Subtask of #588 to implement Stop/Start ConsumerGroups in response to control-protocol messaging.

## Proposed Changes

- Created controlprotocol/controlprotocol.go, a small wrapper around the knative.dev/control-protocol ControlServer
- Created consumer/consumer_manager.go, a container that allows temporary stop/start of managed Sarama ConsumerGroups.
- Split the KafkaConsumerGroupFactory's StartConsumerGroup function into createConsumerGroup and startExistingConsumerGroup (no change to the way external callers use StartConsumerGroup)
- Changed the distributed dispatcher to use the new KafkaConsumerGroupManager

Note that the "stop" and "start" of a managed ConsumerGroup is an abstraction of value to the user of the manager; internally this is accomplished by closing and recreating groups with the same Group ID.

If no control-protocol messages are sent to the dispatcher, visible behavior is unchanged .


pkg/common/kafka/testing/mocks.go
```diff
@@ -18,47 +18,57 @@ package testing
...
+	CloseErr    bool
 }
 
-// Mock ConsumerGroup Constructor
-func NewMockConsumerGroup(t *testing.T) *MockConsumerGroup {
-	return &MockConsumerGroup{
-		t:           t,
-		errorChan:   make(chan error),
+func NewStubbedMockConsumerGroup() *MockConsumerGroup {
+	mockGroup := &MockConsumerGroup{
```
travis-minke-sap, line 42: Why is this called `NewStubbedMockConsumerGroup()` instead of just `NewMockConsumerGroup()` ?

pkg/common/kafka/testing/mocks.go
```diff
@@ -18,47 +18,57 @@ package testing
...
+	mockGroup.On("Errors").Return(nil)
+	mockGroup.On("Close").Return(nil)
+	return mockGroup
 }
 
 func (m *MockConsumerGroup) Consume(ctx context.Context, topics []string, handler sarama.ConsumerGroupHandler) error {
+	_ = m.Called(ctx, topics, handler)
 	<-m.consumeChan                      // Block To Simulate Real Execution
 	return sarama.ErrClosedConsumerGroup // Return ConsumerGroup Closed "Error" For Clean Shutdown
 }
```
travis-minke-sap, line 59: seems odd to hard code a response when using the Testify mocks instead of... 
```
args := m.Called(ctx, topics, handler)
<-m.consumeChan
return args.Error(0)
```

pkg/common/kafka/testing/mocks.go
```diff
@@ -18,47 +18,57 @@ package testing
...
 func (m *MockConsumerGroup) Close() error {
-	close(m.errorChan)
+	_ = m.Called()
+	if m.CloseErr {
+		return fmt.Errorf("error closing consumer group")
+	}
+	close(m.ErrorChan)
 	close(m.consumeChan)
 	m.Closed = true
 	return nil
```
travis-minke-sap, line 74: same here if using testify mocks? (throughout)

pkg/common/consumer/consumer_manager.go
```diff
@@ -0,0 +1,391 @@
...
+			processAsyncGroupNotification(commandMessage, manager.startConsumerGroup)
+		})
+
+	return manager
+}
+
+// Reconfigure will incorporate a new set of brokers and Sarama config settings into the manager
+// without requiring a new control-protocol server or losing the current map of managed groups.
+// It will stop and start all of the managed groups in
+func (m *kafkaConsumerGroupManagerImpl) Reconfigure(brokers []string, config *sarama.Config) error {
```
travis-minke-sap, line 114: in what?  the suspense is killing me!  ; )

pkg/common/consumer/consumer_manager.go
```diff
@@ -0,0 +1,391 @@
...
+	// Make sure a managed group is "started" before closing the inner ConsumerGroup; otherwise anything
+	// waiting for the manager to restart the group will never return.
+	managedGrp.closeRestartChannel()
+	if managedGrp.cancelErrors != nil {
+		managedGrp.cancelErrors() // This will terminate any transferError call that is waiting for a restart
+	}
+	if managedGrp.cancelConsume != nil {
+		managedGrp.cancelConsume() // This will stop the factory's consume loop after the ConsumerGroup is closed
+	}
+	err := managedGrp.group.Close()
```
travis-minke-sap, line 189: should we log warnings if the `cancelErrors()` and `cancelConsume()` are nil (unexpected right?)

pkg/common/consumer/consumer_manager.go
```diff
@@ -0,0 +1,391 @@
...
+	if err != nil {
+		groupLogger.Error("Failed To Close Managed ConsumerGroup", zap.Error(err))
+		return err
+	}
+	// Remove this groupId from the map so that manager functions may not be called on it
+	delete(m.groups, groupId)
+	return nil
```
travis-minke-sap, line 196: does this delete need to be in the same synchronization as the "add" above?

pkg/common/consumer/consumer_manager.go
```diff
@@ -0,0 +1,391 @@
...
+// in the commandMessage, after verifying that the command version is correct.  It then calls the
+// appropriate Async response function on the commandMessage (NotifyFailed or NotifySuccess)
+func processAsyncGroupNotification(commandMessage ctrlservice.AsyncCommandMessage, groupFunction func(groupId string) error) {
+	if cmd, ok := commandMessage.ParsedCommand().(*commands.ConsumerGroupAsyncCommand); ok {
+		if cmd.Version != commands.ConsumerGroupAsyncCommandVersion {
+			commandMessage.NotifyFailed(fmt.Errorf("version mismatch; expected %d but got %d", commands.ConsumerGroupAsyncCommandVersion, cmd.Version))
+		} else {
+			// Calling NotifyFailed with a nil error is the same as calling NotifySuccess
+			commandMessage.NotifyFailed(groupFunction(cmd.GroupId))
+		}
```
travis-minke-sap, line 298: Well, it might be the same but... it's awkward to read and requires this comment and relies on the control-protocol library to continue behaving that way (not really documented).   I guess it's functionally ok but it gave me pause thinking... "what?  we ALWAYS fail?"

pkg/common/consumer/consumer_manager.go
```diff
@@ -0,0 +1,391 @@
...
+	}
+}
+
+// managedGroup contains information about a Sarama ConsumerGroup that is required to start (i.e. re-create) it
+type managedGroup struct {
+	group              sarama.ConsumerGroup
```
travis-minke-sap, line 304: should all this managedGroup be broken out into a separate file (same package) - seems like it's own thing - could have a test, etc?

pkg/common/controlprotocol/controlprotocol.go
```diff
@@ -0,0 +1,110 @@
...
+		router:       messageRouter,
+		cancelServer: serverCancelFn,
+	}, nil
+}
+
+// Shutdown cancels the control-protocol server and wait for it to stop
+func (s serverHandlerImpl) Shutdown() {
+	s.cancelServer()
+	waitChannelClosed(s.server.ClosedCh())
+}
```
travis-minke-sap, line 86: do we need a timeout of some sort here - will it block shutdown forever if channel doesn't return?

pkg/common/controlprotocol/controlprotocol_test.go
```diff
@@ -0,0 +1,112 @@
...
+		},
+		{
+			name:      "Server Error",
+			serverErr: fmt.Errorf("test error"),
+		},
+	} {
+		t.Run(testCase.name, func(t *testing.T) {
+			mockService := &ctrltesting.MockControlProtocolService{}
+			mockService.On("MessageHandler", mock.Anything).Return()
+			startServerWrapper = func(_ context.Context, _ ...network.ControlServerOption) (*network.ControlServer, error) {
```
travis-minke-sap, line 55: not asserting expectations on `mockService` ?

pkg/common/controlprotocol/controlprotocol_test.go
```diff
@@ -0,0 +1,112 @@
...
+	waitGroup.Wait()
+}
+
+func TestHandlers(t *testing.T) {
+	saveStartServer := startServerWrapper
+	defer func() { startServerWrapper = saveStartServer }()
+
+	mockService := &ctrltesting.MockControlProtocolService{}
+	mockService.On("MessageHandler", mock.Anything).Return()
+
```
travis-minke-sap, line 88: asserting expectations?

pkg/common/kafka/testing/mocks.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 The Knative Authors
+Copyright 2021 The Knative Authors
```
travis-minke-sap, line 3: Did moving the file cause this?  I don't think we are supposed to roll the original copyright dates forward as we edit files?

travis-minke-sap:
Looking good Eric - I added a few questions / nit-picky things ; )

pkg/common/kafka/testing/mocks.go
eric-sap: Because originally there were two; will rename

pkg/common/kafka/testing/mocks.go
eric-sap: These are all because I back-filled the testify stuff into hard-coded mocks; I'll clean them up to be testify-style-only

pkg/common/kafka/testing/mocks.go
eric-sap: cleaned up

pkg/common/consumer/consumer_manager.go
eric-sap: addressed prepositional failure

pkg/common/consumer/consumer_manager.go
eric-sap: Good idea; I don't think they can be non-nil unless someone creates a managedGroup without using the createManagedGroup() function.  (And guess what I did that prompted these nil-checks...)

pkg/common/consumer/consumer_manager.go
eric-sap: Oh yes, that's a good point.

pkg/common/consumer/consumer_manager.go
eric-sap: I had it that way originally and was thinking of putting it back anyway.  If it made you pause as well I will go ahead and change it.

pkg/common/consumer/consumer_manager.go
eric-sap: It's probably got just enough of its own functions to be worth an additional file.  It already had its own explicit tests so was fairly simple to split out.

pkg/common/controlprotocol/controlprotocol.go
eric-sap: That's a good idea; it also removes the need for the extra nil channel test

pkg/common/controlprotocol/controlprotocol_test.go
eric-sap: Ach, I completely forgot about that when I converted the mocks to testify-style.  I will go through the tests and make sure they're doing that for all the ones I added .On()

pkg/common/controlprotocol/controlprotocol_test.go
eric-sap: this and quite a few others

pkg/common/kafka/testing/mocks.go
eric-sap: I moved this file several times; it wouldn't surprise me, but I will replace it.

pkg/common/controlprotocol/controlprotocol.go
```diff
@@ -0,0 +1,106 @@
...
+
+// AddAsyncHandler will add an async handler for the given opcode to the message router used by the
+// control-protocol server, which will send a message using the resultOpcode when it finishes.
+func (s serverHandlerImpl) AddAsyncHandler(opcode ctrl.OpCode, resultOpcode ctrl.OpCode, payloadType message.AsyncCommand, handler AsyncHandlerFunc) {
+	s.router[opcode] = ctrlservice.NewAsyncCommandHandler(s.server, payloadType, resultOpcode, handler)
+}
+
+// AddSyncHandler will add a handler to the control-protocol server for the given opcode
+func (s serverHandlerImpl) AddSyncHandler(opcode ctrl.OpCode, handler ctrl.MessageHandlerFunc) {
```
slinkydeveloper, line 99: Note that this is going to create race conditions if it's invoked while the router is being used

eric-sap: That's a good point; I will clear the router first before calling the cancelServer function

eric-sap: Thanks - I've modified the logic here to include a setHandler() function that uses a copy of the router instead of allowing direct modifications to the map.

eric-sap:
/test pull-knative-sandbox-eventing-kafka-upgrade-tests
This test seems to have an error in VerifyContinualTests/SourceContinualTests but never finishes after that.  Not sure how that test works.. guess I'll dig into it if a retest doesn't work

PR #725, opened 2021-06-18
Resetoffset controller control protocol

travis-minke-sap:
Subtask of #588 to add control-protocol communication with dataplane to the ResetOffset Controller.

## Proposed Changes

- 🧽   Refactor "ref mapper" results into a single RefInfo
- 🎁   Provide distributed KafkaChannel mappers
- 🎁   Added new mappers to account for dataplane labelling differences
- 🎁   Implement ConnectionPool and AsyncNotificationStore communication with dataplane services.
- 🎁   Unit test everything delivered so far

**Notes**
- The ResetOffset Controller is still "inactive" and has not been added to any Channels/Sources/etc.
- The ResetOffset Controller is still not "complete" and this is a stepping stone in the path to implementing the feature.  There are still TODO's in the code.
- This PR and #724 are two halves of the control-protocol communication and contain some overlap that will require updating depending on which merges first.  This is to be expected and accounts for some of the TODOs.

**Release Note**

**Docs**


pkg/common/commands/resetoffset/controller/dataplane.go
```diff
@@ -0,0 +1,215 @@
...
+
+	// Return The Result
+	return err
+}
+
+// generateCommandId returns an int64 hash based on the specified ResetOffset.
+func generateCommandId(resetOffset *kafkav1alpha1.ResetOffset, podIP string, opCode ctrl.OpCode) (int64, error) {
+	hash := fnv.New32a()
+	_, err := hash.Write([]byte(fmt.Sprintf("%s-%d-%s-%d", string(resetOffset.UID), resetOffset.Generation, podIP, opCode)))
+	if err != nil {
```
eric-sap, line 211: The sum32 implementation of the Hash32 Write() call can't actually return an error (though it's probably not bad to check it anyway; just mentioning it - might simplify the generateCommandId call chain).

travis-minke-sap: Yeah, seems like bad form to code against the inner implementation of a third party library rather than against the interface.  Good future foot-gun if the implementation changes in the future and starts returning errors ; )

pkg/common/commands/resetoffset/controller/dataplane.go
```diff
@@ -0,0 +1,215 @@
...
+	// Wait For All The AsyncCommand Results
+	waitGroup.Wait()
+
+	// Close & Drain The Error Channel
+	close(errChan)
+	var multiErr error
+	for err := range errChan {
+		if err != nil {
+			multiErr = multierr.Append(multiErr, err)
+		}
```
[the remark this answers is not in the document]
travis-minke-sap: cool - thanks - will change

slinkydeveloper:
@travis-minke-sap Am I wrong or the control protocol commands in this PR it's the same code from #724?

pkg/common/commands/resetoffset/controller/controller.go
```diff
@@ -42,15 +47,25 @@ func NewControllerFactory(refMapperFactory refmappers.ResetOffsetRefMapperFactor
...
 		// Get The Needed Informers
+		podInformer := pod.Get(ctx)
 		resetoffsetInformer := resetoffset.Get(ctx)
 
 		// Create The RefMapper Via The Supplied Factory Using Initialized Context
 		refMapper := refMapperFactory.Create(ctx)
 
+		// Create A Control-Protocol AsyncCommandNotificationStore - No-Op Enqueue Function For ResetOffset Use Case
+		enqueueFunc := func(key types.NamespacedName) {
+			logger.Info("Control-Protocol Enqueue Function", zap.String("Key", key.String()))
```
slinkydeveloper, line 58: debug

pkg/common/commands/resetoffset/controller/dataplane.go
```diff
@@ -0,0 +1,213 @@
...
+		logger.Error("Failed to getting data-plane Pod IPs", zap.Error(err))
+		return nil, err
+	}
+
+	// Append The Control-Protocol Server Port Number To The PodIPs If Not Already Present
+	for index, podIP := range podIPs {
+		if !strings.Contains(podIP, ":") {
+			podIPs[index] = fmt.Sprintf("%s:%d", podIP, ControlProtocolServerPort)
+		}
+	}
```
slinkydeveloper, line 68: I don't think this is necessary if you use the default control protocol port, it should automatically append the port if no one is specified

pkg/common/commands/resetoffset/controller/dataplane.go
```diff
@@ -0,0 +1,213 @@
...
+	// Create And Send A The ConsumerGroupAsyncCommand to The Control-Protocol Service
+	consumerGroupAsyncCommand := commands.NewConsumerGroupAsyncCommand(commandId, refInfo.TopicName, refInfo.GroupId)
+	err = service.SendAndWaitForAck(opCode, consumerGroupAsyncCommand)
+	if err != nil {
+		logger.Error("Failed to send ConsumerGroup AsyncCommand", zap.Int64("CommandID", commandId), zap.Error(err))
+		return fmt.Errorf("failed to send ConsumerGroup AsyncCommand '%d' : %v", commandId, err)
+	}
+
+	// Wait For The AsyncCommand Result & Return Results
+	return r.waitForAsyncCommandResult(resetOffset, podIP, consumerGroupAsyncCommand)
```
slinkydeveloper, line 172: So you're making the command send asynchronous blocking the control loop right?

pkg/common/commands/resetoffset/controller/controller.go
travis-minke-sap: 👍  good call - will change

travis-minke-sap:
> @travis-minke-sap Am I wrong or the control protocol commands in this PR it's the same code from #724?

Yes, @eric-sap and I are working on the opposite sides of this (sending / receiving these commands) and have both included them in our PR.  Whichever merges first is fine and the other person will rebase, etc.  After we get #724 and #725 in we can do some formal testing and probably will have a small alignment / cleanup PR - there are a few things in this PR that will need to be adjusted/moved/etc...

pkg/common/commands/resetoffset/controller/dataplane.go
travis-minke-sap: In #724 @eric-sap has created a constant (temporarily mirrored in this PR) for the Port number instead of relying on the control-protocol default.  Therefore, I am coding against the "same" constant here for consistency.  I don't care too much whether we use the default, a constant, or expose it as a config though - whatever you think is best, We can change both the PRs either way.

pkg/common/commands/resetoffset/controller/dataplane.go
travis-minke-sap: Yes, we have made the Stop/Start ConsumerGroupAsyncCommands asynchronous by nature as they are also intended to be used for the Pause/Resume feature.  For the ResetOffset use case though I want to make all the async calls in parallel and then block on their completion before proceeding.

eric-sap (approved):
lgtm

PR #739, opened 2021-06-24
[Source] Optimize InitOffsets

lionelvillard:
InitOffsets can take a long time (several seconds, sometimes minutes), when the number of partitions is high (> 500), due to Sarama inability to batch OffsetFetch requests. 

## Proposed Changes

- batch OffsetFetch requests, one per leading broker. 
- 
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/common/kafka/sarama/batch_get_offsets.go
+++ b/pkg/common/kafka/sarama/batch_get_offsets.go
@@ -1,0 +1,83 @@
+/*
+Copyright 2021 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


steven0711dong:
* If my understanding is correct, instead of calling the GetOffset function partition number equivalent of times, we are only making broker number equivalent of times to get all the offsets? 
* When we construct the OffsetRequest, don't we need to fill the blocks as well? The way we do it is very simple by only filling out the version. 
* Lastly, have you tested this PR somewhere?

lionelvillard:
> * If my understanding is correct, instead of calling the GetOffset function partition number equivalent of times, we are only making broker number equivalent of times to get all the offsets?

Correct.

> * When we construct the OffsetRequest, don't we need to fill the blocks as well? The way we do it is very simple by only filling out the version.

See `request.AddBlock`. 

> * Lastly, have you tested this PR somewhere?

Besides unit testing and local testing? Existing e2e tests run this code and there is an open [PR](https://github.com/knative-sandbox/eventing-kafka/pull/636) to specifically test there is no missing events after the source is ready.

steven0711dong (approved):
LGTM

PR #742, opened 2021-06-28
Lock unlock protocol flags

eric-sap:
Subtask of #588 to add locks to ConsumerGroupAsyncCommands (dispatcher side)

Proposed Changes
🎁. Added functionality to the managedGroup (resetLockTimer/canUnlock/removeLock) to allow locking and unlocking of a managedGroup via an arbitrary token.
🎁  Changed the stopConsumerGroup and startConsumerGroup functions to call processLock, which understands the fields in the new CommandLock struct of the ConsumerGroupAsyncCommand and will lock/unlock the managedGroup
🧽  Change the Reconfigure function to also obey the managedGroup lock/unlock using an internal CommandLock token

The ResetOffset Controller is still "inactive" and has not been added to any Channels/Sources/etc.
The ResetOffset Controller is still not "complete" and this is a stepping stone in the path to implementing the feature.

Release Note

Docs


pkg/common/consumer/consumer_manager.go
```diff
@@ -246,64 +278,123 @@ func (m *kafkaConsumerGroupManagerImpl) consume(ctx context.Context, groupId str
 
 // stopConsumerGroups closes the managed ConsumerGroup identified by the provided groupId, and marks it
 // as "stopped" (that is, "able to be restarted" as opposed to being closed by something outside the manager)
-func (m *kafkaConsumerGroupManagerImpl) stopConsumerGroup(groupId string) error {
+func (m *kafkaConsumerGroupManagerImpl) stopConsumerGroup(lock *commands.CommandLock, groupId string) error {
+	if err := m.processLock(lock, groupId, true); err != nil {
+		return err
```
travis-minke-sap, line 283: log error?

eric-sap: Good idea; will do with startConsumerGroup as well

pkg/common/consumer/consumer_manager.go
```diff
@@ -246,64 +278,123 @@ func (m *kafkaConsumerGroupManagerImpl) consume(ctx context.Context, groupId str
...
+	if err != nil {
+		groupLogger.Error("Failed To Close Managed ConsumerGroup", zap.Error(err))
+		// Don't leave the start channel open if the group.Close() call failed; that would be misleading
+		managedGrp.closeRestartChannel()
+		return err
 	}
-	groupLogger.Info("ConsumerGroup Not Managed - Ignoring Stop Request")
-	return fmt.Errorf("stop requested for consumer group not in managed list: %s", groupId)
+
+	return m.processLock(lock, groupId, false)
```
travis-minke-sap, line 309: Comment here and above maybe?  not clear (yet - still reading) what processLock() does and how the initial boolean flag affects it?

pkg/common/consumer/consumer_manager.go
```diff
@@ -246,64 +278,123 @@ func (m *kafkaConsumerGroupManagerImpl) consume(ctx context.Context, groupId str
...
+	m.groups[groupId].group = group
+	m.groups[groupId].closeRestartChannel() // Closing this allows the waitForStart function to finish
+	m.groupLock.Unlock()
+
+	return m.processLock(lock, groupId, false)
+}
+
+// getGroup returns a group from the groups map using the groupLock mutex
+func (m *kafkaConsumerGroupManagerImpl) getGroup(groupId string) *managedGroup {
+	m.groupLock.RLock()
```
travis-minke-sap, line 341: Seems like there are lots of direct access (with their own locking) to m.groups elsewhere.  Seems "risky" to have two different approaches - can/should the others all use this (pass in a function to to in a locked state) or maybe just remove this function and lock/unlock inline as is done everywhere else?

pkg/common/consumer/consumer_manager.go
```diff
@@ -246,64 +278,123 @@ func (m *kafkaConsumerGroupManagerImpl) consume(ctx context.Context, groupId str
...
+	defer m.groupLock.RUnlock()
+	return m.groups[groupId]
+}
+
+// processLock handles setting and removing the managedGroup's lock status.
+// For the managedGroup with the given groupId, if that group exists and the provided CommandLock fields warrant it,
+// this function will:
+// - if "before" is true, set the lockedBy field of the group
+// - if "before" is false, remove the lockedBy field of the group
```
travis-minke-sap, line 350: not entirely true - if "before" is false AND lock indicates `unlockafter` right ?

eric-sap: That's part of "and the CommandLock field warrants it" -- but I will clean up this comment; it's confusing

pkg/common/consumer/managed_group.go
```diff
@@ -19,34 +19,107 @@ package consumer
...
+	managedGrp.lockedBy.Store("")   // Empty token string indicates "unlocked"
+	managedGrp.stopped.Store(false) // A managed group defaults to "started" when created
+
 	return &managedGrp
 }
 
+// resetLockTimer will set the "lockedBy" field to the given lockToken (need not be the same as the
+// existing one) and reset the timer to the provided timeout.  After the timer expires, the lockToken
+// will be set to an empty string (representing "unlocked")
+func (m *managedGroup) resetLockTimer(lockToken string, timeout time.Duration) {
```
travis-minke-sap, line 64: I would rename to `resetLock` since it doesn't only change the timer but replaces the token ?

pkg/common/consumer/managed_group.go
```diff
@@ -19,34 +19,107 @@ package consumer
...
+
+	// Stop any existing timer (without releasing the lock) so that it won't inadvertently do an unlock later
+	if m.cancelLockTimeout != nil {
+		m.cancelLockTimeout()
+	}
+
+	if lockToken != "" && timeout != 0 {
```
travis-minke-sap, line 71: Are we really NEVER replacing a token/timeout with just a token ?

pkg/common/consumer/consumer_manager.go
eric-sap: I'll add some hopefully demystifying comments.  The idea was that "processLock" would either lock or unlock the group based on the third argument ("before").  I originally had a "lockBefore" and "unlockAfter" function but they were 95% the same.  I am not opposed to recreating those and having them just call a common function if "processLock" is confusing

pkg/common/consumer/managed_group.go
eric-sap: The timeout is required at the low-level of the managedGroup itself.  There is a default timeout applied at a higher level (in the processLock function) between the managedGroup and the async command.

pkg/common/consumer/managed_group.go
eric-sap: That's probably more succinct.  Originally I was thinking "reset lock and timer"

pkg/common/consumer/managed_group.go
travis-minke-sap: ok, thanks!

pkg/common/consumer/managed_group.go
travis-minke-sap: that's fine too - whatever you prefer to make it clear.

pkg/common/consumer/consumer_manager.go
eric-sap: You're right; that was a bit half-done.  I've added proper getGroup/setGroup/removeGroup functions and removed the mutex/map access from elsewhere

pkg/common/consumer/consumer_manager.go
eric-sap: I added back the lockBefore/unlockAfter wrapper functions; they're trivial but I think it makes reading the stopConsumerGroup and startConsumerGroup functions less confusing.

matzew (approved):
/lgtm
/approve

PR #746, opened 2021-06-30
Fix stop start without locking

eric-sap:
This fixes a few bugs that cropped up due to #742 

## Proposed Changes

- 🐛 The managedGroup lock/unlock was being allowed if no CommandLock struct was passed to processLock, even if it was already locked with a token.
- 🐛 CloseConsumerGroup was logging warnings indicating that cancelErrors and cancelConsume were nil when they, in fact, were not
- 🧽 Several places were still directly accessing the groups map instead of using the getGroup/setGroup/deleteGroup functions
- 🧽 Added an interface for the managedGroup (now has managedGroupImpl) to clear up some confusion

The lock/unlock bug was due in no small part to functionality not being very well isolated to the appropriate structs (manager vs. managedGroup), so I have changed "managedGroup" to an interface that the manager now uses instead of arbitrarily changing fields.  To that end, this interface now has "consume/start/stop/close/errors/processLock" as the lifecycle functions that the manager calls, and a bit of functionality has moved from the manager into the managedGroupImpl (particularly consume and processLock).

I realize this is a bit of a nuisance refactor right after merging the original lock/unlock code, but none of that is live yet (the resetoffset controller is not enabled which is what would send the control-protocol messages) and I believe this is a much better way to handle the managedGroup.

### DIFF
--- a/pkg/common/consumer/consumer_manager.go
+++ b/pkg/common/consumer/consumer_manager.go
@@ -49,7 +49,7 @@ import (
 
 const (
 	defaultLockTimeout = 5 * time.Minute
-	internalToken = "internal-token"
+	internalToken      = "internal-token"
 )
 
 // KafkaConsumerGroupManager keeps track of Sarama consumer groups and handles messages from control-protocol clients
--- a/pkg/common/consumer/consumer_manager.go
+++ b/pkg/common/consumer/consumer_manager.go
@@ -49,6 +49,7 @@ import (
 
 const (
 	defaultLockTimeout = 5 * time.Minute
+	internalToken = "internal-token"
 )
 
 // KafkaConsumerGroupManager keeps track of Sarama consumer groups and handles messages from control-protocol clients
@@ -118,7 +119,7 @@ func (m *kafkaConsumerGroupManagerImpl) Reconfigure(brokers []string, config *sa
 	var multiErr error
 	groupsToRestart := make([]string, 0, len(m.groups))
 	for groupId := range m.groups {
-		err := m.stopConsumerGroup(&commands.CommandLock{Token: "internal-token", LockBefore: true}, groupId)
+		err := m.stopConsumerGroup(&commands.CommandLock{Token: internalToken, LockBefore: true}, groupId)
 		if err != nil {
 			// If we couldn't stop a group, or failed to obtain a lock, note it as an error.  However,
 			// in a practical sense, the new brokers/config will be used when whatever locked the group
@@ -135,7 +136,7 @@ func (m *kafkaConsumerGroupManagerImpl) Reconfigure(brokers []string, config *sa
 	// Restart any groups this function stopped
 	m.logger.Info("Reconfigure Consumer Group Manager - Starting All Managed Consumer Groups")
 	for _, groupId := range groupsToRestart {
-		err := m.startConsumerGroup(&commands.CommandLock{Token: "internal-token", UnlockAfter: true}, groupId)
+		err := m.startConsumerGroup(&commands.CommandLock{Token: internalToken, UnlockAfter: true}, groupId)
 		if err != nil {
 			multierr.AppendInto(&multiErr, err)
 		}
--- a/pkg/common/consumer/consumer_manager.go
+++ b/pkg/common/consumer/consumer_manager.go
@@ -273,14 +273,16 @@ func (m *kafkaConsumerGroupManagerImpl) startConsumerGroup(lock *commands.Comman
 		return fmt.Errorf("start requested for consumer group not in managed list: %s", groupId)
 	}
 
-	group, err := m.factory.createConsumerGroup(groupId)
+	createGroup := func() (sarama.ConsumerGroup, error) {
+		return m.factory.createConsumerGroup(groupId)
+	}
+
+	// Instruct the managed group to use this new ConsumerGroup
+	err := managedGrp.start(createGroup)
 	if err != nil {
 		groupLogger.Error("Failed To Restart Managed ConsumerGroup", zap.Error(err))
 		return err
 	}
-
-	// Instruct the managed group to use this new ConsumerGroup
-	managedGrp.start(group)
 
 	// Unlock the managedGroup after starting it, if lock.UnlockAfter is true
 	if err = m.unlockAfter(lock, groupId); err != nil {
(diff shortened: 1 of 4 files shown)


eric-sap:
/hold
Would like to see what @travis-minke-sap thinks of this before merging

pkg/common/consumer/consumer_manager.go
```diff
@@ -123,7 +118,7 @@ func (m *kafkaConsumerGroupManagerImpl) Reconfigure(brokers []string, config *sa
 	var multiErr error
 	groupsToRestart := make([]string, 0, len(m.groups))
 	for groupId := range m.groups {
-		err := m.stopConsumerGroup(getInternalLockCommand(true), groupId)
+		err := m.stopConsumerGroup(&commands.CommandLock{Token: "internal-token", LockBefore: true}, groupId)
```
travis-minke-sap, line 121: should "internal-token" be a constant ?

pkg/common/consumer/managed_group.go
```diff
@@ -18,50 +18,151 @@ package consumer
...
 // empty string (i.e. "unlocked") after that time has passed.
-func createManagedGroup(logger *zap.Logger, group sarama.ConsumerGroup, cancel func()) *managedGroup {
+func createManagedGroup(ctx context.Context, logger *zap.Logger, group sarama.ConsumerGroup, cancelErrors func(), cancelConsume func()) managedGroup {
 
-	managedGrp := managedGroup{
-		logger:       logger,
-		group:        group,
-		errors:       make(chan error),
-		cancelErrors: cancel,
+	managedGrp := managedGroupImpl{
```
travis-minke-sap, line 64: not to nit pick but... it seems a bit more "standard / idiomatic" to do `managedGrp := &managedGroupImpl{}` here instead of returning the address below?

pkg/common/consumer/consumer_manager.go
eric-sap: It should; that was laziness on my part

pkg/common/consumer/managed_group.go
eric-sap: I'll buy that

pkg/common/consumer/managed_group.go
```diff
@@ -18,50 +18,151 @@ package consumer
...
+func (m *managedGroupImpl) stop() error {
+	// The managedGroup's start channel must be created (that is, the managedGroup must be marked
+	// as "stopped") before closing the internal ConsumerGroup, otherwise the consume function would
+	// return control to the factory.
+	m.createRestartChannel()
+	// Close the inner sarama ConsumerGroup, which will cause our consume() function to stop
+	// and wait for the managedGroup to start again.
+	m.groupMutex.Lock()
+	group := m.group
+	m.groupMutex.Unlock()
```
travis-minke-sap, line 94: should the close below be in the same mutex block?  if not the comment above should be moved down to that block since this block just acquires the group?

eric-sap: I don't think I want to expand the scope of the mutex here; I'm just gating read-access to the m.group field (vs. writing it in the start() function), not the Close() function of the interface.  However, you are definitely correct about the comment position being unclear; I will move that closer to the Close() call.

pkg/common/consumer/consumer_manager.go
```diff
@@ -345,14 +290,14 @@ func (m *kafkaConsumerGroupManagerImpl) startConsumerGroup(lock *commands.Comman
...
-func (m *kafkaConsumerGroupManagerImpl) getGroup(groupId string) *managedGroup {
+func (m *kafkaConsumerGroupManagerImpl) getGroup(groupId string) managedGroup {
 	m.groupLock.RLock()
 	defer m.groupLock.RUnlock()
 	return m.groups[groupId]
 }
 
 // getGroup associates a group with a groupId in the groups map using the groupLock mutex
-func (m *kafkaConsumerGroupManagerImpl) setGroup(groupId string, group *managedGroup) {
+func (m *kafkaConsumerGroupManagerImpl) setGroup(groupId string, group managedGroup) {
```
travis-minke-sap, line 301: comment is "getGroup" for function "setGroup"

pkg/common/consumer/consumer_manager.go
```diff
@@ -345,14 +290,14 @@ func (m *kafkaConsumerGroupManagerImpl) startConsumerGroup(lock *commands.Comman
...
 	m.groupLock.Lock()
 	m.groups[groupId] = group
 	m.groupLock.Unlock()
```
travis-minke-sap, line 304: why is the unlock not a `defer` like elsewhere ?

pkg/common/consumer/consumer_manager.go
```diff
@@ -367,51 +312,20 @@ func (m *kafkaConsumerGroupManagerImpl) removeGroup(groupId string) {
 
 // lockBefore will lock the managedGroup corresponding to the groupId, if lock.LockBefore is true
 func (m *kafkaConsumerGroupManagerImpl) lockBefore(lock *commands.CommandLock, groupId string) error {
-	return m.processLock(lock, groupId, true)
+	group := m.getGroup(groupId)
+	if group == nil {
+		return nil // Can't lock a nonexistent group
```
travis-minke-sap, line 318: should we log a warning here - shouldn't be happening right?

pkg/common/consumer/consumer_manager.go
eric-sap: I can never win, heh.

pkg/common/consumer/consumer_manager.go
eric-sap: It doesn't need it here; I only use the defer for mutex.Unlock when I'm trying to "return" the same struct that's being protected.

pkg/common/consumer/managed_group.go
```diff
@@ -18,50 +18,151 @@ package consumer
...
+	if err := group.Close(); err != nil {
+		// Don't leave the start channel open if the group.Close() call failed; that would be misleading
+		m.closeRestartChannel()
+		return err
+	}
+	return nil
+}
+
+// start starts the managed group, using the provided group as the "restarted" internal group
+func (m *managedGroupImpl) start(group sarama.ConsumerGroup) {
```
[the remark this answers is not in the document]
eric-sap: I think the confusion is that I use the term "start" and "stop" in reference to the "managedGroup" and not the internal Sarama ConsumerGroup.  The managedGroup is "started" if it has a valid ConsumerGroup.

I will rethink the comments and process here though; you're right in that the nomenclature isn't super-clear.

pkg/common/consumer/managed_group.go
```diff
@@ -18,50 +18,151 @@ package consumer
...
+	return m.group.Close()
+}
+
+// consume calls the Consume function on the managed ConsumerGroup, supporting the stop/start functionality
+func (m *managedGroupImpl) consume(ctx context.Context, topics []string, handler sarama.ConsumerGroupHandler) error {
+	for {
+		// Call the internal sarama ConsumerGroup's Consume function directly
+		m.groupMutex.Lock()
+		group := m.group
+		m.groupMutex.Unlock()
```
travis-minke-sap, line 143: same here on deferring unlock?

pkg/common/consumer/managed_group.go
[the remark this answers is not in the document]
travis-minke-sap: sounds good - up to you - just style/naming mostly ; )

travis-minke-sap:
Thanks Eric!
/lgtm

PR #755, opened 2021-07-07
Change kafkachannel status when paused

eric-sap:
Subtask of #588 to update the "subscribers" section of the status block for KafkaChannel resources when a ConsumerGroup is started or stopped.

## Proposed Changes
🎁  Added Go channel notification option to the KafkaConsumerGroupManager (GetNotificationChannel) so that external entities may be informed of changes to a managed ConsumerGroup's status (create/start/stop/close)
🎁. Modified the distributed Dispatcher's KafkaChannel controller to use this new channel from the Manager and call EnqueueKey in order to trigger a reconciliation when Consumer Groups are started or stopped.
🎁  Modified the distributed Dispatcher's UpdateSubscriptions() call to return a map of status infomation that includes "stopped," so that the reconciler can use that information to update the subscribers status block properly.

**Notes**

- When the KafkaChannel's status is updated, the related Knative Eventing subscription resource is automatically (no change by this PR) changed to a status of "False" with a reason of SubscriptionNotMarkedReadyByChannel ("Subscription marked by Channel as False").  I believe this is sufficient.
- The ResetOffset Controller is still "inactive" and has not been added to any Channels/Sources/etc.

**Release Note**

**Docs**

### DIFF
--- a/pkg/channel/distributed/dispatcher/controller/kafkachannel.go
+++ b/pkg/channel/distributed/dispatcher/controller/kafkachannel.go
@@ -111,27 +111,42 @@ func NewController(
 			w.Stop()
 		}
 	}()
+	if err := reconciler.processManagerEvents(managerEvents); err != nil {
+		logger.Warn("Could not begin processing events from the Consumer Group Manager", zap.Error(err))
+	}
+	return reconciler.impl
+}
 
-	// Process events from the Consumer Group Manager
+// processManagerEvents will listen on the channel provided by the KafkaConsumerGroupManager for events
+// related to changes in the status of a managed ConsumerGroup.  This function is non-blocking, and the
+// internal goroutine will exit when the channel is closed.
+func (r Reconciler) processManagerEvents(events <-chan commonconsumer.ManagerEvent) error {
+	// If the events channel is nil, there's no point listening to it; it will just block forever
+	if events == nil {
+		return fmt.Errorf("no event channel provided")
+	}
+	// Find the namespace and name of the KafkaChannel this dispatcher is monitoring
+	namespace, name, err := cache.SplitMetaNamespaceKey(r.channelKey)
+	if err != nil {
+		return fmt.Errorf("invalid resource key")
+	}
+	key := types.NamespacedName{Namespace: namespace, Name: name}
 	go func() {
-		for event := range managerEvents {
+		for event := range events {
 			// Stopping or Starting a consumer group requires a reconciliation in order to adjust the status block
-			// of the KafkaChannel - calling GlobalResync will force that re-reconciliation
-			// TODO: Is it okay to just resync everything?  We could possibly update a date-style annotation
-			// on the kafkachannel instead to cause a reconciliation, or something similar
+			// of the KafkaChannel - EnqueueKey will do that
 			switch event.Event {
 			case commonconsumer.GroupStopped:
-				logger.Debug("Processing GroupStopped Event From Consumer Group Manager")
-				reconciler.impl.GlobalResync(reconciler.kafkachannelInformer)
+				r.logger.Debug("Processing GroupStopped Event From Consumer Group Manager")
+				r.impl.EnqueueKey(key)
 			case commonconsumer.GroupStarted:
-				logger.Debug("Processing GroupStarted Event From Consumer Group Manager")
-				reconciler.impl.GlobalResync(reconciler.kafkachannelInformer)
+				r.logger.Debug("Processing GroupStarted Event From Consumer Group Manager")
+				r.impl.EnqueueKey(key)
 			}
 		}
-		logger.Debug("Manager event channel closed")
+		r.logger.Debug("Manager event channel closed")
 	}()
-
-	return reconciler.impl
+	return nil
 }
 
 func (r Reconciler) Reconcile(ctx context.Context, key string) error {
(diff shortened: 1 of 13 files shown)


eric-sap:
/hold
holding for unit test race condition experiment

pkg/channel/distributed/dispatcher/controller/kafkachannel.go
```diff
@@ -106,10 +111,44 @@ func NewController(
...
+			// Stopping or Starting a consumer group requires a reconciliation in order to adjust the status block
+			// of the KafkaChannel - EnqueueKey will do that
+			switch event.Event {
+			case commonconsumer.GroupStopped:
+				r.logger.Debug("Processing GroupStopped Event From Consumer Group Manager")
+				r.impl.EnqueueKey(key)
+			case commonconsumer.GroupStarted:
+				r.logger.Debug("Processing GroupStarted Event From Consumer Group Manager")
+				r.impl.EnqueueKey(key)
+			}
```
travis-minke-sap, line 145: Can the log messages include the ConsumerGroup ID ?

Can you add a "default" case that logs unexpected/unsupported event type (just as defensive coding and so that handling of this is clear?)

pkg/common/consumer/consumer_manager.go
```diff
@@ -111,6 +134,38 @@ func NewConsumerGroupManager(logger *zap.Logger, serverHandler controlprotocol.S
 	return manager
 }
 
+// AddNotification creates a new ManagerEvent channel and returns it.  Manager events will be
+// broadcast to all channels created this way.
+func (m *kafkaConsumerGroupManagerImpl) AddNotification() <-chan ManagerEvent {
+	eventChan := make(chan ManagerEvent)
```
travis-minke-sap, line 140: `AddNotification()` seems like a weird name for this?  Maybe `RegisterForNotifications()` or `ReceiveNotifications()` or `GetNotificationChannel()` or something?  Not a big deal - up to you - just a suggestion.

pkg/channel/distributed/dispatcher/controller/kafkachannel.go
eric-sap: Sure; I added the groupId to a new logger and filled out the rest of the cases

pkg/common/consumer/consumer_manager.go
eric-sap: I like GetNotificationChannel; I'll use that.  I think you're right that the consumer of the interface doesn't care that it's adding a channel to a list internally.

pkg/channel/distributed/dispatcher/dispatcher/dispatcher.go
```diff
@@ -177,9 +178,15 @@ func (d *DispatcherImpl) UpdateSubscriptions(subscriberSpecs []eventingduck.Subs
...
-
 			// Otherwise Just Add To List Of Active Subscribers
 			activeSubscriptions[subscriberSpec.UID] = true
+
+			// If the group is stopped, it's still active but the reconciler needs to know about it in order
+			// to not treat it as a failure (which would re-create the group, effectively un-stopping it)
+			if d.consumerMgr.IsStopped(groupId) {
+				d.Logger.Debug("Adding Stopped ConsumerGroup To Stopped Map", zap.String("GroupId", groupId))
+				stoppedSubscriptions[subscriberSpec.UID] = struct{}{}
+			}
```
travis-minke-sap, line 189: Still reviewing but... I'm wondering why the `stoppedSubscriptions` has a type of struct and really whether the separate map is needed at all?  The only different treatment of the two maps (failed/stopped) is that we hard-code an error for the stopped ones.  Couldn't we put then all in the same map and just hardcode the stopped message here instead?

eric-sap: I looked into combining the maps before, but not in that way (I was trying to combine all three maps).  I'll give it another try with only those two maps and see what happens (I can't "just have the error" unless I compare the error against the hard-coded value to see if it's stopped vs. failed after reconcile() calls UpdateSubscriptions -- right now if "length of failedSubscriptions > 0" then reconcile() returns an error, which we don't want if all of our "failures" are actually just stopped)

pkg/common/consumer/consumer_manager.go
```diff
@@ -111,6 +134,38 @@ func NewConsumerGroupManager(logger *zap.Logger, serverHandler controlprotocol.S
...
+	m.eventLock.Lock()
+	m.notifyChannels = append(m.notifyChannels, eventChan)
+	m.eventLock.Unlock()
+	return eventChan
+}
+
+// ClearNotifications closes and removes all channels from the notification list
+func (m *kafkaConsumerGroupManagerImpl) ClearNotifications() {
+	for _, eventChan := range m.notifyChannels {
```
travis-minke-sap, line 149: This doesn't appear to be used in the production code - is it just here for API "completeness" ?  OK if so, wasn't sure.

eric-sap: Hm, you're right.  I thought I called that in the shutdown code but I seem to have forgotten it.  I'll put it where it belongs.

pkg/channel/distributed/dispatcher/dispatcher/dispatcher.go
travis-minke-sap: ok, i missed that nuance - don't have to force it into one array - i thought the only use was to know to hardcode the error message.

eric-sap: I refactored this using a new type of "SubscriberStatusMap" and I think I've gotten it a fair bit cleaner now, to the point at least where this confusion wouldn't have happened :)

travis-minke-sap:
thanks for the changes Eric
/lgtm

PR #768, opened 2021-07-15
Next generation MT Scheduler and Descheduler with pluggable policies

### LINKED ISSUE #593: [scheduler,mtsource] Implementing EvenSpread HA support across nodes within a zone

aavarghese:
Fixes #593 
🎁  Implementing next generation MT scheduler and descheduler for Knative Eventing sources with plugins for filtering pods
 and scoring them to pick the best placement pod. 

## Proposed Changes

- Controller reads and validates the policy profile (containing predicates and priorities) for both scheduling and descheduling from ConfigMaps.
- Scheduler imports all the available plugins which invokes their init() and registers all the core and source specific (kafka) plugins.
- Scheduler runs all the predicates and priorities specified in the policy (configmap) file to eliminate all the pods that do not meet the predicate, and then prioritizes the feasible pods to find the best placement given the current state of the cluster.
- Same for descheduler (has separate predicates and priorities).
- Controls are also shared with autoscaler, and state accessor.
- Autoscaler runs autoscaling synchronously using a lock so that autoscaling does not run together with the scheduler, and also scales up number of pods using a scale factor based on number of resources in the HA domain 
- State accessor computes the spread of existing and reserved vreplicas across pods, nodes and zones and stores a map of vpod key to a map of podname/nodename/zonename to vreplica count. It also stores the number of pods, nodes, zones and a list of "schedulable" pods (unschedulable pods are pods that are marked for eviction for compacting, and pods that are on unschedulable nodes (cordoned or unreachable nodes))
- Pending vreplicas that cannot be scheduled triggers the autoscaler to add new pods. This leads to rebalancing of all the already placed vreplicas together with the pending vreplicas.
- When its time to autoscale down (after refresh period), compacting the pods is attempted if space is available on lower pods. Higher pods are evicted and removed from the "schedulable" pods list. This leads to another rebalancing of all the vreplicas.
- Recovery situations are addressed by making sure pods on failed nodes/zones are restarted on new nodes (statefulset adapter settings) and state collector is aware of nodes/zones being down.

**Release Note**

```
Next generation Multi-Tenant Scheduler and Descheduler: uses a plugin interface to specify a Scheduler profile with predicates and priorities that run filtering and scoring of pods, respectively to compute the best vreplica placements. When the autoscaler adds new pods, scheduler performs a rebalancing of the already placed vreplicas along with the new vreplicas. A Descheduler profile must be installed when vreplicas need to be scaled down and placements removed. 
```


pkg/common/scheduler/statefulset/scheduler.go
```diff
@@ -181,32 +222,57 @@ func (s *StatefulSetScheduler) scheduleVPod(vpod scheduler.VPod) ([]duckv1alpha1
...
+			logger.Infow("number of replicas per domain", zap.Int32("spreadVal", spreadVal))
+			placements, left = s.addReplicasEvenSpread(state, vpod.GetVReplicas()-tr, placements, spreadVal)
 		} else {
-			placements = s.removeReplicas(tr-vpod.GetVReplicas(), placements)
+			placements, left = s.addReplicas(state, vpod.GetVReplicas()-tr, placements)
 		}
 
-		// Do not trigger the autoscaler to avoid unnecessary churn
+	} else if s.policy != nil { //Predicates and priorities must be used for scheduling
+		if errs := ValidatePolicy(s.policy); errs != nil {
```
lionelvillard, line 263: Can we find another place to validate policies? Seems wasteful to do this here.

aavarghese: Ok - yeah, could move it to be done right after init within [here](https://github.com/knative-sandbox/eventing-kafka/pull/768/files#diff-58d2cad92561a8118d077178816fb6d2689b58a9ea043ba65969c94077913ebfR216)

pkg/source/reconciler/mtsource/kafkasource.go
```diff
@@ -183,3 +188,46 @@ func (r *Reconciler) createCloudEventAttributes(src *v1beta1.KafkaSource) []duck
 	}
 	return ceAttributes
 }
+
+// initPolicyFromConfigMap reads predicates and priorities data from configMap
+func initPolicyFromConfigMap(ctx context.Context, configMapName string, policy *sts.SchedulerPolicy) error {
+	policyConfigMap, err := kubeclient.Get(ctx).CoreV1().ConfigMaps(system.Namespace()).Get(context.TODO(), configMapName, metav1.GetOptions{})
```
lionelvillard, line 194: `context.TODO()` -> `ctx`

pkg/source/reconciler/mtsource/kafkasource.go
```diff
@@ -183,3 +188,46 @@ func (r *Reconciler) createCloudEventAttributes(src *v1beta1.KafkaSource) []duck
...
+	if !found {
+		return fmt.Errorf("missing policy config map value at key priorities")
+	}
+	if err := json.NewDecoder(strings.NewReader(priors)).Decode(&policy.Priorities); err != nil {
+		return fmt.Errorf("invalid policy: %v", err)
+	}
+	logging.FromContext(ctx).Infof("Priorities to be registered: %v", policy.Priorities)
+
+	if errs := validatePolicy(policy); errs != nil {
+		return errors.New("validating policy failed")
```
lionelvillard, line 218: you may want to return `errs` or at least log them

pkg/source/reconciler/mtsource/controller.go
```diff
@@ -119,15 +120,19 @@ func NewController(
...
 		logging.FromContext(ctx).Debugw("Patched resource", zap.Any("patch", patch), zap.Any("patched", patched))
 		return nil
 	}
 
-	c.scheduler = stsscheduler.NewScheduler(ctx,
-		system.Namespace(), mtadapterName, c.vpodLister, rp, env.PodCapacity, env.SchedulerPolicy,
-		nodeInformer.Lister(), evictor)
+	policy := &stsscheduler.SchedulerPolicy{}
+	if err := initPolicyFromConfigMap(ctx, env.SchedulerPolicyConfigMap, policy); err != nil {
+		return nil
```
lionelvillard, line 131: policies are immutable, admin-controlled parameters so it's fine to `panic` here.

config/source/multi/600-config-scheduler.yaml
```diff
@@ -0,0 +1,36 @@
+# Copyright 2021 The Knative Authors
```
lionelvillard, line 1: seems like it should be 400-config-scheduler.yaml to be consistent

aavarghese: Done

pkg/source/reconciler/mtsource/kafkasource.go
aavarghese: Done.

pkg/source/reconciler/mtsource/kafkasource.go
aavarghese: Sorry, didn't follow @lionelvillard. You are recommending setting the ctx via context.TODO() instead of being passed in as input?

lionelvillard: instead of `.Get(context.TODO(), ...`, what about just `.Get(ctx, ....`?

aavarghese: Ohh yes, totally missed that. Good catch - thank you!

aavarghese: Done

pkg/common/scheduler/state/state.go
```diff
@@ -0,0 +1,299 @@
+/*
```
aavarghese, line 1: @lionelvillard State file somehow got screwed up when moving to a new location and now shows up as a new file. So lost git history. Tried a few things but I think the ultimate fix might be a new branch and start from scratch to retrieve history....

lionelvillard: no biggie.

config/source/multi/400-config-descheduler.yaml
```diff
@@ -0,0 +1,35 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-descheduler
```
lionelvillard, line 18: I would name it `config-kafka-source-descheduler` to make it clear this configmap is only relevant for KafkaSource

config/source/multi/400-config-scheduler.yaml
```diff
@@ -0,0 +1,38 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config-scheduler
```
lionelvillard, line 18: same here

config/source/multi/400-config-descheduler.yaml
```diff
@@ -0,0 +1,35 @@
...
+  namespace: knative-eventing
+  labels:
+    kafka.eventing.knative.dev/release: devel
+data:
+  predicates: |+
```
lionelvillard, line 23: can you add an `_example` section? Here an example: https://github.com/knative/eventing/blob/main/config/core/configmaps/default-pingsource.yaml

lionelvillard: The `_example` checksum should also be generated. It's done in https://github.com/knative-sandbox/eventing-kafka/pull/855 which adds the necessary tooling to do so.

config/source/multi/deployments/adapter.yaml
```diff
@@ -78,7 +78,7 @@ spec:
         - name: profiling
           containerPort: 8008
 
-      terminationGracePeriodSeconds: 10
+      terminationGracePeriodSeconds: 0 # for forceful termination and restart for quick recovery
```
lionelvillard, line 81: It should really not be `0`. What about keeping `10` and address this quick recovery in a followup PR?

pkg/common/scheduler/plugins/core/availabilitynodepriority/availability_node_priority.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	skewVal := state.AvailabilityNodePriorityArgs{}
+	decoder := json.NewDecoder(strings.NewReader(spreadArgs))
+	decoder.DisallowUnknownFields()
+	if err := decoder.Decode(&skewVal); err != nil {
+		return 0, state.NewStatus(state.Unschedulable, ErrReasonInvalidArg)
+	}
+
+	if states.Replicas > 0 { //need atleast a pod to compute spread
```
lionelvillard, line 74: nit: `atleast` -> `at least`

config/source/multi/400-config-descheduler.yaml
```diff
@@ -0,0 +1,35 @@
...
+              []
+  priorities: |+
```
lionelvillard, line 25: in order to be consistent with the rest of Knative I would use YAML instead of JSON. Examples:
- https://github.com/knative/eventing/blob/main/config/core/configmaps/default-broker.yaml
- https://github.com/knative-sandbox/eventing-kafka/blob/main/config/channel/distributed/300-eventing-kafka-configmap.yaml

This can be done in a followup PR

pkg/common/scheduler/state/helpers.go
```diff
@@ -0,0 +1,98 @@
...
+
+func OrdinalFromPodName(podName string) int32 {
+	ordinal, err := strconv.ParseInt(podName[strings.LastIndex(podName, "-")+1:], 10, 32)
+	if err != nil {
+		return math.MaxInt32
+	}
+	return int32(ordinal)
+}
+
+func Retry(f func() error) error {
```
lionelvillard, line 40: what about using this function instead: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#Poll ?

pkg/common/scheduler/state/interface.go
```diff
@@ -0,0 +1,209 @@
...
+	NoMaxResourceCount                 = "NoMaxResourceCount"
+	EvenPodSpread                      = "EvenPodSpread"
+	AvailabilityNodePriority           = "AvailabilityNodePriority"
+	AvailabilityZonePriority           = "AvailabilityZonePriority"
+	LowestOrdinalPriority              = "LowestOrdinalPriority"
+	RemoveWithEvenPodSpreadPriority    = "RemoveWithEvenPodSpreadPriority"
+	RemoveWithAvailabilityNodePriority = "RemoveWithAvailabilityNodePriority"
+	RemoveWithAvailabilityZonePriority = "RemoveWithAvailabilityZonePriority"
+	RemoveWithHighestOrdinalPriority   = "RemoveWithHighestOrdinalPriority"
+)
```
lionelvillard, line 38: this feels strange as these are plugins and this file should know nothing about them, right?

pkg/common/scheduler/state/interface.go
```diff
@@ -0,0 +1,209 @@
...
+	// Score is called by the scheduler.
+	// All ScorePlugins should return "Success" unless the args are invalid.
+	Score(ctx context.Context, args interface{}, state *State, feasiblePods []int32, key types.NamespacedName, podID int32) (uint64, *Status)
+
+	// ScoreExtensions returns a ScoreExtensions interface if it implements one, or nil if does not
+	ScoreExtensions() ScoreExtensions
+}
+
+// NoMaxResourceCountArgs holds arguments used to configure the NoMaxResourceCount plugin.
+type NoMaxResourceCountArgs struct {
```
lionelvillard, line 72: same comment as above, seems like it should belongs in the NoMaxResourceCount plugin

pkg/common/scheduler/state/interface.go
```diff
@@ -0,0 +1,209 @@
...
+	NumPartitions int
+}
+
+// EvenPodSpreadArgs holds arguments used to configure the EvenPodSpread plugin.
+type EvenPodSpreadArgs struct {
+	MaxSkew int32
+}
+
+// AvailabilityZonePriorityArgs holds arguments used to configure the AvailabilityZonePriority plugin.
+type AvailabilityZonePriorityArgs struct {
```
lionelvillard, line 82: same

pkg/common/scheduler/state/interface.go
```diff
@@ -0,0 +1,209 @@
...
+	MaxSkew int32
+}
+
+// AvailabilityNodePriorityArgs holds arguments used to configure the AvailabilityNodePriority plugin.
+type AvailabilityNodePriorityArgs struct {
```
lionelvillard, line 87: same

pkg/common/scheduler/state/state.go
```diff
@@ -0,0 +1,399 @@
...
+	nodes, err := s.nodeLister.List(labels.Everything())
+	if err != nil {
+		return nil, err
+	}
+
+	nodeToZoneMap := make(map[string]string)
+	zoneMap := make(map[string]struct{})
+	for i := 0; i < len(nodes); i++ {
+		node := nodes[i]
+		if node.Spec.Unschedulable {
```
lionelvillard, line 211: I think this line requires some explanation. In theory the vpod scheduler shouldn't have to deal with nodes directly. A pod on an unschedulable node might still be viable, right?

pkg/common/scheduler/state/state.go
```diff
@@ -0,0 +1,399 @@
...
+		})
+
+		var unschedulable bool
+		annotVal, ok := pod.ObjectMeta.Annotations[constants.PodAnnotationKey] //Pod is marked for eviction - CANNOT SCHEDULE VREPS on this pod
+		if ok {
+			unschedulable, _ = strconv.ParseBool(annotVal)
+		}
+
+		nodeName := pod.Spec.NodeName
+		node, err := s.nodeLister.Get(nodeName) //Node could be marked as Unschedulable - CANNOT SCHEDULE VREPS on a pod running on this node
```
lionelvillard, line 247: same here. Maybe we just don't care about pod eviction. This has to be dealt somewhere else anyway

config/source/multi/400-config-descheduler.yaml
aavarghese: Done

pkg/common/scheduler/state/interface.go
aavarghese: @lionelvillard Interface.go turned out to be like the common place to store all the plugin name strings and arg details. Mainly for reuse (between similar scheduler and descheduler plugins) and also to expose info to the `Autoscaler`. Autoscaler HA mode is turned on (Scale up/down by number of domains) by checking the presence of an HA plugin like `AvailabilityZonePriority`

pkg/common/scheduler/state/interface.go
aavarghese: This is shared between `AvailabilityZonePriority` plugin and `RemoveWithAvailabilityZonePriority` plugin

pkg/common/scheduler/state/interface.go
```diff
@@ -0,0 +1,209 @@
...
+	ScoreExtensions() ScoreExtensions
+}
+
+// NoMaxResourceCountArgs holds arguments used to configure the NoMaxResourceCount plugin.
+type NoMaxResourceCountArgs struct {
+	NumPartitions int
+}
+
+// EvenPodSpreadArgs holds arguments used to configure the EvenPodSpread plugin.
+type EvenPodSpreadArgs struct {
```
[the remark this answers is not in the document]
aavarghese: same as above.

pkg/common/scheduler/state/interface.go
aavarghese: same as above.

pkg/common/scheduler/state/state.go
aavarghese: Thin line I feel. A node is marked as Unschedulable in its Spec after it has been optionally drained and cordoned. K8 scheduler won't place new pods onto that Node, so idea was that we _should not_ place "new vreplicas" onto this pod either. Because the next step for this Node might be rebooting/shutting down/upgrading etc. So this is a preventative step to add more rebalancing work. That was my thinking..

pkg/common/scheduler/state/state.go
aavarghese: Each plugin must be given some info to prevent it from selecting a pod that is in the process of eviction. Not sure yet of how to do that without the state collector telling it so...will think about this a bit more.

PR #779, opened 2021-07-26
add initial offset to kafka source spec

### LINKED ISSUE #289: Provide option for configuring initial offset in kafka source
**Problem**
Continuing the discussion on this [issue](https://github.com/knative/eventing-contrib/issues/1492) in the new repo.

Currently the kafka source starts from the newest offset. With this option the kafka source needs to be created before the producer sends message. This is not the case in most of the kafka use cases as producer is independent of the consumer. In these scenarios initial offset needs to be the earliest. 

If offset was already saved for a partition, sarama-cluster will resume consumption from that offset. The Config.Consumer.Offsets.Initial option is used only if no saved offset is present (first run for a consumer group)

https://github.com/Shopify/sarama/blob/0cdf4a6c31734ea9473b03a36fff21947a8e9d5c/config.go#L381

cc @slinkydeveloper @aliok

itsmurugappan:
Fixes #289 

## Proposed Changes

- Adds initial offset to the kafkasource spec

**Release Note**

```release-note
🎁  - Provide an option for the user to specify the initial offset for a consumer group in Kafka source, this field is honored only if there are no prior offsets committed for the consumer group.
```

/cc @lionelvillard

### DIFF
--- a/pkg/apis/sources/v1beta1/kafka_types.go
+++ b/pkg/apis/sources/v1beta1/kafka_types.go
@@ -75,7 +75,7 @@ type KafkaSourceSpec struct {
 	// InitialOffset is the Initial Offset for the consumer group.
 	// should be earliest or latest
 	// +optional
-	InitialOffset string `json:"initialOffset,omitempty"`
+	InitialOffset Offset `json:"initialOffset,omitempty"`
 
 	// inherits duck/v1 SourceSpec, which currently provides:
 	// * Sink - a reference to an object that will resolve to a domain name or
@@ -85,15 +85,19 @@ type KafkaSourceSpec struct {
 	duckv1.SourceSpec `json:",inline"`
 }
 
+type Offset string
+
 const (
 	// KafkaEventType is the Kafka CloudEvent type.
 	KafkaEventType = "dev.knative.kafka.event"
 
 	KafkaKeyTypeLabel = "kafkasources.sources.knative.dev/key-type"
 
-	OffsetEarliest = "earliest"
+	// OffsetEarliest denotes the earliest offset in the kafka partition
+	OffsetEarliest Offset = "earliest"
 
-	OffsetLatest = "latest"
+	// OffsetLatest denotes the latest offset in the kafka partition
+	OffsetLatest Offset = "latest"
 )
 
 var KafkaKeyTypeAllowed = []string{"string", "int", "float", "byte-array"}
--- a/pkg/apis/sources/v1beta1/kafka_types.go
+++ b/pkg/apis/sources/v1beta1/kafka_types.go
@@ -73,8 +73,9 @@ type KafkaSourceSpec struct {
 	ConsumerGroup string `json:"consumerGroup,omitempty"`
 
 	// InitialOffset is the Initial Offset for the consumer group.
+	// should be earliest or latest
 	// +optional
-	InitialOffset *int64 `json:"initialOffset,omitempty"`
+	InitialOffset string `json:"initialOffset,omitempty"`
 
 	// inherits duck/v1 SourceSpec, which currently provides:
 	// * Sink - a reference to an object that will resolve to a domain name or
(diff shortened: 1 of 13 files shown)


pkg/source/reconciler/source/resources/receive_adapter_test.go
```diff
@@ -38,6 +39,7 @@ func TestMakeReceiveAdapter(t *testing.T) {
 		Spec: v1beta1.KafkaSourceSpec{
 			Topics:        []string{"topic1,topic2"},
 			ConsumerGroup: "group",
+			InitialOffset: ptr.Int64(-2),
```
matzew, line 42: what's `-2` ?

matzew:
Isn't the official kafka (jvm) client just doing `earliest` and `latest` ?

pkg/source/reconciler/source/resources/receive_adapter_test.go
itsmurugappan: earliest, i will change to use the constants from sarama library

pkg/apis/sources/v1beta1/kafka_types.go
```diff
@@ -72,6 +72,10 @@ type KafkaSourceSpec struct {
 	// +optional
 	ConsumerGroup string `json:"consumerGroup,omitempty"`
 
+	// InitialOffset is the Initial Offset for the consumer group.
+	// +optional
+	InitialOffset *int64 `json:"initialOffset,omitempty"`
```
pierDipi, line 77: We should not use sarama specific things in the API.

can we use `earliest` or `latest` as string?

/hold

pierDipi (changes requested):

pkg/apis/sources/v1beta1/kafka_types.go
itsmurugappan: what if we want to give a specific offset ?

pierDipi: Do you plan to set a single offset for all partitions?

pierDipi: Or, let me ask another question, as of now, if I set 50 here, what should happen?

itsmurugappan: sorry, I had a wrong assumption that we could also give a specific offset. 
But looks like it needs to be earliest or latest. I will change the API  to make the initial offset a string that will accept "earliest" or "latest"

itsmurugappan: i have made the changes

pierDipi: Thanks!

/unhold

pkg/source/reconciler/source/resources/receive_adapter.go
```diff
@@ -70,6 +70,13 @@ func MakeReceiveAdapter(args *ReceiveAdapterArgs) *v1.Deployment {
 		})
 	}
 
+	if args.Source.Spec.InitialOffset != "" {
+		env = append(env, corev1.EnvVar{
+			Name:  "INITIAL_OFFSET",
```
lionelvillard, line 75: seems like it should be `KAFKA_INITIAL_OFFSET` to be consistent with the rest.

lionelvillard:
thanks @itsmurugappan! 

Can you also update the documentation (if not done already), or a least open an issue in `knative/docs`? 

/approve

@pierDipi @matzew for LGTM...

itsmurugappan:
docs pr -> https://github.com/knative/docs/pull/4080

lionelvillard:
firs time I see this flake in the scheduler. I'll take a look when I get a chance

/test pull-knative-sandbox-eventing-kafka-unit-tests

lionelvillard:
@matzew are you ok with this?

pierDipi:
There is only one comment left https://github.com/knative-sandbox/eventing-kafka/pull/779#discussion_r677736553 that still applies but at a different line https://github.com/knative-sandbox/eventing-kafka/pull/779/files#diff-aa0fe2380b2926d6809e7204b18c7c78f0d0158748106f0dca91a667ab700d21R88

PR #781, opened 2021-07-28
Resetoffset E2E rekt Tests

travis-minke-sap:
Subtask of #588 to add e2e "rekt" style tests for KafkaChannel ResetOffset.

The ResetOffset feature is now active on the distributed KafkaChannel implementation and the last bit of work is to provide e2e tests to ensure its operational status.  The new "rekt" style testing has been used for these tests.  This is the first time using this framework for the KafkaChannel implementations so some base level tests (readiness, event sending, etc) have also been provided.

## Proposed Changes

- 🎁  Added KafkaChannel and ResetOffset "resources" to the existing "rekt" tests.
- 🎁  Added KafkaChannel "features" to the existing "rekt" tests.
- 🎁  Added new KafkaChannel Tests to the existing "rekt" tests.
- 🧽  Updated the e2e_common.sh script to isolate existing KafkaSource tests for appropriate execution.

**Docs**
Documentation has already been provided by subtasks of #588


travis-minke-sap:
Holding for review by @lionelvillard who has both eventing-kafka and rekt experience  (thanks in advance ; )

/hold

test/rekt/resources/resetoffset/resetoffset.go
```diff
@@ -0,0 +1,92 @@
...
+		interval, timeout := k8s.PollTimings(ctx, timings)
+		env := environment.FromContext(ctx)
+		if err := k8s.WaitForResourceReady(ctx, env.Namespace(), name, GVR(), interval, timeout); err != nil {
+			t.Error(GVR(), "did not become succeeded", err)
+		}
+	}
+}
+
+// WithVersion overrides the default API version
+func WithVersion(version string) CfgFn {
```
eric-sap, line 65: Is anything using this function?  If it's only for future use that's no problem, just checking.

eric-sap (commented):
This looks pretty good; I'd be happier if someone who understands the original source rekt tests also looked at it though.

test/rekt/resources/resetoffset/resetoffset.go
travis-minke-sap: No just there for future use and consistency with the KafkaSource "resource" implementations.

test/rekt/features/kafkachannel/util.go
```diff
@@ -0,0 +1,192 @@
...
+// CloudEvent Utilities
+//
+
+// newEvent returns a new CloudEvent suitable for testing.
+func newEvent(name string, senderName string) cloudevents.Event {
+	event := cloudevents.NewEvent()
+	event.SetSource(senderName)
+	event.SetType(name + "-type")
+	event.SetSubject(name + "-subject")
+	_ = event.SetData(cloudevents.ApplicationJSON, map[string]string{"name": name})
```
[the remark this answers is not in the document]
travis-minke-sap: I'd rather keep "t" out of newEvent() and instead just return the error.  Originally I had ignored the error because this is just test code that should never fail given it has "mostly" fixed content and should never fail, but... you are right there is still a possibility of it happening.  I'll make changes to return error and check in caller.

lionelvillard:
/lgtm

sorry for the delay.

travis-minke-sap:
Thanks Lionel - no worries on the delay - I know you we're out ; )
/unhold

PR #783, opened 2021-07-29
[distributed KafkaChannel] Add receiver/dispatcher labels to Pods as well as Deployments

travis-minke-sap:
It is useful in production environments to be able to identify KafkaChannel Dispatcher Pods via a label.  Currently we have labels on the Deployments but not the Pods.  This PR fixes that by applying the labels in both places.

## Proposed Changes
- 🧽  Enhance the Controller logic for creating Receiver/Dispatcher Deployments so that the labels are applied to resulting Pods as well.


matzew (approved):
/lgtm
/approve

PR #795, opened 2021-08-05
fix-reconfigure-and-reconciliation-issues

### LINKED ISSUE #794: [Distributed KafkaChannel] Dispatcher Reconciliation & Reconfigure Lifecycle Problems
**Describe the bug**
There are a couple problems with the Dispatcher's "controller" which is watching KafkaChannels in order to start/stop Kafka ConsumerGroups...

1. The reconciler does not requeue upon reconciliation errors so subscribers can become "stuck" in a bad state until the KafkaChannel is changed.
2. An existing kafkachannel can be put into a not ready state by a bad Reconfigure (Kafka Secret change) and is not returned when the Secret is fixed.
3. A "stopped" subscription would unintentionally be restarted by a Reconfigure (Kafka Secret Change)

 **Expected behavior**
1. The reconciler should return errors to force a requeue of the KafkaChannel in an attempt to resolve the problem.
2. A Subscription is returned to READY state after a bad Kafka Secret change is corrected.
3. A "stoped" subscription should not be affected by a Reconfigure (Kafka Secret Change)

 **To Reproduce**
1. Create a KafkaChannel with a Subscription.  Change the Kafka Secret to invalid values which will trigger a reconfigure which will result in the Subscription being "MarkedNotReadyByChannel".  Verify the reconciler requeues the KafkaChannel and keeps trying (backoff)
2. Create a KafkaChannel with a Subscription.  Change the Kafka Secret to invalid values which will trigger a reconfigure which will result in the Subscription being "MarkedNotReadyByChannel".  Fix the Kafka Secret back to valid values and notice that the Subscription is not returned to a READY state.
3. Really hard to reproduce now since the "Pause" capability is not in place - would have to catch a ResetOffset at just the right moment.  

 **Knative release version**
/main

 **Additional context**
This code needs to respect the control-protocol messaging used to Stop/Start Consumer Groups for the ResetOffset (and eventually the Pause) features.  Subscriptions which are "stopped" should not be restarted as part of a Reconfigure (Kafka Secret Change)

travis-minke-sap:
Fixes #794 

## Proposed Changes

- 🧽  Restructure Dispatcher main() to create Controller prior to SecretWatcher.
- 🧽  Change Dispatcher's SecretObserver to be a factory capable of requeueing KafkaChannels.
- 🧽  Add util function to convert from GroupId back to Subscription.UID.
- 🧽  Return errors from Dispatcher Controller Reconcile() to requeue on failure.
- 🧽  Create new ReconfigureError type to contain the failed ConsumerGroup IDs
- 🧽  Dispatcher.SecretChanged() should handle ReconfigureErrors by removing failed ConsumerGroups from Dispatcher.subscribers so that the next reconcile will not ignore them (as being previously closed)
- 🐛 Dispatcher.Reconfigure() bug fixed so that "paused" (stopped) aren't restarted.
- 🧽  Dispatcher.Reconfigure() tracks failed ConsumerGroup IDs and returns new ReconfigureError type.
- 🧽  Removed unneeded DispatcherConfig.SubscriberSpecs

**Release Note**

```release-note
Distributed KafkaChannel Dispatcher will now re-queue KafkaChannels after failed reconciliation which will improve various failure recovery scenarios, but could consume CPU resources if reconciliation is blocked by underlying system issues (bad Kafka Secret config, etc.)
```

Holding for further testing...
/hold

### DIFF
--- a/cmd/channel/distributed/dispatcher/main.go
+++ b/cmd/channel/distributed/dispatcher/main.go
@@ -232,7 +232,7 @@ func NewSecretObserver(kcController *kncontroller.Impl, channelKey string, dispa
 		// Signal The Dispatcher To Recreate Kafka Config And Reconnect All Current ConsumerGroups
 		dispatcher.SecretChanged(ctx, secret)
 
-		// Requeue The KafkaChannel To Fix Any "Stuck" Subscriptions From A Prior Bad Kafka Secret
+		// Requeue The KafkaChannel To Fix Any Not-Ready Subscriptions
 		channelNamespace, channelName, err := cache.SplitMetaNamespaceKey(channelKey)
 		if err != nil {
 			logger.Error("Invalid KafkaChannel Key", zap.String("ChannelKey", channelKey), zap.Error(err))
--- a/cmd/channel/distributed/dispatcher/main.go
+++ b/cmd/channel/distributed/dispatcher/main.go
@@ -24,8 +24,10 @@ import (
 
 	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
 	"knative.dev/eventing/pkg/kncloudevents"
 	injectionclient "knative.dev/pkg/client/injection/kube/client"
 	"knative.dev/pkg/configmap"
@@ -54,7 +56,6 @@ import (
 
 // Variables
 var (
-	logger        *zap.Logger
 	dispatcher    dispatch.Dispatcher
 	managerEvents <-chan commonconsumer.ManagerEvent
 )
@@ -85,7 +86,7 @@ func main() {
 	ctx = commonk8s.LoggingContext(ctx, constants.Component, k8sClient)
 
 	// Get The Logger From The Context & Defer Flushing Any Buffered Log Entries On Exit
-	logger = logging.FromContext(ctx).Desugar()
+	logger := logging.FromContext(ctx).Desugar()
 	defer flush(logger)
 
 	// Load Environment Variables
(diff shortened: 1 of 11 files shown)


cmd/channel/distributed/dispatcher/main.go
```diff
@@ -219,21 +210,35 @@ func flush(logger *zap.Logger) {
...
+		// Ignore Startup Scenario Where Dispatcher Reference Might Be Nil
+		if dispatcher == nil {
+			logger.Debug("Dispatcher is nil during call to secretObserver; ignoring changes")
+			return
+		}
+
+		// Signal The Dispatcher To Recreate Kafka Config And Reconnect All Current ConsumerGroups
+		dispatcher.SecretChanged(ctx, secret)
+
+		// Requeue The KafkaChannel To Fix Any "Stuck" Subscriptions From A Prior Bad Kafka Secret
```
eric-sap, line 235: This is really "from any error trying to start the ConsumerGroups," right?  Seems disingenuinely precise to call this out explicitly (perhaps "e.g." or just a more general reason?)

cmd/channel/distributed/dispatcher/main.go
```diff
@@ -219,21 +210,35 @@ func flush(logger *zap.Logger) {
...
+		channelNamespace, channelName, err := cache.SplitMetaNamespaceKey(channelKey)
+		if err != nil {
+			logger.Error("Invalid KafkaChannel Key", zap.String("ChannelKey", channelKey), zap.Error(err))
+		} else {
+			logger.Debug("Requeue-ing KafkaChannel due to Kafka Secret change", zap.String("ChannelKey", channelKey))
+			kcController.EnqueueKey(types.NamespacedName{Namespace: channelNamespace, Name: channelName})
+		}
```
eric-sap, line 242: Instead of passing in the channelKey and doing this SplitMetaNamespaceKey every time (with the error handling), would it be better to just pass in the NamespacedName to NewSecretObserver?  If I'm not mistaken it will never change during the life of the pod and would make this function less complicated.

pkg/channel/distributed/common/kafka/util/util.go
```diff
@@ -20,17 +20,26 @@ import (
 	"fmt"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/types"
+
 	"knative.dev/eventing-kafka/pkg/channel/distributed/common/kafka/constants"
 )
 
+const GroupIdPrefix = "kafka"
```
eric-sap, line 28: Seems like the prefix is "kafka." - why not include the dot here instead of having to add it both in GroupId() and Uid() and the tests ?

pkg/common/consumer/consumer_manager.go
```diff
@@ -120,6 +120,20 @@ type kafkaConsumerGroupManagerImpl struct {
...
 
+// ReconfigureError is a custom error type returned by the Reconfigure() function.
+type ReconfigureError struct {
+	MultiError error    // A MultiError with all ConsumerGroup failures.
+	GroupIds   []string // The ConsumerGroup IDs which failed to stop/start.
+}
+
+// Error implements the golang error interface.
+func (r *ReconfigureError) Error() string {
+	return r.MultiError.Error()
```
eric-sap, line 131: Is this okay if r.MultiError is nil?  It won't be in the code that uses it but since this is an exported struct it might be good to check.

eric-sap (commented):
Some minor questions

pkg/common/consumer/consumer_manager.go
travis-minke-sap: No, good call - I was thinking of returning the GroupIDs in the error string anyway - should be made more robust - I'll fix it!

pkg/channel/distributed/common/kafka/util/util.go
travis-minke-sap: I originally had it that way but felt the "." was a "separator" instead of the prefix - I think it's ok as is.

cmd/channel/distributed/dispatcher/main.go
travis-minke-sap: You're right that the `channelKey` value is "fixed" for the Dispatcher instance (since it's only ever dealing with one KafkaChannel) but I'd rather not pass in the NamespacedName since it's only used here, etc...

cmd/channel/distributed/dispatcher/main.go
travis-minke-sap: Yeah, that's true - will adjust the comment language - thanks

PR #798, opened 2021-08-09
[Consolidated KafkaChannel] Set KafkaChannel consolidated dispatcher OwenrRef

devguyio:
## Proposed Changes

- Set the consolidated KafkaChannel controller deployment as the owner of the dispatcher deployment. This means that the way to uninstall the KafkaChannes is by deleting the controller deployment only. 

**Release Note**

```release-note
- 🧽 The consolidated KafkaChannel dispatcher is now owned by the controller. 
```


pkg/channel/consolidated/reconciler/controller/controller.go
```diff
@@ -46,15 +41,26 @@ import (
...
-	roleLabelKey      = "messaging.knative.dev/role"
-	roleLabelValue    = "dispatcher"
+	channelLabelKey          = "messaging.knative.dev/channel"
+	channelLabelValue        = "kafka-channel"
+	roleLabelKey             = "messaging.knative.dev/role"
+	dispatcherRoleLabelValue = "dispatcher"
+	controllerRoleLabelValue = "controller"
 )
 
+var trueConst = true
```
matzew, line 62: do we need this?

devguyio: yes, wherever we need a ref to a true, which in this PR in line 121

aliok: how about https://pkg.go.dev/k8s.io/utils/pointer#Bool ?

devguyio: good one! didn't know about it.

matzew:
@devguyio
You need to delete this sym-link https://github.com/devguyio/eventing-kafka/blob/srvke-910/config/channel/consolidated/500-dispatcher.yaml

matzew:
Works... see `ownerReferences` 

```
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    deployment.kubernetes.io/revision: "1"
  creationTimestamp: "2021-08-10T11:18:02Z"
  generation: 1
  name: kafka-ch-dispatcher
  namespace: knative-eventing
  ownerReferences:
  - apiVersion: apps/v1
    controller: true
    kind: Deployment
    name: kafka-ch-controller
    uid: 738f6cb0-dd46-4fd9-856f-341bb35c4737
```

and therefore the clean up works...

pkg/channel/consolidated/reconciler/controller/resources/dispatcher.go
```diff
@@ -56,8 +58,9 @@ func MakeDispatcher(args DispatcherArgs) *v1.Deployment {
 			Kind:       "Deployments",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      dispatcherName,
-			Namespace: args.DispatcherNamespace,
+			Name:            dispatcherName,
+			Namespace:       args.DispatcherNamespace,
+			OwnerReferences: []metav1.OwnerReference{args.OwnerRef},
```
matzew, line 63: Should we add the `KafkaChannel` itself to the `Args`, and do smth like:
```go
OwnerReferences: []metav1.OwnerReference{*kmeta.NewControllerRef(args.kc)},
```
we do that w/ the source, like here: https://github.com/knative-sandbox/eventing-kafka/blob/83e1b7a7c4353f129e83e39f137f4f662765d242/pkg/source/reconciler/source/resources/receive_adapter.go#L86

matzew:
```
Aug 11 00:00:28.591 install_head_consolidated_channel [OUT] The Deployment "kafka-ch-controller" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/component":"controller", "app.kubernetes.io/name":"eventing-kafka-channel", "messaging.knative.dev/channel":"kafka-channel", "messaging.knative.dev/role":"controller"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```

pkg/channel/consolidated/reconciler/controller/resources/dispatcher.go
devguyio: that would mean you're setting the ownerRef of the dispatcher deployment to be owned by that KafkaChannel, which is not what we want to do here. This true in the source case because the receive adapter is per KafkaSource, but here the dispatcher is shared across all KafkaChannel object.

matzew:
@devguyio would it make sense to move the refactoring to a separate pull request?

matzew:
That would make the fix a bit more portable to older versions. So the refactoring goes in after?

devguyio:
@matzew I needed the refactoring to reason about that block which I need to touch and change with or without the refactoring. So I'm not very sure if without that refactoring it's gonna be more portable to older versions because that block needs to change anyway, and if it exists differently in older versions then even without the refactoring I'll have to manually fix the conflict in those versions.

matzew (approved):
/lgtm
/approve

PR #799, opened 2021-08-09
Separate CRDs, like we have on eventing core

matzew:

as per title 

## Proposed Changes

- separation of CRDs, like on eventing
-
-

**Release Note**

```release-note
Separated CRDs to extra, additional yaml manifest (similar to eventing)
```

**Docs**

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -26,8 +26,10 @@ declare -A COMPONENTS
 COMPONENTS=(
   ["channel-consolidated.yaml"]="config/channel/consolidated"
   ["channel-distributed.yaml"]="config/channel/distributed"
+  ["channel-crds.yaml"]="config/channel/resources"
   ["source.yaml"]="config/source/single"
   ["mt-source.yaml"]="config/source/multi"
+  ["source-crds.yaml"]="config/source/common/resources"
 )
 readonly COMPONENTS
 


hack/release.sh
```diff
@@ -26,8 +26,10 @@ declare -A COMPONENTS
 COMPONENTS=(
   ["channel-consolidated.yaml"]="config/channel/consolidated"
   ["channel-distributed.yaml"]="config/channel/distributed"
+  ["channel-crds.yaml"]="config/channel/resources"
```
aliok, line 29: In which file the CRDs are currently at?

matzew: it is defined in here: https://github.com/knative-sandbox/eventing-kafka/blob/main/config/channel/resources/kafkachannel-crd.yaml


And like on "eventing", we will have an extra file for _just_ the CRDs. See here as reference:
https://github.com/knative/eventing/blob/main/hack/generate-yamls.sh#L91-L94

This PR allows better reuse of just the CRDs

devguyio:
I don't know how the current release script works exactly but I'm fine with this change. Maybe it's a good idea to add a release note saying that the CRDs are now part of the release artifacts? 

/lgtm
/hold

matzew:
release notes added!

PR #809, opened 2021-08-18
removing confusing error message

### LINKED ISSUE #810: Remove confusing error message
**Describe the bug**
`Kafka source is not ready` is always produced as an event and this message is confusing 

 **Expected behavior**
Simply log the error and do not output this error as an event 

 **To Reproduce**
Create a Kafka source and describe the Kafka source and you will see the events 

 **Knative release version**

 **Additional context**

steven0711dong:
Fixes #810 

## Proposed Changes

-
-
-

**Release Note**

```release-note
🐛 Remove confusing error message `Kafka source is not ready`. This error is returned as an event regardless the Kafka source has any real issue or not and users should not see this error message unless it is a legitimate error or it is an error they can fix. 
```

**Docs**

### DIFF
--- a/pkg/source/mtadapter/kafkasource.go
+++ b/pkg/source/mtadapter/kafkasource.go
@@ -18,8 +18,8 @@ package mtadapter
 
 import (
 	"context"
-	"fmt"
 
+	"go.uber.org/zap"
 	"knative.dev/pkg/kmeta"
 	"knative.dev/pkg/reconciler"
 
@@ -30,6 +30,7 @@ import (
 // Reconciler updates the internal Adapter cache kafkaSources
 type Reconciler struct {
 	mtadapter MTAdapter
+	logger    *zap.SugaredLogger
 }
 
 // Check that our Reconciler implements ReconcileKind.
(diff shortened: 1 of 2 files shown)


pkg/source/mtadapter/kafkasource.go
```diff
@@ -30,14 +30,15 @@ import (
...
+	logger    *zap.SugaredLogger
 }
 
 // Check that our Reconciler implements ReconcileKind.
 var _ kafkasourcereconciler.Interface = (*Reconciler)(nil)
 
 func (r *Reconciler) ReconcileKind(ctx context.Context, source *v1beta1.KafkaSource) reconciler.Event {
 	if !source.Status.IsReady() {
-		return fmt.Errorf("warning: KafkaSource is not ready")
+		r.logger.Error("warning: KafkaSource is not ready")
```
lionelvillard, line 41: What about outputting a warning instead?

pkg/source/mtadapter/kafkasource.go
```diff
@@ -30,14 +30,15 @@ import (
...
 	}
```
lionelvillard, line 42: the return statement is missing. just `return nil` here.

PR #815, opened 2021-08-20
Add delete deployment permission to kafka channel controller

devguyio:
## Proposed Changes

- Add delete deployment permission to the consolidated KafkaChannel controller.

**Release Note**

```release-note
The consolidated KafkaChannel controller now has the permission to delete deployment which it needs to update the owner ref of its dispatcher deployment.
```

### DIFF
--- a/config/channel/consolidated/roles/controller-clusterrole.yaml
+++ b/config/channel/consolidated/roles/controller-clusterrole.yaml
@@ -88,7 +88,14 @@ rules:
     resources:
       - deployments
       - deployments/status
-    verbs: *everything
+    verbs:
+      - get
+      - list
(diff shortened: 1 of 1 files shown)


config/channel/consolidated/roles/controller-clusterrole.yaml
```diff
@@ -88,7 +88,14 @@ rules:
     resources:
       - deployments
       - deployments/status
-    verbs: *everything
```
matzew, line 91: why the hack is `everything` not .... _everything_ ? :face_with_head_bandage:

matzew (approved):
/lgtm
/approve

Issue #813, opened 2021-08-20
defaultNumPartitions from Config-Kafka ConfigMap is not working


matzew:
**Describe the bug**

Apply the standard yaml for `consolidated-channel`, with updated `broker`, like:

```yaml
apiVersion: v1
data:
  eventing-kafka: |
    kafka:
      brokers: my-cluster-kafka-bootstrap.kafka:9092
      topic:
        defaultNumPartitions: 4
        defaultReplicationFactor: 1 # Cannot exceed the number of Kafka Brokers!
        defaultRetentionMillis: 604800000  # 1 week
  version: 1.0.0

```

seem to not recognize the `defaultNumPartitions`.  Creating a channel like:

```yaml
apiVersion: messaging.knative.dev/v1beta1
kind: KafkaChannel
metadata:
  name: my-testchannel
```

Does still give you a topic w/ just one partition, see:

```shell
NAME                                                                                               CLUSTER      PARTITIONS   REPLICATION FACTOR   READY
knative-messaging-kafka.default.my-testchannel                                                     my-cluster   1            1                    True

```

 **Expected behavior**
yaml works :smile_cat: 

 **To Reproduce**
Steps to reproduce the behavior.

 **Knative release version**

 **Additional context**
Add any other context about the problem here such as proposed priority

travis-minke-sap:
Hey @matzew - I think this is a sub-set of the larger issue (#817) I've just created a PR (#818) for... maybe once I get this merged we can retest/close this?

matzew:
sounds good!

PR #818, opened 2021-08-20
[KafkaChannels] Add RetentionDuration to KafkaChannel CRD

### LINKED ISSUE #817: [KafkaChannels] Topic Configuration Inconsistencies
**Describe the bug**
This issue is a combination of a bug and a feature request, all pertaining to the same area of the code, which will require a refactor to align the behavior on a consistent approach.

First the bug...

The ConfigMap, shared by the two KafkaChannel implementations, includes a section **eventing-kafka -> kafka -> topic** which contains "default" values used when creating Kafka topics (NumPartitions, ReplicationFactor, RetentionMillis). These values are ONLY "defaults" to be used IF the KafkaChannel CR did not include them in the Spec.  The bug is that the default **NumPartitions** and **ReplicationFactor** are NEVER used, because the defaulting WebHook implementation will instead set KafkaChannel CR instances using hardcoded constants.  So, by the time the defaults from the ConfigMap are to be used, the KafkaChannel CR always has values and doesn't need the ConfigMap settings.

Next the feature...

The reason the **RetentionMillis** default in the ConfigMap (as described above) IS used, is because that value is NOT exposed in the KafkaChannel Spec with the other two fields.  There is no good reason for this omission other than the rich history of merging the two KafkaChannel implementations together ; ).  Excluding this config value from the KafkaChannel Spec prevents users from creating Topics with different retention periods.  We should expose the Kafka Topic Retention time in the KafkaChannel Spec with the other two fields, and handle them all in a consistent manner.  A PR (#722) was previously created to do this, but had some shortcomings.  It would be good to create a more comprehensive solution that addressed the general handling of these Topic configurations.

 **Expected behavior**
The proposed solution for aligning the handling of Topic configuration is to...
- Remove the **eventing-kafka -> kafka -> topic** section from the ConfigMap
- Add the **RetentionDuration** (as an ISO-8601 Duration String) to the KafkaChannel Spec
- Add Validation and Parsing of the new **RetentionDuration** ISO-8601 value.
- Add the **RetentionDuration** to the Defaulting WebHook to set default values to the 7 day period previously used in the base ConfigMap default.
- Remove the no-op logic of checking the ConfigMap if values in KafkaChannel are not set (since the Defaulting WebHook guarantees they will be set)
- Create a Post-Install job that will populate existing KafkaChannel instances with the **RetentionDuration** value matching their current configuration state.

 **To Reproduce**
- Create a KafkaChannel without any Spec and notice that the defaults in the ConfigMap are not used.
- Also, notice that the KafkaChannel Spec does not include the Retention value.

 **Knative release version**
All releases - /main as of this PR.

 **Additional context**
The plan is to add the new **RetentionDuration** field to the KafkaChannel Spec, while keeping the KafkaChannel version on **v1beta1**.  The handling of existing KafkaChannels (who have not specified this value) will operate as follows...

- Existing KafkaChannels (prior to upgrading eventing-kakfa) will have already been reconciled, with the corresponding Kafka Topic having been created using the ConfigMap default.  Future re-reconciliations of the KafkaChannel do NOT re-create the Kafka Topic and so it's configuration will not be modified.
- If a manual edit is made to the KafkaChannel (adding a label for instance), then the Defaulting WebHook will update the Spec with the default value.  If the user had previously changed the default/base **RetentionMillis** value in the ConfigMap, then these two values will not match, and the KafkaChannel CR Spec will not accurately reflect the true value configured in Kafka for that Topic.  The actual Kafka Topic will not be modified however, and will continue to operate as it had before.  The new default value would only ever take effect if the Topic was somehow torn-down and the KafkaChannel re-reconciled, which is not a supported production use case.  
- If desired, users can protect themselves from the above situation, by running the **config/channel/post-install** job to bring all existing KafkaChannels into alignment with the current Kafka Topic configuration.

**Tasks**
- [X] Add RetentionDuration to KafkaChannel CRD and remove Topic configuration from ConfigMap. (PR #828)
- [X] Close PR #722
- [X] Implement post-install job to populate RetentionDuration of existing KafkaChannels from current Topic configuration. (PR #834)
- [X] Close Issue #813 as no longer relevant.
- [X] Create a follow-up Issue to remove the post-install job once v0.26.0 has been built. (Issue #846)

travis-minke-sap:
This is part of the solution for Issue #817.

## Proposed Changes

- 🗑️   Remove "topic" fields from KafkaChannel ConfigMap
- 🎁  Add RetentionDuration field to KafkaChannel Spec (CRD)
- 🐛  Default RetentionDuration field in Defaulting WebHook & remove broken usage of ConfigMap fields.

**Release Note**

```release-note
KafkaChannel CRD Spec now includes a RetentionDuration field, allowing per-channel control over retention.  A Post-Install Job (config/post-install/retentionupdate) is available to migrate existing KafkaChannels forward by populating this field.  The "topic" configuration defaults which were not working have been removed from the config-kafka ConfigMap.
```

Holding for review...
/hold


config/channel/resources/kafkachannel-crd.yaml
```diff
@@ -39,15 +39,18 @@ spec:
...
               properties:
                 numPartitions:
-                  description: NumPartitions is the number of partitions of a Kafka topic. By default, it is set to 1.
+                  description: 'NumPartitions is the number of partitions of a Kafka topic. By default, it is set to 1.'
                   type: integer
                   format: int32
                   default: 1
                 replicationFactor:
-                  description: ReplicationFactor is the replication factor of a Kafka topic. By default, it is set to 1.
+                  description: 'ReplicationFactor is the replication factor of a Kafka topic. By default, it is set to 1.'
```
matzew, line 47: why do we add the `'` only to those two? Should other description do that too - for openAPI? If so. let's do that separately

pkg/apis/messaging/v1beta1/kafka_channel_types_test.go
```diff
@@ -47,3 +49,54 @@ func TestKafkaChannelGetStatus(t *testing.T) {
 		t.Errorf("GetStatus did not retrieve status. Got=%v Want=%v", config.GetStatus(), status)
 	}
 }
+
+func TestKafkaChannelSpec_ParseRetentionDuration(t *testing.T) {
```
matzew, line 53: nit: let's not the `_` - let's do camelCase

pkg/apis/messaging/v1beta1/kafka_channel_types_test.go
```diff
@@ -33,7 +35,7 @@ func TestKafkaChannel_GetGroupVersionKind(t *testing.T) {
 	}
 }
 
-func TestKafkaChannelGetStatus(t *testing.T) {
+func TestKafkaChannel_GetStatus(t *testing.T) {
```
matzew, line 38: nit: let's not the `_` - let's do camelCase

matzew:
Overall LGTM

I am happy we address https://github.com/knative-sandbox/eventing-kafka/issues/813  - and keep these configs on the `CR` for the KafkaChannel.

Before we had implicitly used the default, now it's explicitly possible to set a retention val. on the `CR` for the channel - seems good.

FWIW, I see there is a different PR for a post-install, so once that is in too, we might want to think if we must change a bit of the upgrade testing here... /cc @cardil

pkg/apis/messaging/v1beta1/kafka_channel_types_test.go
travis-minke-sap: Yeah, assuming that was just IntelliJ - sorry - will fix!

pkg/apis/messaging/v1beta1/kafka_channel_types_test.go
travis-minke-sap: The above instance was IntelliJ - not sure how this crept in though - good catch - thanks - will fix!

travis-minke-sap:
> FWIW, I see there is a different PR for a post-install, so once that is in too, we might want to think if we must change a bit of the upgrade testing here... /cc @cardil

Yes, as requested I separated that PR.  It is ready to go once once this PR goes in.  I am unfamiliar with any upgrade test impacts (thanks for calling it out) - please advise ; )

config/channel/resources/kafkachannel-crd.yaml
travis-minke-sap: Yeah, you're right... I noticed that some of the `description` fields were quoted and didn't bother to understand why you quote things in YAML.  Seems it's only recommended when necessary to prevent values from being misinterpreted as fields (when they contain a colon or similar characters).  There are probably some `description` fields that could be unquoted but, as you said we can leave that for another time and keep these changes to the bare minimum.  thanks!

travis-minke-sap:
seems stuck?
/test all

travis-minke-sap:
This PR is busted - github actions are stuck queueing and I haven't been able to un-stick them so.... closing this PR and creating a new one.  All comments contained in this PR have been addressed - fwiw ; )

PR #821, opened 2021-08-23
Post-Install Release Artifacts Bug

### LINKED ISSUE #749: Post install for 0.23 not generated
**Describe the bug**
The `post-install` files on this branch: https://github.com/knative-sandbox/eventing-kafka/tree/release-0.23/config/post-install
are not showing up on the "assets" of matching releases: https://github.com/knative-sandbox/eventing-kafka/releases/tag/v0.23.3

I do see this, however:   https://github.com/knative-sandbox/eventing-kafka/blob/release-0.23/hack/release.sh#L52-L58

 **Expected behavior**
Generator works as expected

 **To Reproduce**

 **Knative release version**

 **Additional context**

travis-minke-sap:
Fixes #749 

## Proposed Changes
- 🐛  Use **REPO_ROOT_DIR** instead of **YAML_REPO_ROOT**.  The latter was copied from _generate-yamls.sh_ which we are not using and so the value was not set.
- 🎁  Support more granular "channel" and "source" level post-install artifacts, so that users can more easily decide if they are necessary.

### DIFF
--- a/hack/release.sh
+++ b/hack/release.sh
@@ -46,16 +46,22 @@ function build_release() {
   local all_yamls=()
   for yaml in "${!COMPONENTS[@]}"; do
     local config="${COMPONENTS[${yaml}]}"
-    echo "Building Knative Eventing Contrib - ${config}"
+    echo "Building Knative Eventing Kafka - ${config}"
     ko resolve ${KO_FLAGS} -f ${config}/ | "${LABEL_YAML_CMD[@]}" > ${yaml}
     all_yamls+=(${yaml})
   done
 
-  if [ -d "${YAML_REPO_ROOT}/config/post-install" ]; then
-    echo "Resolving post install manifests"
+  if [ -d "${REPO_ROOT_DIR}/config/post-install/channel" ]; then
+    echo "Resolving channel post-install manifests"
+    local yaml="channel-post-install.yaml"
+    ko resolve ${KO_FLAGS} -f config/post-install/channel | "${LABEL_YAML_CMD[@]}" > ${yaml}
+    all_yamls+=(${yaml})
+  fi
 
-    local yaml="post-install.yaml"
-    ko resolve ${KO_FLAGS} -f config/post-install/ | "${LABEL_YAML_CMD[@]}" > ${yaml}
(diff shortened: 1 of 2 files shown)


hack/release.sh
```diff
@@ -46,16 +46,22 @@ function build_release() {
...
     local config="${COMPONENTS[${yaml}]}"
-    echo "Building Knative Eventing Contrib - ${config}"
+    echo "Building Knative Eventing Kafka - ${config}"
     ko resolve ${KO_FLAGS} -f ${config}/ | "${LABEL_YAML_CMD[@]}" > ${yaml}
     all_yamls+=(${yaml})
   done
 
-  if [ -d "${YAML_REPO_ROOT}/config/post-install" ]; then
-    echo "Resolving post install manifests"
+  if [ -d "${REPO_ROOT_DIR}/config/channel/post-install" ]; then
```
matzew, line 54: :man_facepalming:  
haha - good catch!

Since you are on it.  Mind _align_ a bit more like here:

https://github.com/knative/eventing/blob/main/hack/generate-yamls.sh#L112-L120

thanks!

travis-minke-sap: I'm happy to make changes, but am not sure how far you're suggesting we go to make this look like the code you referenced?  For example...

1. Do you just want to re-order the variable print lines, and use "readonly" instead of "local" and use capital letter variable names, OR...
2. Are you suggesting we create a temporary directory (**YAML_OUTPUT_DIR**) to place the YAML into, and that we should perform the handling of **KO_YAML_FLAGS** based on gcr.io, etc?

I tried to make these changes match the rest of the minimal content in the current _release.sh_ script rather than the _generate-yamls_ script from eventing.   If you are suggesting number 2 above, then would we apply the same logic (ko flags and yaml dir, etc.) to the generation of the "standard" yaml files (channels/source/etc) for consistency?  Essentially re-writing the _release.sh_ script?

thanks!

matzew: yeah., we should see to vendor this.

Also, I think it's _wise_ to have a permanent `post job`, see: https://github.com/knative/eventing/pull/5664

matzew (approved):
/lgtm
/approve

PR #822, opened 2021-08-24
control maximum allowed vreplica count

steven0711dong:
Fixes #

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**


pkg/common/scheduler/scheduler.go
```diff
@@ -40,6 +40,8 @@ type Evictor func(vpod VPod, from *duckv1alpha1.Placement) error
 type Scheduler interface {
 	// Schedule computes the new set of placements for vpod.
 	Schedule(vpod VPod) ([]duckv1alpha1.Placement, error)
+	GetVReplicaMPS() int32
+	GetMaxEventPerSecondPerPartition() int32
```
lionelvillard, line 44: the scheduler is a generic component that does not know anything about Kafka. Please remove.

pkg/source/reconciler/mtsource/kafkasource.go
```diff
@@ -120,14 +120,14 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, src *v1beta1.KafkaSource
...
-	err = client.InitOffsets(ctx, c, src.Spec.Topics, src.Spec.ConsumerGroup)
+	err, totalPartitions := client.InitOffsets(ctx, c, src.Spec.Topics, src.Spec.ConsumerGroup)
 	if err != nil {
 		logging.FromContext(ctx).Errorw("unable to initialize consumergroup offsets", zap.Error(err))
 		src.Status.MarkInitialOffsetNotCommitted("OffsetsNotCommitted", "Unable to initialize consumergroup offsets: %v", err)
 		return err
 	}
 	src.Status.MarkInitialOffsetCommitted()
-
+	src.Status.MaxAllowedVReplicas = totalPartitions*r.scheduler.GetMaxEventPerSecondPerPartition()/r.scheduler.GetVReplicaMPS() + 1
```
lionelvillard, line 130: only the Kafka controller should know about MaxEventPerSecondPerPartition and VReplicaMPS, not the scheduler.

pkg/common/scheduler/scheduler.go
steven0711dong: How do I pass the two new env vars to `pkg/source/reconciler/mtsource/kafkasource.go`?

pkg/source/reconciler/mtsource/kafkasource.go
steven0711dong: ok will change

pkg/source/client/offsets.go
```diff
@@ -32,25 +32,26 @@ import (
 // is closed before at least one message is consumed from ALL partitions.
 // Without InitOffsets, an event sent to a partition with an uninitialized offset
 // will not be forwarded when the session is closed (or a rebalancing is in progress).
-func InitOffsets(ctx context.Context, kafkaClient sarama.Client, topics []string, consumerGroup string) error {
+func InitOffsets(ctx context.Context, kafkaClient sarama.Client, topics []string, consumerGroup string) (error, int32) {
```
lionelvillard, line 35: in idiomatic go the error is always the last returned value.

pkg/source/reconciler/mtsource/controller.go
```diff
@@ -124,7 +145,7 @@ func NewController(
 		logging.FromContext(ctx).Debugw("Patched resource", zap.Any("patch", patch), zap.Any("patched", patched))
 		return nil
 	}
-
+	fmt.Println("Two new envs are: ", env.MaxEventPerSecondPerPartition, " :", env.VReplicaMPS)
```
lionelvillard, line 148: please remove.

pkg/source/reconciler/mtsource/controller.go
```diff
@@ -58,8 +79,8 @@ func NewController(
 ) *controller.Impl {
 	logger := logging.FromContext(ctx)
 
-	env := &envConfig{}
-	if err := envconfig.Process("", env); err != nil {
+	env, err := GetConfig()
+	if err != nil {
```
lionelvillard, line 83: not sure I understand why changing here. Instead of a global var I suggest to pass the value as a function parameter.

steven0711dong: which function? No function is called in mtsource Kafkasource.go belongs to this file.

pkg/source/client/offsets_test.go
```diff
@@ -138,7 +138,7 @@ func TestInitOffsets(t *testing.T) {
 			}
 			defer sc.Close()
 			ctx := logtesting.TestContextWithLogger(t)
-			err = InitOffsets(ctx, sc, tc.topics, group)
+			_, err = InitOffsets(ctx, sc, tc.topics, group)
```
lionelvillard, line 141: can you assert the returned number of partitions?

pkg/source/reconciler/mtsource/kafkasource.go
```diff
@@ -120,14 +124,14 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, src *v1beta1.KafkaSource
...
-	err = client.InitOffsets(ctx, c, src.Spec.Topics, src.Spec.ConsumerGroup)
+	totalPartitions, err := client.InitOffsets(ctx, c, src.Spec.Topics, src.Spec.ConsumerGroup)
 	if err != nil {
 		logging.FromContext(ctx).Errorw("unable to initialize consumergroup offsets", zap.Error(err))
 		src.Status.MarkInitialOffsetNotCommitted("OffsetsNotCommitted", "Unable to initialize consumergroup offsets: %v", err)
 		return err
 	}
 	src.Status.MarkInitialOffsetCommitted()
-
+	src.Status.MaxAllowedVReplicas = totalPartitions*r.MaxEventPerSecondPerPartition/r.VReplicaMPS + 1
```
lionelvillard, line 134: by default there should be no limits. I would make MaxEventPerSecondPerPartition and VReplicaMPS optional and default to -1. If set then set MaxAllowedVReplicas, otherwise do nothing (no max)

pkg/apis/duck/v1alpha1/placement_types.go
```diff
@@ -27,7 +27,8 @@ import (
 // Placeable is a list of podName and virtual replicas pairs.
 // Each pair represents the assignment of virtual replicas to a pod
 type Placeable struct {
-	Placement []Placement `json:"placements,omitempty"`
+	MaxAllowedVReplicas int32
```
lionelvillard, line 30: the `json: ...` tag is missing.

lionelvillard: and please make MaxAllowedVReplicas optional (`*int32`)

pkg/apis/sources/v1beta1/kafka_scheduling.go
```diff
@@ -32,6 +32,9 @@ func (k *KafkaSource) GetVReplicas() int32 {
 	if k.Spec.Consumers == nil {
 		return 1
 	}
+	if *k.Spec.Consumers > k.Status.MaxAllowedVReplicas {
```
lionelvillard, line 35: make sure to check `MaxAllowedVReplicas` is set

pkg/source/client/offsets_test.go
steven0711dong: what do you mean assert the returned number of partitions?

lionelvillard: instead of ignoring the returned total number of partitions, just check against the expected value.

PR #823, opened 2021-08-24
Plumb through using the `controller.Impl` tracker

matzew:

Similar to upstream PR: https://github.com/knative/eventing/pull/5656

/kind cleanup

- :broom: Update or clean up current behavior

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/source/reconciler/binding/controller.go
+++ b/pkg/source/reconciler/binding/controller.go
@@ -36,7 +36,6 @@ import (
 	"knative.dev/pkg/controller"
 	"knative.dev/pkg/injection/clients/dynamicclient"
 	"knative.dev/pkg/logging"
-	"knative.dev/pkg/tracker"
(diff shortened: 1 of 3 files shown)


pkg/source/reconciler/binding/controller.go
```diff
@@ -88,7 +87,7 @@ func NewController(
 	kfkInformer.Informer().AddEventHandler(controller.HandleAll(impl.Enqueue))
 	namespaceInformer.Informer().AddEventHandler(controller.HandleAll(impl.Enqueue))
```
mattmoor, line 88: Hmm, not your change, but this is a "code smell", and without a comment, I'd venture a guess that there's a bug here.  Generally there should be exactly one type/informer that queues its own key, so having two here feels off.  Happy to discuss in slack, if it'd be helpful.

mattmoor (commented):
/lgtm

PR #828, opened 2021-08-24
Add retention time to kafkachannel crd

travis-minke-sap:
This is part of the solution for Issue #817.  Original PR for this delivery (#818) was closed due to stuck github actions - all comments from that PR had already been addressed.

## Proposed Changes

- 🗑️   Remove "topic" fields from KafkaChannel ConfigMap
- 🎁  Add RetentionDuration field to KafkaChannel Spec (CRD)
- 🐛  Default RetentionDuration field in Defaulting WebHook & remove broken usage of ConfigMap fields.

**Release Note**

```release-note
KafkaChannel CRD Spec now includes a RetentionDuration field, allowing per-channel control over retention.  A Post-Install Job (config/post-install/retentionupdate) is available to migrate existing KafkaChannels forward by populating this field.  The "topic" configuration defaults which were not working have been removed from the config-kafka ConfigMap.
```

Holding for review...
/hold

/assign @devguyio

PR #615, opened 2021-05-11
 post-install script to migrate KafkaSource to v1beta1

matzew:
Signed-off-by: Matthias Wessendorf <mwessend@redhat.com>

Followup of #452

## Proposed Changes

- 🧽 Added a post-install migration job to migrate KafkaSource to v1beta1

**Release Note**

```release-note
🧽 Added a post-install migration job to migrate KafkaSource to v1beta1. You must run it after installing eventing-kafka 0.23
```

**Docs**

PR #834, opened 2021-08-25
[KafkaChannels] RetentionDuration v0.26.0 post-install job

travis-minke-sap:
This is part of the solution for Issue #817.

This PR includes a KafkaChannel post-install which updates existing KafkaChannel instances to include the new RetentionDuration Spec field based on the current Kafka Topic configuration.  This job is "optional" in that the Kafka Topic will continue to function with the same configuration, and the job only updates the KafkaChannel CRD to accurately represent the configuration.  If the job is not run, however, any subsequent update of the KafkaChannel will result in the new RetentionDuration field being set to the default value of PT168H (precise 1 week).  This will not update the actual Topic in any way and might or might not accurately reflect the Topic's configuration.

## Proposed Changes
- 🎁  New post-install job for both KafkaChannels which updates the KafkaChannel custom resources to accurately reflect the current Kafka Topic retention configuration.

**Note** - The release note was added in #828.  See the `config/channel/post-install/README.md` for instructions.

### DIFF
--- a/cmd/channel/post-install/updater.go
+++ b/cmd/channel/post-install/updater.go
@@ -134,9 +134,7 @@ func getKafkaChannelNamespacedName(topicName string) (*types.NamespacedName, err
 	}
 
 	// Strip The Topic Name Prefix Used By Consolidated KafkaChannel
-	if strings.HasPrefix(topicName, "knative-messaging-kafka.") {
-		topicName = strings.TrimPrefix(topicName, "knative-messaging-kafka.")
-	}
+	topicName = strings.TrimPrefix(topicName, "knative-messaging-kafka.")
 
 	// Split The KafkaChannel Namespace/Name From The Topic Name
 	topicNameComponents := strings.Split(topicName, ".")


config/channel/post-install/400-kafkachannel-update-job.yaml
```diff
@@ -0,0 +1,49 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: eventing-kafka-v0.26-channel-post-install-job
```
matzew, line 18: should we use `generatedName`?  (which would require `k create`, instead of `k apply`)

travis-minke-sap: > should we use generatedName? (which would require k create, instead of k apply)

Hmmm... I'm not sure what the best practice is in that regard?   From a quick internet scan I see the following...

**PROs...**
- Can re-run Job without first deleting old instance

**CONs...**
- Have to use `create` instead of `apply`
- Does not work with Helm (See [Issue](https://github.com/helm/helm/issues/3348#issuecomment-482369133))

Knative itself doesn't seem to have settled on anything, I saw the use of `generateName` in a Serving post-install job, but not in the most-recent Eventing post-install job (nor in most-recent `eventing-kafka` [post-install job](https://github.com/knative-sandbox/eventing-kafka/pull/615)).

I vote to leave it as is given the downsides and lack of a Knative standard.  As an example of potential user impact - we (SAP) would have to make a small change in some of our tooling to use `create` instead of `apply`.  If you feel strongly though I'm happy to change it.

matzew (approved):
/lgtm
/approve

PR #847, opened 2021-09-02
Make configmap reading func exported

aliok:
Fixes #

## Proposed Changes

- Make configmap reading func exported
-
-

**Release Note**

```release-note

```

**Docs**


aliok:
Ideally, we should extract these parts without Sarama configuration into a separate package but we're fine for now.

devguyio (approved):
/lgtm

PR #857, opened 2021-09-06
Make status.address optional in the KafkaChannel CRD

### LINKED ISSUE #744: `status.address` shouldn't be `required`
**Describe the bug**
When the data plane crashes, we have no indication about that in the KafkaChannel CR status. It is because `status.address` is set as `required` and in case of a data plane problem such as dispatcher not up, we don't have the address and Kube API server doesn't allow us to save the status w/o an address set.

 **Expected behavior**
`status.address` shouldn't be `required` so that we can set other status/conditions on the CR.

It is not required in other channels:

- Eventing generic `Channel`: https://github.com/knative/eventing/blob/9cd28581cf4d2c4c4dfeb97f984b84d3e04fc1ff/config/core/resources/channel.yaml#L162
- IMC: https://github.com/knative/eventing/blob/9cd28581cf4d2c4c4dfeb97f984b84d3e04fc1ff/config/channels/in-memory-channel/resources/in-memory-channel.yaml#L134

 **To Reproduce**
Example: create a consolidated CH with invalid bootstrap server in `kafka-configmap`.

 **Knative release version**

 **Additional context**

aliok:
Fixes #744

## Proposed Changes

- Make `status.address` optional in KafkaChannel CRD
- It is already marked as optional in the API code
- Check https://github.com/knative-sandbox/eventing-kafka/issues/744 for more details

**Release Note**

```release-note
- 🐛 Fix bug: `status.address` of KafkaChannel is not required anymore, thus conditions are reflected successfully in case of an early stage error
```

**Docs**

### DIFF
--- a/config/channel/resources/kafkachannel-crd.yaml
+++ b/config/channel/resources/kafkachannel-crd.yaml
@@ -147,8 +147,6 @@ spec:
             status:
               description: Status represents the current state of the KafkaChannel. This data may be out of date.
               type: object
-              required:
-                - address
               properties:
                 address:
                   type: object


aliok:
cc @matzew @travis-minke-sap @eric-sap @pierDipi

pierDipi (approved):

PR #866, opened 2021-09-14
Updating to Strimzi 0.25 version

matzew:

Fixes #

## Proposed Changes

- updating to 0.25 of Strimzi
- removing `authz` from the users, since it was NEVER enabled on the broker before. But its presence was causing errors on reconciling the users with 0.25 (somewhat related: https://cwiki.apache.org/confluence/display/KAFKA/KIP-504+-+Add+new+Java+Authorizer+Interface) 

**Release Note**

```release-note

```

**Docs**


matzew:
```
k get kafkausers -nkafka
NAME           CLUSTER      AUTHENTICATION   AUTHORIZATION   READY
my-sasl-user   my-cluster   scram-sha-512    simple          
my-tls-user    my-cluster   tls              simple
```

looks like the users aren't ready, hence there is no matching secret in the `kafka` NS

matzew:
adding explicit `authz` on the broker, due to API change on Kafka. Thanks @scholzj for the hint

Hoping for better test results now :pray:

matzew:
removing `authz` all together, since it was never enabled/used at all

pierDipi (approved):
Should we upgrade Kafka to version 2.8.0?

/lgtm
/hold if you want to upgrade Kafka

matzew:
@pierDipi we have it already at 2.8.0
https://github.com/matzew/eventing-kafka/blob/strimzi_0.25/test/config/100-kafka-ephemeral-triple-2.8.0.yaml#L7

/unhold

PR #876, opened 2021-09-17
Updating deps to include pkg fix

devguyio:
- Updating deps to include knative/pkg#2288


devguyio:
we need knative/pkg/pull#2290 as well since this will be overriden overnight by the bot as we're in release week.

aliok (approved):
/lgtm
/approve

PR #883, opened 2021-09-22
Move initOffsets functionality to a common package

aliok:
Fixes #

## Proposed Changes

- Move initOffsets functionality to a common package - so that we can also use it in Channels later
- Do not create the admin client within the function. Get that as parameter. In channel impl (which will use this func soon), we already have the adminClient and we would just pass that as a parameter.
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/common/kafka/offset/offsets.go
+++ b/pkg/common/kafka/offset/offsets.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package client
+package offset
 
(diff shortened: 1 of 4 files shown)


aliok:
cc @lionelvillard

aliok:
/retest

Infra

devguyio (approved):
/lgtm
/approve

aliok:
I will manually cherry pick this onto 0.24

PR #885, opened 2021-09-22
[0.24] Move initOffsets functionality to a common package

aliok:
Cherry pick https://github.com/knative-sandbox/eventing-kafka/pull/883

- Return partition count in `initOffsets` function, but do not use it as it is not used in 0.24 codebase


aliok:
cc @devguyio @lionelvillard 

Cherry pick https://github.com/knative-sandbox/eventing-kafka/pull/883

PR #886, opened 2021-09-22
[0.24] KafkaChannel to init offsets before dispatcher 

### LINKED ISSUE #549: [Consolidated Channel] events potentially undelivered if kafka-ch-dispatcher restarted before first events sent
**Describe the bug**

In a following scenario:
1. Subscription to a KafkaChannel gets Ready
2. kafka-ch-disptacher is restarted (e.g. pod delete)
3. events are sent just as new kafka-ch-disptacher starts

Some of the first events are not delivered to the subscriber.

It seems that if no events were yet commited, the new kafka-ch-dispatcher starts partition consumers at the "-1" offset, even if the subscription was already Ready before.

```
{"level":"info","ts":"2021-04-18T08:05:05.639Z","logger":"kafkachannel-dispatcher","caller":"consumer/consumer_handler.go:76","msg":"Starting partition consumer, topic: knative-messaging-kafka.foobar.channel-eupfwclx, partition: 0, initialOffset: -1","knative.dev/pod":"kafka-ch-dispatcher-5dbdb9bc75-z9xsx","ConsumeGroup":"kafka.foobar.channel-eupfwclx.6231d7a9-9741-4a5c-9cd6-91b7e2fbe64e"}
```

It should probably start at "0" in that case?

 **Expected behavior**
Events should be delivered even if kafka-ch-dispatcher was restarted just before first events were sent

 **To Reproduce**

Reproducer: https://github.com/maschmid/kafkachannel-subscription-ready-before-partition-consumers-started ( specifically TestDispatcherRestartBeforeCommitedEvents )

```
make apply
go test -v -run TestDispatcherRestartBeforeCommitedEvents
```

 **Knative release version**
0.20

 **Additional context**
Reproduced on OCP 4.7.6, knative installed with Serverless Operator, with Kafka installed on the OCP cluster with Strimzi.

aliok:
Fixes #549

## Proposed Changes

- Init offsets, similar to KafkaSource, to make sure dispatcher doesn't fetch events from latest when it starts (more details on the issue)
- Consumer groups are now started after making sure the initial offsets are committed by the controller(s)
- **NOTE**: Check for the initial offsets are done in the ConsumerFactory. This means, KafkaSource adapters and distributed channel **can** get this functionality for free. However, I didn't enable it for them.
- Get rid of status probing mechanism for subscription readiness in Kafka Channel
- Needed to touch some `distributed` channel code as I changed some interfaces. But, no business logic changes

**Release Note**

```release-note
- 🐛  Fixed a bug that was causing Kafka Channel dispatcher to miss some events in case the offset for the consumer group was not initialized
```

**Docs**


aliok:
/retest

Wtf

aliok:
/retest

Let's see what happens

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
```diff
@@ -477,7 +492,20 @@ func (r *Reconciler) reconcileChannelService(ctx context.Context, dispatcherName
 	return svc, nil
 }
 
-func (r *Reconciler) createClient(ctx context.Context) (sarama.ClusterAdmin, error) {
+func (r *Reconciler) createClients(ctx context.Context) (sarama.Client, sarama.ClusterAdmin, error) {
+	kafkaClient := r.kafkaClient
```
devguyio, line 496: you're not setting 
`r.kafkaClient` anywhere, is that intentional? wouldn't that mean you're creating a new client with every reconcile loop?

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
```diff
@@ -249,12 +250,26 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, kc *v1beta1.KafkaChannel
 	return newReconciledNormal(kc.Namespace, kc.Name)
 }
 
-func (r *Reconciler) reconcileSubscribers(ctx context.Context, ch *v1beta1.KafkaChannel) error {
+func (r *Reconciler) reconcileSubscribers(ctx context.Context, ch *v1beta1.KafkaChannel, kafkaClient sarama.Client, kafkaClusterAdmin sarama.ClusterAdmin) error {
 	after := ch.DeepCopy()
 	after.Status.Subscribers = make([]v1.SubscriberStatus, 0)
+	logger := logging.FromContext(ctx)
 	for _, s := range ch.Spec.Subscribers {
```
devguyio, line 257: A good optimization once we get this working is to do this only for the diff between `ch.Spec.Subscribers` and ready ones of `ch.Status.Subscribers` . Cause now with the smallest change in the channel, you're hitting kafka with api calls for every single subscription. 

Or in `reconcileInitialOffset` you can check if the sub is not ready then initialize the offsets.

devguyio (commented):
I did a short review, will continue later today.

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
aliok: I refactored the code that creates the admin client and created the normal client there.

For clusterAdmin, there's this ticket: https://github.com/Shopify/sarama/issues/1162. I see that ticket didn't really get any resolution.

I haven't tried reusing the regular kafkaClient. I can try that, once I made the failing tests working.

aliok:
/test pull-knative-sandbox-eventing-kafka-integration-test-channel-consolidated

Retesting to see if there's a pattern

pkg/channel/consolidated/reconciler/controller/controller.go
```diff
@@ -174,22 +159,23 @@ func NewController(
 			knativeReconciler.LabelFilterFunc(roleLabelKey, dispatcherRoleLabelValue, false),
 		),
 		Handler: cache.ResourceEventHandlerFuncs{
-			// Cancel probing when a Pod is deleted
-			DeleteFunc: getPodInformerEventHandler(ctx, logger, statusProber, impl, kafkaChannelInformer, "Delete"),
-			AddFunc:    getPodInformerEventHandler(ctx, logger, statusProber, impl, kafkaChannelInformer, "Add"),
+			// TODO: do we still need this?
+			// Global sync when a Pod is deleted or added
+			DeleteFunc: getPodInformerEventHandler(ctx, logger, impl, kafkaChannelInformer, "Delete"),
+			AddFunc:    getPodInformerEventHandler(ctx, logger, impl, kafkaChannelInformer, "Add"),
```
matzew, line 165: It was introduced as a change on top of adding the prober... which we now remove, hence I _think_ we can remove the hndler here

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
matzew: pretty bad sarama bug :disappointed:

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
matzew: I think I like the suggestion for doing the check inside the `reconcileInitialOffset`

pkg/common/consumer/consumer_factory.go
```diff
@@ -64,11 +72,38 @@ func (c *customConsumerGroup) Close() error {
...
+	defer clusterAdmin.Close()
+
+	// this is a blocking func
+	// do not proceed until the check is done
+	err = c.kcoi.checkOffsetsInitialized(ctx, groupID, topics, logger, client, clusterAdmin)
 	if err != nil {
 		return nil, err
 	}
+
+	logger.Infow("All offsets are initialized", zap.Any("topics", topics), zap.Any("groupID", groupID))
```
matzew, line 105: do we really want this info?

pkg/common/consumer/consumer_offset_initalizer.go
```diff
@@ -0,0 +1,43 @@
...
+	"context"
+	"fmt"
+
+	"github.com/Shopify/sarama"
+	"go.uber.org/zap"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"knative.dev/eventing-kafka/pkg/common/kafka/offset"
+)
+
+type consumerOffsetInitializer interface {
```
devguyio, line 13: 1. this name gives me the indication that it is an "initializer" , while it's actually only checking.
2. consumer is not a consumer group. 

So maybe rename it to something like "CGOffsetsMonitor" or something like that, "Poller" , "Checker" .

pkg/common/kafka/offset/offsets.go
```diff
@@ -101,3 +93,61 @@ func InitOffsets(ctx context.Context, kafkaClient sarama.Client, kafkaAdminClien
...
+	// Retrieve all partitions
+	topicPartitions := make(map[string][]int32)
+	for _, topic := range topics {
+		partitions, err := kafkaClient.Partitions(topic)
+		totalPartitions += len(partitions)
+		if err != nil {
+			return -1, nil, fmt.Errorf("failed to get partitions for topic %s: %w", topic, err)
+		}
+
+		topicPartitions[topic] = partitions
```
devguyio, line 150: I think it's safer to assign a copy of the partitions slice instead of using the one returned by the client (I traced the code, it's cached in Sarama so I think we should play it safe and just copy it here)

pkg/common/kafka/offset/offsets.go
```diff
@@ -101,3 +93,61 @@ func InitOffsets(ctx context.Context, kafkaClient sarama.Client, kafkaAdminClien
...
+	}
+
+	for topic, partitions := range offsets.Blocks {
+		for partitionID, block := range partitions {
+			if block.Offset == -1 { // not initialized?
+				partitionsOffsets, ok := topicOffsets[topic]
+				if !ok {
+					// topic may have been deleted. ignore.
+					continue
+				}
```
devguyio, line 122: Why are you handling this case here? why can't it just focus on the offsets, and if it's not there then `CheckIfAllOffsetsInitialized` returns simply false. (i.e. this function assuming that topics exists, and leaves it up to the caller to only call it for topics that exist)  I feel this is too much assumption here (that if the topic is deleted then that's ok)

devguyio (commented):
Round two, wanna send the comments in batches.

pkg/common/kafka/offset/offsets.go
```diff
@@ -101,3 +93,61 @@ func InitOffsets(ctx context.Context, kafkaClient sarama.Client, kafkaAdminClien
...
+
+				_, ok = partitionsOffsets[partitionID]
+				if !ok {
+					// partition may have been deleted. ignore.
+					continue
+				}
+
+				return false, nil
```
devguyio, line 130: same as above

pkg/common/kafka/offset/offsets.go
devguyio: I just noticed this is the same logic for the `InitOffset` method. Still wondering why we need to have it though.

pkg/channel/consolidated/reconciler/controller/controller.go
aliok: Done

pkg/common/consumer/consumer_factory.go
aliok: Changed to debug

pkg/common/consumer/consumer_offset_initalizer.go
```diff
@@ -0,0 +1,43 @@
...
+	checkOffsetsInitialized(ctx context.Context, groupID string, topics []string, logger *zap.SugaredLogger, client sarama.Client, clusterAdmin sarama.ClusterAdmin) error
```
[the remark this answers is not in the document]
aliok: Thanks, renamed to `ConsumerGroupOffsetsChecker`

pkg/common/consumer/consumer_factory.go
```diff
@@ -19,25 +19,33 @@ package consumer
...
+	"time"
 
 	"github.com/Shopify/sarama"
 	"go.uber.org/zap"
+	"knative.dev/pkg/logging"
 )
 
-// newConsumerGroup is a wrapper for the Sarama NewConsumerGroup function, to facilitate unit testing
+const OffsetInitRetryTimeout = 60 * time.Second
+const OffsetInitRetryInterval = 5 * time.Second
```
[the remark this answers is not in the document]
aliok: Done, thanks

pkg/common/consumer/consumer_offset_initalizer.go
aliok: LOL, you're right. I was also doing the offset initialization inside this interface but then didn't think about the changing the name.

It is changed now.

pkg/common/kafka/offset/offsets.go
aliok: @lionelvillard told me that we won't need to handle the deleted topic and partitions too in the check function. I deleted that code now.

He can maybe give more info on why we need to handle these cases in the `initoffsets` function.

pkg/common/kafka/offset/offsets.go
aliok: well this function is only used in this file and we certainly do not modify the array. ...but, let's play safe as you said for future generations.
This is done.

aliok:
cc @lionelvillard @devguyio 
PR is ready.

Also cc @travis-minke-sap . I haven't made any business logic change in any components other than the consolidated channel.

matzew:
Just tested

```
Receiving duration has passed!
Total received message count: 5455
Latest message index: 5455
Missing messages count: 0
Missing messages: 
Duplicate messages: {}
```

:tada: 

will now review 

/assign @matzew

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
```diff
@@ -578,7 +622,7 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, kc *v1beta1.KafkaChannel)
 	logger := logging.FromContext(ctx)
 	channel := fmt.Sprintf("%s/%s", kc.GetNamespace(), kc.GetName())
 	logger.Debugw("FinalizeKind", zap.String("channel", channel))
-	kafkaClusterAdmin, err := r.createClient(ctx)
+	_, kafkaClusterAdmin, err := r.createClients(ctx)
```
devguyio, line 625: you probably wanna capture the kafkaClient here and defere closing it

pkg/common/consumer/consumer_factory.go
```diff
@@ -64,13 +66,17 @@ func (c *customConsumerGroup) Close() error {
 var _ sarama.ConsumerGroup = (*customConsumerGroup)(nil)
 
 // StartConsumerGroup creates a new customConsumerGroup and starts a Consume goroutine on it
-func (c kafkaConsumerGroupFactoryImpl) StartConsumerGroup(groupID string, topics []string, logger *zap.SugaredLogger, handler KafkaConsumerHandler, options ...SaramaConsumerHandlerOption) (sarama.ConsumerGroup, error) {
+func (c kafkaConsumerGroupFactoryImpl) StartConsumerGroup(ctx context.Context, groupID string, topics []string, handler KafkaConsumerHandler, options ...SaramaConsumerHandlerOption) (sarama.ConsumerGroup, error) {
+	logger := logging.FromContext(ctx)
+
 	consumerGroup, err := c.createConsumerGroup(groupID)
 	if err != nil {
+		logger.Errorw("Unable to create consumer group", zap.Error(err))
```
devguyio, line 74: NIT: error message starting with lower case

pkg/common/consumer/consumergroup_offsets_checker.go
```diff
@@ -0,0 +1,74 @@
...
+
+type KafkaConsumerGroupOffsetsChecker struct {
+}
+
+func (k *KafkaConsumerGroupOffsetsChecker) WaitForOffsetsInitialization(ctx context.Context, groupID string, topics []string, logger *zap.SugaredLogger, addrs []string, config *sarama.Config) error {
+	logger.Infow("Checking if all offsets are initialized", zap.Any("topics", topics), zap.Any("groupID", groupID))
+
+	client, err := newSaramaClient(addrs, config)
+	if err != nil {
+		logger.Errorw("Unable to create Kafka client", zap.Error(err))
```
devguyio, line 42: 1. NIT error message starting with lower case
1. would be helpful to add the groupID and topics

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
```diff
@@ -249,12 +250,26 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, kc *v1beta1.KafkaChannel
...
 	after.Status.Subscribers = make([]v1.SubscriberStatus, 0)
+	logger := logging.FromContext(ctx)
 	for _, s := range ch.Spec.Subscribers {
-		if r, _ := r.statusManager.IsReady(ctx, *ch, s); r {
-			logging.FromContext(ctx).Debugw("marking subscription", zap.Any("subscription", s))
+		logger.Debugw("Reconciling initial offset for subscription", zap.Any("subscription", s))
+		err := r.reconcileInitialOffset(ctx, ch, s, kafkaClient, kafkaClusterAdmin)
+
+		if err != nil {
+			logger.Errorw("Reconcile failed fo initial offset for subscription", zap.Any("subscription", s), zap.Error(err))
```
travis-minke-sap, line 262: Typo `failed fo initial`

pkg/common/consumer/consumergroup_offsets_checker.go
```diff
@@ -0,0 +1,74 @@
...
+		return err
+	}
+	defer client.Close()
+
+	clusterAdmin, err := newSaramaClusterAdmin(addrs, config)
+	if err != nil {
+		logger.Errorw("Unable to create Kafka cluster admin client", zap.Error(err))
```
devguyio, line 49: same as above

pkg/common/consumer/consumergroup_offsets_checker.go
```diff
@@ -0,0 +1,74 @@
...
+	check := func() (bool, error) {
+		if initialized, err := offset.CheckIfAllOffsetsInitialized(client, clusterAdmin, topics, groupID); err == nil {
+			if initialized {
+				return true, nil
+			} else {
+				logger.Infow("Offsets not yet initialized, going to try again")
+				return false, nil
+			}
+		} else {
+			return true, fmt.Errorf("error checking if offsets are initialized. stopping trying. %w", err)
```
devguyio, line 63:
```suggestion
			return false, fmt.Errorf("error checking if offsets are initialized. stopping trying. %w", err)
```

I think this is the correct behavior when you want to stop retrying in the error case. `PollUntil` will stop when there's an error

pkg/common/consumer/consumer_factory.go
```diff
@@ -94,6 +101,16 @@ func (c kafkaConsumerGroupFactoryImpl) startExistingConsumerGroup(
 	ctx, cancel := context.WithCancel(context.Background())
 
 	go func() {
+		// this is a blocking func
+		// do not proceed until the check is done
+		err := c.offsetsChecker.WaitForOffsetsInitialization(ctx, groupID, topics, logger, c.addrs, c.config)
+		if err != nil {
+			logger.Errorw("Error while checking if offsets are initialized", zap.Error(err))
```
devguyio, line 108: 1. NIT error message starting with lowercase 
2. useful to add groupID and topics

devguyio (commented):
halfway through, submitting those and will continue

devguyio:
I left some comments above,
other than those it LGTM
nice work @aliok !!! 🎉

travis-minke-sap:
> Also cc @travis-minke-sap . I haven't made any business logic change in any components other than the consolidated channel.

Thanks for the heads-up Ali!

In general I'm ok with the approach, and the way you've allowed the distributed channel to NoOp seems fine.  Is there any reason the distributed channel wouldn't want to do this same initial offset committing?  I'm thinking I can create a separate Issue to add it to the distributed channel's dispatcher on `main/` once you get these changes up there?

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
```diff
@@ -249,12 +250,26 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, kc *v1beta1.KafkaChannel
...
+			logger.Errorw("marking subscription unready", zap.Any("subscription", s))
```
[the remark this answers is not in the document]
matzew: > error logs start with lower case.

I think serving/eventing are doing it with Uppercase.  E.g.:
* https://github.com/knative/eventing/blob/c820f4239730dacf5b2e0e5eed67a496bdc9b424/pkg/reconciler/apiserversource/apiserversource.go#L103
* https://github.com/knative/serving/blob/67429ed977913af216d28245fe24cbf326659538/cmd/queue/main.go#L248


As well +1 on adding the channel for debugging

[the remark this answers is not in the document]
matzew: ah... now I understand. With `error logs start with lower case.` you mean they are incorrectly starting with lowercase :smile:

pkg/common/consumer/consumer_factory.go
matzew: hrm - not sure what you mean here

pkg/common/consumer/consumergroup_offsets_checker.go
matzew: > NIT error message starting with lower case

the `Unable to create Kafka client` is starting with upper case... which is correct? :thinking:

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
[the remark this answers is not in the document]
matzew: looking at the other entries, regarding "lowercase", I am actually not sure

pkg/common/consumer/consumergroup_offsets_checker.go
matzew: I agree here too.  I'd also give a little comment above the  `	check := func() (bool, error)...` regarding this, and meanings

pkg/common/consumer/consumergroup_offsets_checker.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (c *NoopConsumerGroupOffsetsChecker) WaitForOffsetsInitialization(ctx context.Context, groupID string, topics []string, logger *zap.SugaredLogger, addrs []string, config *sarama.Config) error {
+	return nil
+}
+
+type KafkaConsumerGroupOffsetsChecker struct {
+}
+
+func (k *KafkaConsumerGroupOffsetsChecker) WaitForOffsetsInitialization(ctx context.Context, groupID string, topics []string, logger *zap.SugaredLogger, addrs []string, config *sarama.Config) error {
+	logger.Infow("Checking if all offsets are initialized", zap.Any("topics", topics), zap.Any("groupID", groupID))
```
matzew, line 38: should this be really Info?

pkg/common/consumer/consumergroup_offsets_checker.go
```diff
@@ -0,0 +1,74 @@
...
+		return err
+	}
+	defer clusterAdmin.Close()
+
+	check := func() (bool, error) {
+		if initialized, err := offset.CheckIfAllOffsetsInitialized(client, clusterAdmin, topics, groupID); err == nil {
+			if initialized {
+				return true, nil
+			} else {
+				logger.Infow("Offsets not yet initialized, going to try again")
```
matzew, line 59: I'd prob. also do this to debug - to reduce the noise?

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
matzew: Can we add a comment that this thing returns both clients? Normal and ClusterAdmin?

`func (r *Reconciler) createSaramaClients` instead?  not just client?

matzew: Do we really always need _both_?  See Ahmed's comment with stashing and defering the "regular" client.

Just wondering

pkg/common/consumer/consumergroup_offsets_checker.go
aliok:
```
PollUntil tries a condition func until it returns true, an error or stopCh is closed.
```

So, it won't matter if it is returning `true,err` or `false,err` as long as there's an error.

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
devguyio: one alternative can be that we have two functions, one for each.

pkg/common/consumer/consumer_factory.go
devguyio: I mean something like
```go
logger.Errorw("Unable to create consumer group", zap.Error(err), zap.Sting("Consumer Group", groupID))
```

pkg/common/consumer/consumergroup_offsets_checker.go
devguyio: There's a difference in the semantic meaning, returning `true, nil` means the check succeeded, stop retrying, returning `false, err` means the check failed with an error, stop retrying. 

Reading in the corde `true, err` is confusing knowing the semantic meaning behind it.

aliok:
> In general I'm ok with the approach, and the way you've allowed the distributed channel to NoOp seems fine. Is there any reason the distributed channel wouldn't want to do this same initial offset committing? I'm thinking I can create a separate Issue to add it to the distributed channel's dispatcher on main/ once you get these changes up there?

@travis-minke-sap 

Distributed channel can also do the same thing. I don't see any reason.
There are 2 steps:
- Committing the initial offsets in the channel controller using `offset.InitOffsets`
- Not starting consuming messages in the dispatcher until checking if the offsets are committed using `offset.CheckIfAllOffsetsInitialized`

I am going to forward port this PR all the way to the main. I don't think I would have the chance to implement it for distributed channel but at least there would be something in main in the consolidated channel that can be leveraged.

pkg/common/consumer/consumergroup_offsets_checker.go
aliok: Ok, changed it anyway

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
aliok: We now have 2 functions. Code looks much better, thanks for the suggestions.

About stashing into `r.kafkaClient` -> I will try this, one I see tests are passing again.

aliok:
/test pull-knative-sandbox-eventing-kafka-integration-test-channel-consolidated
/test pull-knative-sandbox-eventing-kafka-integration-test-channel-consolidated-sasl
/test pull-knative-sandbox-eventing-kafka-integration-test-channel-consolidated-tls

Stashing the kafka client broke things. I am gonna try here again, before reverting commit 4d57547.

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
aliok: Ok, when I stash the client to `r.kafkaClient` and reuse it, things break. I have no desire to debug why.

See my comment here: https://github.com/knative-sandbox/eventing-kafka/pull/886#issuecomment-932092014

You can have a look at the commit list here, https://github.com/knative-sandbox/eventing-kafka/pull/886/commits, and see the job history for each commit there (not always consistent, especially when there's new code pushed while the jobs were running).

aliok:
> Stashing the kafka client broke things. I am gonna try here again, before reverting commit 4d57547.

I reverted it now. See my comment here: https://github.com/knative-sandbox/eventing-kafka/pull/886#discussion_r720128806

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
aliok: Done now

aliok:
/test pull-knative-sandbox-eventing-kafka-unit-tests

TestEnableSaramaLogging

pkg/channel/consolidated/reconciler/controller/kafkachannel.go
matzew: > . I have no desire to debug why.

that is OK w/ me. That said - mind creating an issue we are having to track this?

aliok: https://github.com/knative-sandbox/eventing-kafka/issues/912

PR #888, opened 2021-09-22
README for new MT Scheduler with pluggable policies

aavarghese:
Signed-off-by: aavarghese <avarghese@us.ibm.com>

Continuation of #768

## Proposed Changes

- README

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/common/scheduler/README.md
+++ b/pkg/common/scheduler/README.md
@@ -1,0 +1,247 @@
+# Knative Eventing Multi-Tenant Scheduler
+
+[![LICENSE](https://img.shields.io/github/license/knative-sandbox/eventing-kafka.svg)](https://github.com/knative-sandbox/eventing-kafka/blob/master/LICENSE)
+[![Slack](https://img.shields.io/badge/Signup-Knative_Slack-white.svg?logo=slack)](https://slack.knative.dev)
+[![Slack](https://img.shields.io/badge/%23eventing-white.svg?logo=slack&color=522a5e)](https://knative.slack.com/archives/C9JP909F0)
+
+## Scheduler with High-Availability
+
+The [scheduler](#scheduler) is responsible for placing virtual replicas (vreplicas) onto real pods. It takes a list of (source, # of vreplicas) tuples and computes a set of Placements.
+Placement info are added to the source status.
+
+Scheduling strategies rely on pods having a sticky identity (see StatefulSet) and the current [State](#state-collector) of the cluster.
+
+When a vreplica cannot be scheduled it is added to the list of pending vreplicas. The [Autoscaler](#autoscaler) monitors this list and allocates more pods for placing it.
+
+To support high-availability the scheduler distributes vreplicas uniformly across failure domains such as zones/nodes/pods containing replicas from a StatefulSet.
+
+### Current Requirements
+
+1. High Availability: Vreplicas for a source must be evenly spread across domains to reduce impact of failure when a zone/node/pod goes unavailable for scheduling.*
+
+2. Equal message consumption: Vreplicas for a source must be evenly spread across adapter pods to provide an equal rate of processing messages. For example, Kafka spreads partitions equally across pods so if vreplicas aren’t equally spread, pods with fewer replicas will consume messages slower than others.
+
+3. Spread not more than partition size: Vreplicas for a Kafka source must be evenly spread across pods such that the total number of pods with placements does not exceed the number of Kafka partitions for the topic it's consuming from. Else, the additional pods have no partitions to consume messages from and waste K8s resources.
+
+* Note: StatefulSet anti-affinity rules guarantee new pods to be scheduled on a new zone and node.
+
+### Components:
+
+#### 1. Scheduler
+The scheduling framework has a pluggable architecture where plugins are registered and compiled into the scheduler. It allows many scheduling features to be implemented as plugins, while keeping the scheduling "core" simple and maintainable.
+
+Scheduling happens in a series of stages:
+
+**Filter**: These plugins (predicaates) are used to filter out pods where a vreplica cannot be placed. If any filter plugin marks the pod as infeasible, the remaining plugins will not be called for that pod. A vreplica is marked as unschedulable if no pods pass all the filters.
+
+**Score**: These plugins (priorities) provide a score to each pod that has passed the filtering phase. Scheduler will then select the pod with the highest weighted scores sum.
+
+Scheduler must be Knative generic with its core functionality implemented as core plugins. Anything specific to an eventing source will be implemented as separate plugins (for example, number of Kafka partitions)
+
+It allocates one vreplica at a time by filtering and scoring schedulable pods.
+
+A vreplica can be unschedulable for several reasons such as pods not having enough capacity, constraints cannot be fulfilled, etc.
+
+#### 2. Descheduler
+
+Similar to scheduler but has its own set of priorities (no predicates today).
+
+#### 3. Autoscaler
+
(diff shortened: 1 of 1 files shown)


lionelvillard:
@aavarghese can you fix the linter errors? thx!

pkg/common/scheduler/README.md
```diff
@@ -0,0 +1,241 @@
...
+                    "Weight": 2}
+                  ]
+```
+
+### Recovery
+
+Some failure scenarios are described below:
+
+1. **Pod failure**:
+When a pod/replica in a StatefulSet goes down due to some reason (but its node and zone are healthy), a new replica is spun up by the StatefulSet with the same pod identit (pod can come up on a different node) almost immediately.
```
pierDipi, line 137:
```suggestion
When a pod/replica in a StatefulSet goes down due to some reason (but its node and zone are healthy), a new replica is spun up by the StatefulSet with the same pod identity (pod can come up on a different node) almost immediately.
```

pkg/common/scheduler/README.md
```diff
@@ -0,0 +1,241 @@
...
+All existing vreplica placements will still be valid and no rebalancing is needed.
+There shouldn’t be any degradation in Kafka message processing.
```
pierDipi, line 139: This is not really true, a consumer group rebalance could degrade message processing especially when [Kafka Consumer Incremental Rebalance Protocol](https://cwiki.apache.org/confluence/display/KAFKA/KIP-429%3A+Kafka+Consumer+Incremental+Rebalance+Protocol) is not being used (which afaik is not implemented in Sarama).

pierDipi (commented):
Thanks, I love this document!

pkg/common/scheduler/README.md
aavarghese: @pierDipi the pod set being referred to here is only talking about the eventing scheduler adapter pods where vreplicas are placed. Since pod will restart, the same placements can be kept without a rebalancing of the vreps. 
I agree with you about the consumer group rebalancing and degradation but that may/may not happen here if the kafka pods are affected, as well.
I hope I'm not missing anything...

pierDipi: > @pierDipi the pod set being referred to here is only talking about the eventing scheduler adapter pods where vreplicas are placed. Since pod will restart, the same placements can be kept without a rebalancing of the vreps.

This is what `All existing vreplica placements will still be valid and no rebalancing is needed.` is saying, I agree and it's clear to me why.

I was referring to `There shouldn’t be any degradation in Kafka message processing.`.

> I agree with you about the consumer group rebalancing and degradation but that may/may not happen here if the kafka pods are affected, as well.
> I hope I'm not missing anything...

so, are you saying that if a pod where vreplicas are placed goes down that won't trigger a consumer group rebalance that affects message processing?

In the worst-case scenario, I'd expect something like this to happen (happy to be wrong):

1. Pod goes down
2. A new pod comes up (same name)
3. Kafka broker sees a new consumer that wants to join the group -> rebalance
4. Kafka detects that the consumer that was consuming messages in the dead pod (1) is not sending heartbeats anymore -> rebalance (again)

at least one rebalance happen. 2 in the worst case since `terminationGracePeriodSeconds = 0 < "time for Kafka to detect that a consumer is dead"`

Is the above not possible? If yes, does that count as a degradation for Kafka message processing?

aavarghese: You're right. This is absolutely possible. 

I made an assumption that the same sticky pod (when restarted) would have the same consumer member ID and using static membership, would get the same assignment or so.

I don't have any numbers to quantify the extent of degradation for these recovery scenarios. Will need to run some performance runs to measure latency. Thank you for catching this @pierDipi !!

pkg/common/scheduler/README.md
```diff
@@ -0,0 +1,247 @@
+# Knative Eventing Multi-Tenant Scheduler
+
+[![LICENSE](https://img.shields.io/github/license/knative-sandbox/eventing-kafka.svg)](https://github.com/knative-sandbox/eventing-kafka/blob/master/LICENSE)
+[![Slack](https://img.shields.io/badge/Signup-Knative_Slack-white.svg?logo=slack)](https://slack.knative.dev)
+[![Slack](https://img.shields.io/badge/%23eventing-white.svg?logo=slack&color=522a5e)](https://knative.slack.com/archives/C9JP909F0)
+
```
lionelvillard, line 6: since this is a sub-README, the badges are not strictly required.

pkg/common/scheduler/README.md
```diff
@@ -0,0 +1,247 @@
...
+## Scheduler with High-Availability
```
lionelvillard, line 7: Before this section, what about defining some of the new concepts? I'm thinking at the minimum `vreplicas` (short for virtual replicas) and vpod (short for virtual pod).

pkg/common/scheduler/README.md
```diff
@@ -0,0 +1,247 @@
...
+* Note: StatefulSet anti-affinity rules guarantee new pods to be scheduled on a new zone and node.
+
+### Components:
+
+#### 1. Scheduler
+The scheduling framework has a pluggable architecture where plugins are registered and compiled into the scheduler. It allows many scheduling features to be implemented as plugins, while keeping the scheduling "core" simple and maintainable.
+
+Scheduling happens in a series of stages:
+
+**Filter**: These plugins (predicaates) are used to filter out pods where a vreplica cannot be placed. If any filter plugin marks the pod as infeasible, the remaining plugins will not be called for that pod. A vreplica is marked as unschedulable if no pods pass all the filters.
```
lionelvillard, line 35: nit: predicaates -> predicates

lionelvillard: Did you mean to indent the `Filter` and `Score` paragraphs?

pkg/common/scheduler/README.md
```diff
@@ -0,0 +1,247 @@
...
+                    "Args": "{\"MaxSkew\": 2}"},
+                    {"Name": "RemoveWithAvailabilityZonePriority",
+                    "Weight": 10,
+                    "Args":  "{\"MaxSkew\": 2}"},
+                    {"Name": "RemoveWithHighestOrdinalPriority",
+                    "Weight": 2}
+                  ]
+```
+
+### Recovery
```
lionelvillard, line 132: In addition to recovery scenario, there should be a section about normal operation, such as Knative upgrade and K8s upgrade. I wouldn't do it in this PR, just something to keep in mind in the future.

pkg/common/scheduler/README.md
aavarghese: Added.

pkg/common/scheduler/README.md
aavarghese: Adding a preliminary section for this here. More details as we use KPerf can come in the future.

pkg/common/scheduler/README.md
```diff
@@ -0,0 +1,256 @@
...
+                    "Weight": 2}
+                  ]
+```
+
+## Normal Operation
+
+1. **Busy scheduler**:
+
+Scheduler can be very busy allocating the best placements for multiple eventing sources at a time using the scheduler predicates and priorities configured. During this time, the cluster could see statefulset replicas increasing (as the autoscaler computes how many more pods are needed to complete scheduling) or replicas decreasing during idle time (as eviction kicks in to attempt compacting the vreplicas into a smaller number of pods or as KEDA lowers the number of vreplicas). The current placements are stored in the eventing source's status field for observability.
+
```
lionelvillard, line 132: Instead of: `replicas decreasing during idle time (as eviction kicks in to attempt compacting the vreplicas into a smaller number of pods or as KEDA lowers the number of vreplicas).`, what about something like that:

`replicas decreasing during idle time, either caused by less events flowing through the system, or the vreplicas compaction or the deletion of sources.`. 

I would avoid mentioning KEDA here (it's fine to mention it as an example).

PR #895, opened 2021-09-27
[Channel] Set defaults of spec.delivery

pierDipi:
Like https://github.com/knative/eventing/pull/5748 but for KafkaChannel.

## Proposed Changes

- [Channel] Set defaults of spec.delivery
- Upgrade eventing to main `go get -u knative.dev/eventing@main`
- Add delete permissions to channelable manipulator cluster role
  - Eventing controller needs to delete channels when a step in
a sequence is removed.
    - knative/eventing#5718

**Release Note**

```release-note
The field `spec.delivery.deadLetterSink.ref.namespace` for KafkaChannel if not specified is defaulted to `metadata.namespace`.
```

/kind enhancement
/assign @aliok @matzew


matzew (approved):
/lgtm
/approve

thanks for doing this

PR #903, opened 2021-09-29
Adding noop migrator for kafkachannels

matzew:

## Proposed Changes

- :broom: Adding noop migrator for the `kafkachannels.messaging.knative.dev` to always have a migrator manifest on each release. 

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/config/channel/post-install/clusterrole.yaml
+++ b/config/channel/post-install/clusterrole.yaml
@@ -41,7 +41,6 @@ rules:
       - "list"
       - "create"
       - "update"
-      - "delete"
       - "patch"
       - "watch"
   - apiGroups:
--- a/config/channel/post-install/clusterrole.yaml
+++ b/config/channel/post-install/clusterrole.yaml
@@ -1,0 +1,53 @@
+# Copyright 2021 The Knative Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: knative-eventing-kafka-post-install-job-role
+  labels:
+    kafka.eventing.knative.dev/release: devel
+rules:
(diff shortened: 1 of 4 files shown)


matzew:
/assign @travis-minke-sap 
please take a look at this PR for the `storage-version` migrator for the `kafkachannel` - thanks!

config/channel/post-install/clusterrole.yaml
```diff
@@ -0,0 +1,53 @@
...
+  - apiGroups:
+      - "messaging.knative.dev"
+    resources:
+      - "kafkachannels"
+    verbs:
+      - "get"
+      - "list"
+      - "create"
+      - "update"
+      - "delete"
```
pierDipi, line 44: Does it need to delete KafkaChannels?

```suggestion
```

matzew: we used to have it here: https://github.com/knative-sandbox/eventing-kafka/blob/release-0.22/config/post-install/clusterrole.yaml#L45

matzew:
I was wondering before, if we should list subscriptions :smile: 

and now looking here, yes we had it too: https://github.com/knative-sandbox/eventing-kafka/blob/release-0.22/config/post-install/clusterrole.yaml#L39

So... I think we should do same ... :thinking:

travis-minke-sap:
Overall seems good to me - thanks for adding this!  I wasn't sure if you were going to make further changes (adding subscriptions permissions) so not adding `l-g-t-m` just yet - let me know if/when you're done and I'll be happy to do so.

I didn't realize we had a drop-in storage version migration - very cool - and I learned about the client-go `pager` - doh!  ; )

In general I would tend to agree with Pier's comments about only specifying the "minimal" set of permissions.  We can always add `delete` and `subscriptions` permissions if/when we need them?   I'm not overly worried about them being there either, mostly theoretical concerns ; )

matzew:
@pierDipi @travis-minke-sap I removed the `delete` for `kafkachannels`.

How do you feel generally adding a NOOP for the subscription - given the relationship between them

travis-minke-sap:
> How do you feel generally adding a NOOP for the subscription - given the relationship between them

OK with me either way...

pierDipi:
> @pierDipi @travis-minke-sap I removed the `delete` for `kafkachannels`.
> 
> How do you feel generally adding a NOOP for the subscription - given the relationship between them

I'm not sure what might happen in case the eventing storage migrator and this one run concurrently and migrate subscriptions in parallel.
It could just work but a possible test/failure case to think about, we can create an issue and discuss or test this later.

matzew:
> It could just work but a possible test/failure case to think about, we can create an issue and discuss or test this later?

good call - with this - we already have _something_ better than before

I will create a new issue for this.... and let's than get this in now :-)

pierDipi (approved):
/lgtm

PR #905, opened 2021-09-29
latest of zipkin

matzew:

Fixes #

## Proposed Changes

- updating to `:latest` 

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/test/config/monitoring.yaml
+++ b/test/config/monitoring.yaml
@@ -43,7 +43,7 @@ spec:
     spec:
       containers:
       - name: zipkin
-        image: docker.io/openzipkin/zipkin:2.13.0
+        image: ghcr.io/openzipkin/zipkin:2
         imagePullPolicy: IfNotPresent
         ports:
         - containerPort: 9411


cardil:
/lgtm

Let's try. Github registry shouldn't be so problematic with quotas as Docker Hub.

test/config/monitoring.yaml
```diff
@@ -43,7 +43,7 @@ spec:
     spec:
       containers:
       - name: zipkin
-        image: docker.io/openzipkin/zipkin:2.13.0
+        image: ghcr.io/openzipkin/zipkin:latest
```
cardil, line 46: Maybe safer would be to use: `ghcr.io/openzipkin/zipkin:2` instead? If they release zipkin 3 that breaks something, this should save us from failure.

matzew: ah, ok - that's good

will update

cardil (approved):
/lgtm

PR #945, opened 2021-10-19
Fix publish test image behavior for Golang 1.17

markusthoemmes:
## Proposed Changes

This is roughly the same change as done in https://github.com/knative-sandbox/net-kourier/commit/e7c7425e9bb2c52ef395349577c70c1871b25d29.


aliok (approved):
/approve
/lgtm

PR #972, opened 2021-11-05
Pin sarama version to v1.29.1

odacremolbap:
Due to recent dependency upgrades at eventing, [this automation](https://github.com/knative-sandbox/actions-downstream-test/blob/a4435de6e0c02b32bcd9338873f63ce2eb9cc4ad/action.yaml#L16-L21) is resolving samara to 1.30.0.

```console
2021-11-05T10:38:04.2440777Z ##[group]Run pushd ./src/knative.dev/eventing-kafka-broker
2021-11-05T10:38:04.2441767Z [36;1mpushd ./src/knative.dev/eventing-kafka-broker[0m
2021-11-05T10:38:04.2442770Z [36;1mgo mod edit -replace knative.dev/eventing=$GITHUB_WORKSPACE/src/knative.dev/eventing[0m
2021-11-05T10:38:04.2443521Z [36;1mpopd[0m
2021-11-05T10:38:04.2487470Z shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
2021-11-05T10:38:04.2488040Z env:
2021-11-05T10:38:04.2488655Z   GOPATH: /home/runner/work/eventing/eventing
2021-11-05T10:38:04.2489332Z   GOROOT: /opt/hostedtoolcache/go/1.17.2/x64
2021-11-05T10:38:04.2489863Z ##[endgroup]
```

Which is in turn making some tests at this repo fail.

https://github.com/knative/eventing/runs/4111898132?check_suite_focus=true
```
FAIL	knative.dev/eventing-kafka/pkg/channel/distributed/common/k8s [build failed]
ok  	knative.dev/eventing-kafka/pkg/channel/distributed/common/kafka/admin	0.001s
ok  	knative.dev/eventing-kafka/pkg/channel/distributed/common/kafka/admin/custom	0.002s
# knative.dev/eventing-kafka/pkg/channel/distributed/common/kafka/admin/kafka [knative.dev/eventing-kafka/pkg/channel/distributed/common/kafka/admin/kafka.test]
pkg/channel/distributed/common/kafka/admin/kafka/adminclient_test.go:50:3: cannot use mockClusterAdmin (type *MockClusterAdmin) as type sarama.ClusterAdmin in return argument:
	*MockClusterAdmin does not implement sarama.ClusterAdmin (missing AlterClientQuotas method)
[...]

PR #975, opened 2021-11-05
Upgrade Sarama To v0.30.0

travis-minke-sap:
In general we want to keep up with Sarama Kafka Client library dependencies.  Specifically not being up to date was causing problems for eventing e2e tests (see PR #972)

## Proposed Changes

- 🎁 Move Sarama from v0.29.1 to v0.30.0
- 🧽 Add missing ClusterAdmin interface functions to mocks
- 🐛 Update common offset tests to return `ApiVersionsRequest` response


matzew (approved):
Thanks for doing this, Travis!

/lgtm
/approve

PR #991, opened 2021-11-15
Replace ghodss package with kubernetes-sigs for yaml 

matzew:
## Proposed Changes

- replacing `ghodss` YAML package, which is unmaintained. Recommendation is to use https://github.com/kubernetes-sigs/yaml instead. It has the same interface and is maintained.

**Release Note**

```release-note

```

**Docs**


markusthoemmes (approved):
/lgtm
/approve

PR #1016, opened 2021-12-06
Fix rekt compile errors

pierDipi:
https://github.com/knative-sandbox/reconciler-test/pull/261
has breaking changes that this patch fixes.


matzew (approved):
/lgtm
/approve

PR #1024, opened 2021-12-09
Differentiate KafkaChannel Status ConditionSets

### LINKED ISSUE #637: Distributed KafkaChannel CRs are marked as "Not Ready" if dispatcher deployment/service isn't perfect
**Describe the bug**
The "Ready" field of a KafkaChannel (visible via "kubectl get kafkachannel [name]" and via the status fields in the full resource description) is set to "Not Ready" in the event of dispatcher deployment errors.
When a KafkaChannel is "Not Ready" then incoming messages to the Receiver service will be kicked back with an internal server error response.
The problem is that the reason for a dispatcher deployment error can be anything from "I asked for 5 replicas but I could only get 4" to a full-blown complete misconfiguration of the deployment.  This lack of nuance, and the fact that a dispatcher is not actually required to be in a functional state for the receiver to collect and forward messages to the Kafka service, makes the "KafkaChannel Not Ready" behavior generally undesirable.

 **Expected behavior**
A KafkaChannel should be "Ready" as long as the KafkaChannel service, Receiver Service/Deployment, and Topic are ready.  The Dispatcher state should be available for inspection but should not prevent messages from being accepted and stored in the Kafka backend.  Alternatively, this behavior could be toggled via a setting in the ConfigMap (or an environment variable), but the default should be as described.

 **To Reproduce**
One easy way to reproduce this is to set the number of replicas for the Dispatcher in the config-kafka configmap to a number higher than the cluster can provide, or the MemoryRequest to a similarly impossible value.

 **Knative release version**
0.23.0

 **Additional context**
The distributed channel shares the same status block with the consolidated channel, which has more of an interest in marking the KafkaChannel as "Not Ready" due to the combined nature of the dispatcher/receiver deployments, so possibly the default behavior should be different between the two channel types (and/or implementing a mechanism for the status blocks to be distinct).

travis-minke-sap:
Fixes #637

The primary concern of Issue #637 is that the Receiver not block ingestion of events into a Kafka Topic when the egress  side (Dispatchers) might be down or scaling.  The Issue presumes a specific solution, specifically **not** marking the KafkaChannel's overall status as Not READY.  This seems to go against the intent of the single aggregated status field, so instead I am proposing that we adjust the Receiver to check the ingress related Status conditions when making the determination.

This raises a long-standing bit of _tech-debt_ regarding the sharing of Status Conditions between the consolidated and distributed KafkaChannel implementations.  The current Status Conditions are historically those of the consolidated implementation and have always been "made to work" by the distributed implementation having to map it's runtime artifacts into those conditions.

Therefore, it seems like a good time to tackle the problem and provide a mechanism for the two channel implementations to each have their own set of Status Conditions that accurately reflect their runtime architecture.   That is what the bulk of this PR covers, though the Receiver check is also included ; )

## Proposed Changes

- 🧽 Create new lifecycle Conditions to align with distributed architecture.
- 🧽 Add utility functions to lifecycle to allow differentiation of ConditionSets.
- 🧽 Add init() blocks to KafkaChannel implementations main.go to specify desired ConditionSet.
- 🧽 Update distributed KafkaChannel Controller to use new Conditions.
- 🧽 Update distributed Receiver channel validation to only consider ingress status (Topic and Receiver).

**Release Note**

```release-note
Distributed KafkaChannel Status conditions refactored to accurately reflect runtime artifacts.  Receiver no longer considers Dispatcher state when rejecting events and instead only evaluates ingress status.
```

**Docs**


pkg/apis/messaging/v1beta1/kafka_channel_lifecycle_test.go
```diff
@@ -373,6 +526,169 @@ func TestChannelIsReady(t *testing.T) {
 	}
 }
 
+func TestDistributedChannelIsReady(t *testing.T) {
```
eric-sap, line 529: There are a lot of tests here, but not any for the "Unknown" statuses (MarkDispatcherDeploymentUnknown, etc). Are those pesky to test? Not a big loss if it's complicated but if it's just a matter of having a "deploymentStatusUnknown" as well it might be good to add them.

travis-minke-sap: Yeah, I noticed that - carryover from the legacy test.  Not sure the check buys much but I will enhance it to do so - helps the GA coverage numbers if nothing else : )

pkg/apis/messaging/v1beta1/kafka_channel_lifecycle.go
```diff
@@ -122,6 +187,34 @@ func (kcs *KafkaChannelStatus) SetAddress(url *apis.URL) {
...
+	kcs.GetConditionSet().Manage(kcs).MarkFalse(KafkaChannelConditionChannelServiceReady, reason, messageFormat, messageA...)
+}
+
+func (kcs *KafkaChannelStatus) MarkChannelServiceTrue() {
+	kcs.GetConditionSet().Manage(kcs).MarkTrue(KafkaChannelConditionChannelServiceReady)
+}
+
+//
+// Consolidated KafkaChannel Condition Markers
+//
```
pierDipi, line 216: All these implementation-specific `Mark*` methods are confusing, if we add another KafkaChannel implementation do we expect to add more of these?

We should instead move them out of this package and put them as part of the reconciler that uses them. WDYT?

pkg/apis/messaging/v1beta1/kafka_channel_lifecycle.go
```diff
@@ -53,22 +79,57 @@ const (
...
+			KafkaChannelConditionConfigReady,
+			KafkaChannelConditionTopicReady,
+			KafkaChannelConditionChannelServiceReady,
+			KafkaChannelConditionReceiverServiceReady,
+			KafkaChannelConditionReceiverDeploymentReady,
+			KafkaChannelConditionDispatcherServiceReady,
+			KafkaChannelConditionDispatcherDeploymentReady,
+		),
+	)
+}
```
pierDipi, line 131: Same comment for these.

pkg/apis/messaging/v1beta1/kafka_channel_lifecycle.go
travis-minke-sap: Yeah, I'm open to that - should be an easy refactor to separate and move them somewhere like `pkg/channel/consolidated/apis/lifecycle.go` .   Actually, this might cause problems for versioning the api (v1beta1, etc) - so it would have to mirror the apis structure like `pkg/channel/consolidated/apis/messaging/v1beta1/lifecycle.go` ?

Anyone else have an opinion?  If not, I'll refactor as described above ; )

pkg/apis/messaging/v1beta1/kafka_channel_lifecycle.go
travis-minke-sap: 👍 will move everything channel specific together based on comment above!

pkg/apis/messaging/v1beta1/kafka_channel_lifecycle.go
pierDipi: The condition set doesn't need versions.

The main reasons are: 
- they're tied to a reconciler architecture and logic, not to the API.
- If the reconciler changes the API that it reconciles to be, let's say, `v1`, the condition set doesn't change automatically.

travis-minke-sap: ah, good call - thanks - makes it a little less messy ; )

travis-minke-sap: I've refactored them out as suggested.  This means these "custom" lifecycle methods are no longer functions on the KafkaChannelStatus but instead pass-through utilities.  There are pros/cons but overall it seems good ; )

pkg/apis/messaging/v1beta1/kafka_channel_lifecycle.go
```diff
@@ -19,54 +19,43 @@ package v1beta1
...
 	duckv1 "knative.dev/pkg/apis/duck/v1"
 )
 
+// A default ConditionSet containing base configuration only for testing.
+// The consolidated and distributed KafkaChannel implementations require
+// differentiated ConditionSets in order to accurately reflect their varied
+// runtime architectures.  One of the channel specific "Register..." functions
+// in pkg/channel/<type>/apis/messaging/kafka_channel_lifecycle.go should be
+// called via an init() in the main() of associated components.
 var kc = apis.NewLivingConditionSet(
```
pierDipi, line 32: If these are only for testing can we put them in a `*_test.go` file?

travis-minke-sap: 👍

travis-minke-sap:
@pierDipi - Any further changes or can we merge this?

pierDipi (approved):
Thanks for this!

/lgtm

PR #1025, opened 2021-12-09
Update KafkaChannel CRD to support experimental-features in Delivery

travis-minke-sap:
In order to allow experimental-features related to the Delivery Spec (Timeout, RetryAfterMax, etc.) we need to loosen constraints on that part of the KafkaChannel CRD.

## Proposed Changes

- 🧽  Add preserve-unknown-fields flag to "delivery" portions of the KafkaChannel CRD.

**Release Note**

```release-note

```

**Docs**


matzew (approved):
Ah,... thanks :smile: 

/lgtm
/approve

PR #1037, opened 2021-12-22
Fix ChannelFeatureMap since KafkaChannel is v1beta1 not v1

pierDipi:
`resources.MessagingAPIVersion` is equal to
messaging.knative.dev/v1

/assign @aliok


aliok (approved):
/lgtm
/approve

PR #1038, opened 2022-01-04
Validate SourceSpec in KafkaSource

pierDipi:
We need to validate the SourceSpec field as part of the KafkaSource
webhook validation.

The E2E test for KafkaSource is trying specify the type as
`CEOverride` which is disallowed by the Source spec:

> When spec.ceOverrides.extensions is specified and non-empty,
> the event sent to the sink is augmented or modified with the
> specified extension attributes.
> The source controller MUST reject extensions overriding the
> required Cloud Events attributes.

Ref: https://github.com/knative/eventing/blob/main/docs/spec/sources.md#duckspec

Signed-off-by: Pierangelo Di Pilato <pdipilat@redhat.com>

Part of upgrading dependencies in #1035

**Release Note**

```release-note
The KafkaSource webhook reject extensions overriding the required Cloud Events attributes as specified by the SourceSpec
https://github.com/knative/eventing/blob/main/docs/spec/sources.md#duckspec.
```

### DIFF
--- a/pkg/apis/sources/v1beta1/kafka_validation.go
+++ b/pkg/apis/sources/v1beta1/kafka_validation.go
@@ -36,8 +36,8 @@ func (ks *KafkaSource) Validate(ctx context.Context) *apis.FieldError {
 func (kss *KafkaSourceSpec) Validate(ctx context.Context) *apis.FieldError {
 	var errs *apis.FieldError
 
-	// Validate sink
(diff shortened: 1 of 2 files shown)


pierDipi:
Sink field validation in SourceSpec is missing: https://github.com/knative/pkg/pull/2384

matzew (approved):
/lgtm
/approve

PR #1044, opened 2022-01-05
Consumer manager use factory merged errors

### LINKED ISSUE #1043: Repeated broker failures cause distributed dispatcher to hang
**Describe the bug**
When using the distributed channel type, the 11th time a consumer group fails (typically due to inability to connect to the broker), the dispatcher pod will hang and not re-call the ConsumerGroup's consume() function.  The only way to recover is to restart the dispatcher pod.

 **Expected behavior**
broker connection issues should not break the dispatcher.  When the broker is available again, the ConsumerGroups should resume their consume() functions.

 **To Reproduce**

- Create a distributed channel and subscription and verify basic message functionality
- Monitor the dispatcher log for the channel
- Stop the broker (for example, with a Strimzi kafka backend, removing the Strimzi pods is sufficient)
- Keep the broker down until the 11th time Sarama fails to reconnect to the broker (e.g. keep killing the Strimzi pods).  The dispatcher log (if debug is enabled in the config-logging configmap) will show a "Managed Consume Finished Without Stop" debug message each time this happens (and if Sarama logging is enabled in the config-kafka configmap there will be voluminous output from Sarama).
- After the 11th failure, the dispatcher pod will hang

 **Knative release version**
1.0.0

 **Additional context**
The hanging has been identified as happening in the startExistingConsumerGroup() function when "errorCh <- err" is executed.  The errorCh is never read so when the channel buffer fills, it blocks.

eric-sap:
Fixes #1043

- 🐛 ConsumerGroup manager was only reading from the ConsumerGroup error channel and not the merged channel from the KafkaConsumerGroupFactory.  This has been rectified.

Note:  As this is potentially a large problem and the fix is relatively small, this fix should be back-ported

### DIFF
--- a/pkg/common/consumer/consumer_factory.go
+++ b/pkg/common/consumer/consumer_factory.go
@@ -102,19 +102,20 @@ func (c kafkaConsumerGroupFactoryImpl) startExistingConsumerGroup(
 	errorCh := make(chan error, 10)
 	releasedCh := make(chan bool)
 	ctx, cancel := context.WithCancel(context.Background())
+	groupLogger := logger.With(zap.Any("topics", topics), zap.String("groupId", groupID), zap.String("channel", channelRef.String()))
 
 	go func() {
 		// this is a blocking func
 		// do not proceed until the check is done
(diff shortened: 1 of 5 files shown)


eric-sap:
/retest

image upload failed?

PR #1060, opened 2022-01-14
fix vulnerabilities

steven0711dong:
Fixes #

## Proposed Changes

-
-
-

**Release Note**

```release-note
Fixes the following vulnerabilities: 
github.com/knative/pkg contains dependency that is subject to DoS attack. 
githun.com/kubernetes/utils contains a security issue that was discovered where a user may be able to create a container with subpath volume mounts to access files & directories outside of the volume, including on the host filesystem.
```

**Docs**


matzew (approved):
/lgtm
/approve

PR #1068, opened 2022-01-26
Custom leader-election configuration for KafkaChannel

travis-minke-sap:
The semi-recent automated update of dependencies #1058 pulled in changes in **knative.dev/pkg/leader-election/config.go** which are breaking the startup of the KafkaChannel Webhook.  Specifically the default config timings were relaxed to be less frequent...

```
func defaultConfig() *Config {
	return &Config{
		Buckets:       1,
		LeaseDuration: 60 * time.Second,
		RenewDeadline: 40 * time.Second,
		RetryPeriod:   10 * time.Second,
	}
}
```

The KafkaChannel Webhook deployment has a 20 second delay on the initial liveness probe...
```
        livenessProbe:
          <<: *probe
          initialDelaySeconds: 20
```

So the webhook is restarted continually before it can finish the leader election setup.

This is a problem when uninstalling and re-installing in an existing namespace as the leases are not removed.  If the namespace is removed this shouldn't be an issue.  For "optional" components like eventing-kakfa that live in the knative-eventing namespace, though it is common.  There is a large Slack thread on this outage in the eventing-kafka channel on Jan 26, 2022.

## Proposed Changes

The goal of this PR is a quick rollback/override to use the prior configuration values so that build/releases may proceed.  If in the future we want to undo this and find another solution we can always do so.

- Add a new **config/channel/config/leader-election-configmap.yaml** with prior configuration values.
- Symlinks from **config/channel/consolidated** and **config/channel/distributed** to include the new ConfigMap
- Change both consolidated and distributed controller Deployments to use the new ConfigMap instead of relying on the standard knative-eventing leader-election ConfigMap.
- Add the CONFIG_LEADERELECTION_NAME env var to the Webhook Deployment(s) referencing the new ConfigMap.

### DIFF
--- a/config/channel/distributed/500-controller-deployment.yaml
+++ b/config/channel/distributed/500-controller-deployment.yaml
@@ -52,7 +52,7 @@ spec:
         - name: CONFIG_LOGGING_NAME
           value: config-logging
         - name: CONFIG_LEADERELECTION_NAME
-          value: config-leader-election
+          value: config-leader-election-kafkachannel
         - name: SERVICE_ACCOUNT
           valueFrom:
             fieldRef:
@@ -70,9 +70,9 @@ spec:
             cpu: 20m
             memory: 25Mi
         volumeMounts:
-          - name: config-kafka
-            mountPath: /etc/config-kafka
-      volumes:
         - name: config-kafka
-          configMap:
-            name: config-kafka
+          mountPath: /etc/config-kafka
+      volumes:
+      - name: config-kafka
+        configMap:
+          name: config-kafka
--- a/config/channel/distributed/500-controller-deployment.yaml
+++ b/config/channel/distributed/500-controller-deployment.yaml
@@ -70,9 +70,9 @@ spec:
             cpu: 20m
             memory: 25Mi
         volumeMounts:
-        - name: config-kafka
-          mountPath: /etc/config-kafka
+          - name: config-kafka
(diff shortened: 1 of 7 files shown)


travis-minke-sap:
Hold for consensus on approach ; )
/hold

config/channel/distributed/500-controller-deployment.yaml
```diff
@@ -70,9 +70,9 @@ spec:
             cpu: 20m
             memory: 25Mi
         volumeMounts:
-          - name: config-kafka
-            mountPath: /etc/config-kafka
-      volumes:
         - name: config-kafka
```
lionelvillard, line 73: the indent is off (but still correct) thinking there is a change when there is not?

travis-minke-sap: ah... intellij.... sigh... - I'll correct it.   pretty sure it works either way but good to stay consistent

lionelvillard: nvm, it's consistent with the rest

evankanderson:
@evankanderson -- unexpected fallout from the pkg changes

Given the above scenario, it seems reasonable / prudent to check on webhook liveness probes and testing in all repos.

Opened https://github.com/knative/pkg/issues/2422 to investigate

PR #1085, opened 2022-02-07
Change KafkaChannel default replication factor and num partitions

### LINKED ISSUE #1006: Change KafkaChannel defaults for partition and replica
**Problem**
The current defaults for setting `partitions` _and_ `replica` to `1` is a pretty poor choice.

We noticed in tests that it is much much better if we default to something like `10 partitions` and `3 replica`

The issue here is that `3` replica does not work single broker instances, but those are generally considered being a toy.

For POCs or demos, the `replica` can be still configured to be `1`, on the actual `KafkaChannel` object - so single broker usage is still possible

**Exit Criteria**

**Time Estimate (optional):**

pierDipi:
Fixes #1006

The current defaults of setting partitions and replication factor to 1
aren't safe defaults for performance and durability.

This PR sets KafkaChannel default replication factor to 3 and
num partitions to 10.

## Proposed Changes

- Change KafkaChannel default replication factor and num partitions

**Release Note**

```release-note
KafkaChannel default replication factor is 3 and num partitions is 10.
```

### DIFF
--- a/pkg/common/constants/constants.go
+++ b/pkg/common/constants/constants.go
@@ -21,10 +21,10 @@ import "time"
 const (
 
 	// DefaultNumPartitions is the KafkaChannel Spec default for the number of partitions
-	DefaultNumPartitions = 1
+	DefaultNumPartitions = 10
 
 	// DefaultReplicationFactor is the KafkaChannel Spec default for the replication factor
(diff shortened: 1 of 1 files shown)


travis-minke-sap:
Thanks!
/lgtm

PR #1090, opened 2022-02-11
[WIP] Individual config in KafkaChannel

### LINKED ISSUE #1005: KafkaChannel Configuration option on a per object base
**Problem**

We currently have a global configuration for _one_ dedicated Apache Kafka cluster:
https://github.com/knative-sandbox/eventing-kafka/blob/main/config/channel/consolidated/configmaps/kafka-config.yaml#L24

this forces _all_ `KafkaChannel` objects to be connected against a single kafka cluster.

This does not fly well, when offering "Knative Kafka" as a service, where customers have desire for individual Kafka cluster usage, on a per channel bases.

The global default is a convenient setting for on prem, but not for more cloud-centric cases 

## Proposal

Adding the following fields to the `KafkaChannel` spec

- `bootstrapServers`
-  `authSecretName` (assumes that the actual secret is living in the namespace of the `KafkaChannel` object)

**Exit Criteria**

**Time Estimate (optional):**

aliok:
Fixes #1005

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/apis/messaging/v1beta1/kafka_channel_types.go
+++ b/pkg/apis/messaging/v1beta1/kafka_channel_types.go
@@ -64,6 +64,22 @@ var (
 
 // KafkaChannelSpec defines the specification for a KafkaChannel.
 type KafkaChannelSpec struct {
+	// BootstrapServers is the Kafka Channel bootstrap servers. When not provided, the implementation will
+	// fall back to the configmap in the config field. If there are also no bootstrapServers defined in that,
+	// implementation will fall back to the global config.
+	// +optional
+	BootstrapServers []string `json:"bootstrapServers,omitempty"`
+
+	// Auth configurations.
+	// +optional
+	Auth *Auth `json:"auth,omitempty"`
+
+	// Config is a KReference to the configuration that specifies
+	// configuration options for this Channel. For example, this could be
+	// a pointer to a ConfigMap.
+	// +optional
+	Config *duckv1.KReference `json:"config,omitempty"`
+
 	// NumPartitions is the number of partitions of a Kafka topic. By default, it is set to 1.
 	NumPartitions int32 `json:"numPartitions"`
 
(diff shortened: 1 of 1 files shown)


pkg/apis/messaging/v1beta1/kafka_channel_types.go
```diff
@@ -64,6 +64,22 @@ var (
 
 // KafkaChannelSpec defines the specification for a KafkaChannel.
 type KafkaChannelSpec struct {
+	// BootstrapServers is the Kafka Channel bootstrap servers. When not provided, the implementation will
+	// fall back to the configmap in the config field. If there are also no bootstrapServers defined in that,
```
matzew, line 68: for the future: do we envision to retire the global config, all together?

pierDipi (commented):
LGTM

can you run `hack/update-codegen.sh`?

pkg/apis/messaging/v1beta1/kafka_channel_types.go
```diff
@@ -64,6 +64,22 @@ var (
...
+
+	// Auth configurations.
+	// +optional
+	Auth *Auth `json:"auth,omitempty"`
+
+	// Config is a KReference to the configuration that specifies
+	// configuration options for this Channel. For example, this could be
+	// a pointer to a ConfigMap.
+	// +optional
+	Config *duckv1.KReference `json:"config,omitempty"`
```
pierDipi, line 81: what's the use case for this?

matzew:
@aliok with the on-going data-plane isolation, I think we should keep this proposal also in our backheads, as this allows individual configuration of the kafka channel's kafka (similar to the `broker.spec.config`).

Currently there is no MT usage o the channel, due to this global setting, and no individual config for kafka cluster

matzew:
/close
the channel has been moved over to the `eventing-kafka-broker` repo

PR #1101, opened 2022-02-21
Set resource limit for single-tenant Kafka source

### LINKED ISSUE #1008: Limit single-tenant Kafka source use of resources
**Problem**
A short explanation of the problem, including relevant restrictions.

Today single-tenant Kafka source when running has no resource limits set - effectively taking as much  (or as little) CPU as available in Kubernetes running the node. 

System Integrator
System Operator

**Exit Criteria**

I have run tests and verified that without resource limits pod will be taking as much resources as it can get.

After fixing the problem we should see that pod resource consumption is limited.

**Time Estimate (optional):**

aslom:
Fixes #1008

## Proposed Changes

- 🧽 Limit single-tenant Kafka source use of resources such as CPU and memory

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/source/reconciler/source/resources/receive_adapter.go
+++ b/pkg/source/reconciler/source/resources/receive_adapter.go
@@ -24,6 +24,7 @@ import (
 
 	v1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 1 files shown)


matzew:
I understand the need for setting those, but not 100% sure on the numbers. Where do they come from?

Also, since we move the source development forward in https://github.com/knative-sandbox/eventing-kafka-broker - we should investigate there too

matzew (approved):
/lgtm
/approve

PR #1158, opened 2022-04-26
Generate consumer group in KafkaSourceUpdate test

pierDipi:
Using the same consumer group for the different sources might produce
unreliable results because Kafka group ids are shared.

This is particularly evident in eventing-kafka-broker where the same
tests (`KafkaSourceUpdate`) is executed 5 times [1] in parallel [2]
(to catch flakes early), which means that there might be different
sources with the same consumer group pointing to 2 different sinks.

[1] https://github.com/knative-sandbox/eventing-kafka-broker/blob/e169e2cb923890a23e8d65651876292283b63908/test/e2e/kafka_source_test.go#L31
[2] https://github.com/knative-sandbox/eventing-kafka-broker/blob/e169e2cb923890a23e8d65651876292283b63908/test/pkg/testing/run.go#L52-L67

Signed-off-by: Pierangelo Di Pilato <pdipilat@redhat.com>

### DIFF
--- a/test/e2e/helpers/kafka_source_test_helper.go
+++ b/test/e2e/helpers/kafka_source_test_helper.go
@@ -78,6 +78,7 @@ func testKafkaSourceUpdate(t *testing.T, name string, test updateTest) {
 	client.WaitForAllTestResourcesReadyOrFail(context.Background())
 
 	kafkaSourceName := "e2e-kafka-source-" + name
+	groupId := client.Namespace + "-" + name
 
 	t.Logf("Creating kafkasource: %s\n", kafkaSourceName)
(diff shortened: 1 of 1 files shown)


test/e2e/helpers/kafka_source_test_helper.go
```diff
@@ -78,6 +78,7 @@ func testKafkaSourceUpdate(t *testing.T, name string, test updateTest) {
 	client.WaitForAllTestResourcesReadyOrFail(context.Background())
 
 	kafkaSourceName := "e2e-kafka-source-" + name
+	groupId := client.Namespace + "-" + name
```
matzew, line 81: ah, that's good for a test fixture

matzew (approved):
/lgtm
/approve

matzew:
@pierDipi wanna backport?

PR #1164, opened 2022-04-29
[release-1.2] Going with debug to reduce verbosity for retention duration default usage

matzew:

Fixes #

## Proposed Changes

- As per title: There is no reason to do an error log, since we do have a default value to be used. Going with debug to reduce verbosity...

**Release Note**

```release-note

```

**Docs**


matzew:
/close
doing on "broker" repo, instead

PR #1259, opened 2022-10-18
Deprecate components

matzew:

## Proposed Changes

- Adding note that some components in repo are no longer maintained.

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,12 @@
 [![Slack](https://img.shields.io/badge/Signup-Knative_Slack-white.svg?logo=slack)](https://slack.knative.dev)
 [![Slack](https://img.shields.io/badge/%23eventing-white.svg?logo=slack&color=522a5e)](https://knative.slack.com/archives/C9JP909F0)
 
+** Deprecation Note**
+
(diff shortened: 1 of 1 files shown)


pierDipi:
Can we remove them from the released YAMLs: https://github.com/knative-sandbox/eventing-kafka/blob/07af4e8e1ee3e5469b94073bc497b2619c310650/hack/release.sh#L26-L34

pierDipi:
Related discussion: https://knative.slack.com/archives/C9JP909F0/p1667841421355339

dprotaso:
If the components are specific to certain packages you can probably mark those packages are deprecated using godoc

https://go.dev/blog/godoc

creydr:
> Can we remove them from the released YAMLs:
> 
> https://github.com/knative-sandbox/eventing-kafka/blob/07af4e8e1ee3e5469b94073bc497b2619c310650/hack/release.sh#L26-L34

Should we continue on this? We have some CI upgrade test failures, as it uses some deprecated API version for PodDisruptionBudget and HPA (see #1287 for example). Or would we first have to deprecate but keep them in the yamls for a few releases?

pierDipi:
@creydr yes, please, we cannot move forward with the release (yesterday) without this

pierDipi:
See https://github.com/knative-sandbox/eventing-kafka/pull/1327

PR #1278, opened 2022-12-26
Use PodDisruptionBudget v1

aliok:
Fixes #

## Proposed Changes

- Use PodDisruptionBudget v1
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/config/channel/webhook/webhook-hpa.yaml
+++ b/config/channel/webhook/webhook-hpa.yaml
@@ -35,7 +35,7 @@ spec:
         averageUtilization: 100
 ---
 # Webhook PDB.
-apiVersion: policy/v1beta1
+apiVersion: policy/v1
 kind: PodDisruptionBudget
 metadata:
   name: kafka-webhook


aliok:
During upgrade tests, installing the previous version fails because of inexistent PodDisruptionBudget in policy/v1beta1.

matzew (approved):
/lgtm
/apporve

matzew:
> During upgrade tests, installing the previous version fails because of inexistent PodDisruptionBudget in policy/v1beta1.

I'd assume that the CI for the previous job is also on the same k8s version as this one? If so, hard to address.
Unless we go all the way back (assuming older versions of our software run on older k8s versions where this actually fits).

That said, but I am not sure really worth it, see:  https://github.com/knative-sandbox/eventing-kafka/pull/1259

aliok:
/close 

Context: https://knative.slack.com/archives/C02C56QF282/p1672925909239859

PR #1281, opened 2023-01-03
[1.7] Use PodDisruptionBudget v1

aliok:
Fixes #

## Proposed Changes

- Use PodDisruptionBudget v1
- Need to fix this old release as there are issues when we try `eventing-kafka` with newer Kubernetes versions
-

**Release Note**

```release-note
🐛 Upgrade PodDisruptionBudget used in KafkaChannel webhook from policy/v1beta1 to policy/v1
```

**Docs**

### DIFF
--- a/config/channel/webhook/webhook-hpa.yaml
+++ b/config/channel/webhook/webhook-hpa.yaml
@@ -35,7 +35,7 @@ spec:
         averageUtilization: 100
 ---
 # Webhook PDB.
-apiVersion: policy/v1beta1
+apiVersion: policy/v1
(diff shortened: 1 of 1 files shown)


aliok:
/close

To make the upgrade tests pass, we need to release 1.5, merge https://github.com/knative-sandbox/eventing-kafka/pull/1282 and then release 1.6. 
Releasing 1.5 doesn't actually seem very possible.

This repo is deprecated anyway.

Context: https://knative.slack.com/archives/C02C56QF282/p1672925909239859

PR #1332, opened 2023-04-19
Updated deps and removed deleted bits

matzew:
Fixes #

## Proposed Changes

- Updating to latest eventing, and getting rid of delted test image "heartbeats", using actual instead.

**Release Note**

```release-note

```

**Docs**


creydr (approved):
/lgtm

PR #1333, opened 2023-04-20
Bump20230420

matzew:
## Proposed Changes

- go 1.19 and bump

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module knative.dev/eventing-kafka
 
-go 1.18
+go 1.19
 
 require (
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module knative.dev/eventing-kafka
 
-go 1.18
+go 1.19
```
dsimansk, line 3: Is it really necessary to bump min go version?

dsimansk: Asking out of curiosity since knative/pkg it still OK at 1.18.

PR #1345, opened 2023-06-01
Update dependencies and fix compile issues

creydr:
Dependency bump as #1344 + fixing the compile issues


pierDipi (approved):
/lgtm
/approve

PR #1371, opened 2023-08-04
Rename knative-sandbox to knative-extensions

davidhadas:
This is an automated PR to replace  knative-sandbox strings with knative-extensions


matzew (approved):
/lgtm
/approve

davidhadas:
This PR fails due to shellcheck tests unrelated to the renaming. 
See https://github.com/knative-extensions/eventing-kafka/actions/runs/5758835365/job/15611946836?pr=1371

Please fix the project code such that it passes shellcheck tests in a separate PR, once this is done we can get this PR going

matzew:
@davidhadas I don't see what's wrong (or why the other PRs go in just fine, like updates of dependencies)

davidhadas:
> @davidhadas I don't see what's wrong (or why the other PRs go in just fine, like updates of dependencies)

My best guess (and maybe @upodroid can confirm or correct me) is that the productivity team made some changes at some past time. Maybe they added `shellcheck` tests or maybe they changed the `shellcheck` configuration. As a result, some repositories became non compliant with respect to the `shellcheck` tests. 

However, these tests are performed only when someone changes a file, so any file not changed since the productivity team made the changes goes unnoticed. Once someone make any change in a file which fail these tests, the tests are performed and the PR fail. 

If this is true, I think this process in productivity  should change  - the process should not rely on future PRs failing as it leads to a very long tail of failing PRs and confused developers.

But as I said, this is nothing more than my working theory :)

PR #1458, opened 2024-01-18
Update deps 2024-01-18

Cali0707:
Fixes the build issues when updating dependencies

/cc @creydr @pierDipi @Leo6Leo


creydr (approved):
Thanks for fixing this @Cali0707 

/lgtm

